PR #2, opened 2022-02-02
Random Utils

jkeon:
Added some utility methods to make working `Random` instances in a DOTS/Entities context easier to manage.

**Creation**
Currently when you create an instance of `Random` you need to supply a seed. Doing so is annoying because you either need to use a different random number generator to get that seed or more commonly, you use a function of the current date and time. 

You can run into a few issues with this:

1. If you create multiple instances of `Random` at the same time and you are using a function of time for your seed, you will have the same seed for all of them, making them not very random. 
2. If you get unlucky and generate a `0` for your seed, the `Random` instance will crash.
3. Depending on how you're generating your seed, it's an unnecessary extra performance cost.  ex: `DateTime.Now`
    3a. Or just confusing if you're using a different Random number generator

**Threading**
Because `Random` is a struct and it gets copied into multiple threads when running in parallel, all the copied instances have the same state and will generate the same random numbers making it not very random at all. 

**Creation**
By using `RandomUtil.CreateSeeded()` you will get a new instance of `Random` returned that has already been seeded with a random seed between `1` and `uint.MaxValue`. 

1. If you create multiple instances with this function at the same time, they will all have different seeds. 
2. You cannot ever get a `0` for a seed.
3. No performance hit or confusion, it uses an internal `Random` instance under the hood.

**Threading**
At the beginning of your `ForEach` or the `Execute` of your `IJob`, you can call `myRandomInstance.UpdateSeedWithNativeThreadIndex` and pass in the thread index. This will ensure that all of your threads who started on the same random state will now have unique states relative to their thread indexes. The values from those randoms will be widely distributed as well.

 - None

 - None

 - [ ] Yes 
 - [x] No

### Usage

```
            //Create the instance on the main thread
            Random random = RandomUtil.CreateSeeded();

            Entities.WithName(nameof(MakeDecisionSystem))
                    .ForEach((Entity entity,
                              int entityInQueryIndex,
                              ref ModelDecision modelDecision,
                              in EntityParentLink entityParentLink,
                              in int nativeThreadIndex) =>
                             {
                                 //Modify to be unique per thread
                                 random.UpdateSeedWithNativeThreadIndex(nativeThreadIndex);
                                 
                                 //Then use as you normally would
                                 float choice = random.NextFloat(0.0f, 1.0f);
                                 
                                 modelDecision.TimeToDecision += choice < 0.5f
                                     ? -random.NextFloat(0.0f, 3.0f)
                                     : random.NextFloat(0.0f, 3.0f);
                             })
                    .ScheduleParallel(Dependency);
```

### Additional Notes

There are quite a few forum posts on the Unity ECS Forums about Random number generation and one of the approaches is to pre-allocate a `NativeArray` of different `Random` instances indexed by the thread index. 

https://forum.unity.com/threads/solved-gen-random-number-in-job-execute-causes-hang.779987/#post-5191244

The purpose here is be able to preserve the state of a `Random` instance after it exits the thread. We found that writing back to the `NativeArray` caused performance to drop by an order of magnitude. 0.05ms increased to 0.5ms for parallel jobs across 100,000 entities. 

Having the state preserved offered no real benefit since it's simple and cheap to create a new instance with a random seed.

Additionally calling `random.UpdateSeedWithNativeThreadIndex(nativeThreadIndex);` will get executed everytime the ForEach loops which is unnecessary but the method is cheap and better than doing an if check to see if you've called it before. 

When C#8 is available in Unity (note the version of Unity which ECS currently supports which at the time of writing is 2020.1.9 instead of the latest 2022 version), this could be done once per thread. 
https://forum.unity.com/threads/is-it-possible-to-allocate-a-new-temp-nativelist-per-thread-in-entities-foreach.885940/

### DIFF
--- a/Scripts/Runtime/Util/Random/RandomExtensions.cs
+++ b/Scripts/Runtime/Util/Random/RandomExtensions.cs
@@ -11,14 +11,15 @@ namespace Anvil.Unity.DOTS.Util
     {
         private const uint SPREAD = uint.MaxValue / JobsUtility.MaxJobThreadCount;
         
+        //TODO: Revisit when we get C# 8 support in Unity ECS https://github.com/decline-cookies/anvil-unity-dots/issues/3
         /// <summary>
         /// Given an instance of <see cref="Random"/>, this extension modifies the <see cref="Random"/>'s state so that
-        /// multiple threads using the same initial state will have output different random values based on the thread
+        /// multiple threads using the same initial state will output different random values based on the thread
         /// index.
         /// </summary>
         /// <remarks>
         /// A typical use case with <see cref="Random"/> is to create an instance on the main thread with a given seed.
-        /// Then pass that instance into a job which will get split across many different threads to do some work that
+        /// Then pass that instance into a job which will get copied across many different threads to do some work that
         /// requires randomized values.
         ///
         /// Because the instance is copied into each of the threads they all have the same state and will all output the
--- a/Scripts/Runtime/Util/Random/RandomExtensions.cs
+++ b/Scripts/Runtime/Util/Random/RandomExtensions.cs
@@ -1,0 +1,38 @@
+using System.Runtime.CompilerServices;
+using Unity.Jobs.LowLevel.Unsafe;
+using Unity.Mathematics;
+
+namespace Anvil.Unity.DOTS.Util
+{
+    /// <summary>
+    /// Extension methods for use with a <see cref="Random"/> instance.
+    /// </summary>
+    public static class RandomExtensions
+    {
+        private const uint SPREAD = uint.MaxValue / JobsUtility.MaxJobThreadCount;
+        
+        /// <summary>
+        /// Given an instance of <see cref="Random"/>, this extension modifies the <see cref="Random"/>'s state so that
+        /// multiple threads using the same initial state will have output different random values based on the thread
+        /// index.
+        /// </summary>
+        /// <remarks>
+        /// A typical use case with <see cref="Random"/> is to create an instance on the main thread with a given seed.
+        /// Then pass that instance into a job which will get split across many different threads to do some work that
+        /// requires randomized values.
+        ///
+        /// Because the instance is copied into each of the threads they all have the same state and will all output the
+        /// exact same "random" values which defeats the purpose. This function uses the unique thread index and
+        /// multiplies it by a spread to take up an evenly distributed wide range of bits of
+        /// an <see cref="uint.MaxValue"/>. This value is then XOR'd with the existing state resulting in each thread
+        /// having it's own unique state capable of producing widely divergent random numbers from each other.
+        /// </remarks>
+        /// <param name="random">A ref to the <see cref="Random"/> instance to modify</param>
+        /// <param name="nativeThreadIndex">The thread index to modify this the state with</param>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static void UpdateSeedWithNativeThreadIndex(this ref Random random, int nativeThreadIndex)
+        {
+            random.InitState((uint)(random.state ^ (nativeThreadIndex * SPREAD)));
+        }
+    }
+}
--- a/Scripts/Runtime/Util/Random/RandomUtil.cs
+++ b/Scripts/Runtime/Util/Random/RandomUtil.cs
@@ -12,11 +12,11 @@ namespace Anvil.Unity.DOTS.Util
     {
         private static Random s_InternalRandom;
 
-        //With DOTS, it's recommended to disable the Domain Reload when entering play mode as it will be pretty slow.
-        //https://docs.unity3d.com/Packages/com.unity.entities@0.17/manual/install_setup.html
-        //Because of this, statics will preserve state across play sessions. By using the RuntimeInitializeOnLoadMethod
-        //we can ensure we reset the state on every play session.
-        //RuntimeInitializeLoadType.SubsystemRegistration represents the earliest moment of all the entries in the enum.
+        // With the Burst compiler, it's common to disable the Domain Reload step when entering play mode to increase iteration time.
+        // https://docs.unity3d.com/Packages/com.unity.entities@0.17/manual/install_setup.html
+        // Because of this, statics will preserve state across play sessions.
+        // `RuntimeInitializeOnLoadMethod` ensures we reset the state on every play session.
+        // `RuntimeInitializeLoadType.SubsystemRegistration` represents the earliest moment of all the entries in the enum.
         [UnityEngine.RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.SubsystemRegistration)]
         private static void Init()
         {
@@ -24,16 +24,16 @@ namespace Anvil.Unity.DOTS.Util
         }
         
         /// <summary>
-        /// Creates a new instance of a <see cref="Random"/> and gives it a random seed between 1 and uint.MaxValue
+        /// Creates a new instance of a <see cref="Random"/> and gives it a random seed between 1 and
+        /// <see cref="uint.MaxValue"/>
         /// </summary>
         /// <remarks>
         /// While you could create a new instance yourself, choosing the seed can be annoying. Common approaches are to
         /// use a different random number generator to generate the seed value or use some aspect of the current time.
-        /// The issue with using the current time is that if you create a bunch of <see cref="Random"/> instance at the
-        /// same time in a loop, the seeds will all be the same and the random numbers generated will all be the same.
+        /// Using current time when generating multiple random instances quickly (or in parallel) risks multiple
+        /// instances sharing the same seed value and will therefore generate the same random numbers.
         /// This method uses an internal instance of <see cref="Random"/> to generate a new seed and ensures that it
-        /// fits the range requirements of the seed not being 0. You are guaranteed to have different seeds each time
-        /// this is called.
+        /// fits the range requirements of the seed not being 0.
         /// </remarks>
         /// <returns>A <see cref="Random"/> instance seeded with a random seed.</returns>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
--- a/Scripts/Runtime/Util/Random/RandomUtil.cs
+++ b/Scripts/Runtime/Util/Random/RandomUtil.cs
@@ -1,0 +1,45 @@
+using System.Runtime.CompilerServices;
+using Unity.Mathematics;
+using DateTime = System.DateTime;
+using RuntimeInitializeLoadType = UnityEngine.RuntimeInitializeLoadType;
+
+namespace Anvil.Unity.DOTS.Util
+{
+    /// <summary>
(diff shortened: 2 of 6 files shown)


mbaker3:
> When C#8 is available in Unity (note the version of Unity which ECS currently supports which at the time of writing is 2020.1.9 instead of the latest 2022 version), this could be done once per thread.
https://forum.unity.com/threads/is-it-possible-to-allocate-a-new-temp-nativelist-per-thread-in-entities-foreach.885940/

Let's capture this in an issue: #3
Do you want to add a reference to the issue in code too?

Scripts/Runtime/Util/Random/RandomExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// </summary>
+    public static class RandomExtensions
+    {
+        private const uint SPREAD = uint.MaxValue / JobsUtility.MaxJobThreadCount;
+        
+        /// <summary>
+        /// Given an instance of <see cref="Random"/>, this extension modifies the <see cref="Random"/>'s state so that
+        /// multiple threads using the same initial state will have output different random values based on the thread
+        /// index.
+        /// </summary>
```
mbaker3, line 18:
```suggestion
        /// <summary>
        /// Given an instance of <see cref="Random"/>, this extension modifies the <see cref="Random"/>'s state so that
        /// multiple threads using the same initial state will output different random values based on the thread
        /// index.
        /// </summary>
```

Scripts/Runtime/Util/Random/RandomExtensions.cs
```diff
@@ -0,0 +1,38 @@
+using System.Runtime.CompilerServices;
+using Unity.Jobs.LowLevel.Unsafe;
+using Unity.Mathematics;
+
+namespace Anvil.Unity.DOTS.Util
```
mbaker3, line 5: Should we put this in the `...DOTS.Random` namespace? I could see other random implementations in the future.

Scripts/Runtime/Util/Random/RandomExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+        
+        /// <summary>
+        /// Given an instance of <see cref="Random"/>, this extension modifies the <see cref="Random"/>'s state so that
+        /// multiple threads using the same initial state will have output different random values based on the thread
+        /// index.
+        /// </summary>
+        /// <remarks>
+        /// A typical use case with <see cref="Random"/> is to create an instance on the main thread with a given seed.
+        /// Then pass that instance into a job which will get split across many different threads to do some work that
+        /// requires randomized values.
```
mbaker3, line 22:
```suggestion
        /// A typical use case with <see cref="Random"/> is to create an instance on the main thread with a given seed.
        /// Then pass that instance into a job which will get copied across many different threads to do some work that
        /// requires randomized values.
```

Scripts/Runtime/Util/Random/RandomUtil.cs
```diff
@@ -0,0 +1,45 @@
...
+    /// </summary>
+    public static class RandomUtil
+    {
+        private static Random s_InternalRandom;
+
+        //With DOTS, it's recommended to disable the Domain Reload when entering play mode as it will be pretty slow.
+        //https://docs.unity3d.com/Packages/com.unity.entities@0.17/manual/install_setup.html
+        //Because of this, statics will preserve state across play sessions. By using the RuntimeInitializeOnLoadMethod
+        //we can ensure we reset the state on every play session.
+        //RuntimeInitializeLoadType.SubsystemRegistration represents the earliest moment of all the entries in the enum.
```
mbaker3, line 19:
```suggestion
        // With the Burst compiler, it's common to disable the Domain Reload step when entering play mode to increase iteration time.
        // https://docs.unity3d.com/Packages/com.unity.entities@0.17/manual/install_setup.html
        // Because of this, statics will preserve state across play sessions.
        // `RuntimeInitializeOnLoadMethod` ensures we reset the state on every play session.
        // `RuntimeInitializeLoadType.SubsystemRegistration` represents the earliest moment of all the entries in the enum.
```

mbaker3: Cleaned this up but I don't know if we need to document this as thoroughly all the time. A quick
```
// Facilitates state reset when DomainReload is disabled
// <Link>
```

Is probably enough. Not saying you delete this just something to prevent us writing DOTS tutorials on every implementation!

Scripts/Runtime/Util/Random/RandomUtil.cs
```diff
@@ -0,0 +1,45 @@
...
+        [UnityEngine.RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.SubsystemRegistration)]
+        private static void Init()
+        {
+            s_InternalRandom = new Random((uint)DateTime.Now.Ticks);
+        }
+        
+        /// <summary>
+        /// Creates a new instance of a <see cref="Random"/> and gives it a random seed between 1 and uint.MaxValue
```
mbaker3, line 27:
```suggestion
        /// Creates a new instance of a <see cref="Random"/> and gives it a random seed between 1 and <see cref="uint.MaxValue"/>
```

Scripts/Runtime/Util/Random/RandomUtil.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// </summary>
+        /// <remarks>
+        /// While you could create a new instance yourself, choosing the seed can be annoying. Common approaches are to
+        /// use a different random number generator to generate the seed value or use some aspect of the current time.
+        /// The issue with using the current time is that if you create a bunch of <see cref="Random"/> instance at the
+        /// same time in a loop, the seeds will all be the same and the random numbers generated will all be the same.
```
mbaker3, line 33:
```suggestion
        /// Using current time when generating multiple random instances quickly (or in parallel) risks multiple instances
        /// sharing the same seed value and therefore generate the same random numbers.
```

Scripts/Runtime/Util/Random/RandomUtil.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// This method uses an internal instance of <see cref="Random"/> to generate a new seed and ensures that it
+        /// fits the range requirements of the seed not being 0. You are guaranteed to have different seeds each time
+        /// this is called.
```
mbaker3, line 36:
```suggestion
        /// fits the range requirements of the seed not being 0.
```

That's not a guarantee. Random can generate the same value multiple times in a row.

mbaker3 (approved):
Minor comments. Address as you see fit.

Not to drag this work out further but while you're in the head space can you take a look at the Latios implementation if you haven't already?
 - https://github.com/Dreaming381/Latios-Framework/blob/master/Documentation~/Core/README.md#rng-and-rngtoolkit
 -  https://github.com/Dreaming381/Latios-Framework/blob/master/Documentation%7E/Core/Rng%20and%20RngToolkit.md

Just make sure we aren't overlooking something obvious. From the Latios summary I think I like what we have better but we're not aiming for the determinism required for networked experiences.

Scripts/Runtime/Util/Random/RandomExtensions.cs
jkeon: I tried but it then messes with the imports everywhere because `Random` is now a namespace. 

If we do have other random implementations in the future I think we'll want to rename these classes anyway and we can change it then.

Scripts/Runtime/Util/Random/RandomExtensions.cs
jkeon: Fixed

Scripts/Runtime/Util/Random/RandomExtensions.cs
jkeon: Fixed

Scripts/Runtime/Util/Random/RandomUtil.cs
jkeon: Changed to your suggestion.

Yeah I was on the fence with how much I write in the code vs the PR vs lost forever to our slack history until we pay for it. 

I have found it helpful currently to over-document just to make sure our own understanding is correct and maybe I'm just over compensating for the very sparse documentation that currently exists in Unity's docs.

Scripts/Runtime/Util/Random/RandomUtil.cs
jkeon: Fixed

Scripts/Runtime/Util/Random/RandomUtil.cs
jkeon: Fixed

Scripts/Runtime/Util/Random/RandomUtil.cs
jkeon: I've deleted that line because you are technically correct. 

However, because it's a PRNG, I'm not sure that `NextUInt` could return the same value multiple times in a row since we're using the whole range of 1 to uint.MaxValue. 

If we were doing `NextUInt` from 1 to 2 then sure, we'd collapse down to multiple 1's or 2's in a row. But having the full range and getting the same value multiple times in a row would just have us be in an endless loop no? The state generated that gets fed in returns the same state which therefore gets fed back in and on and on we go.

jkeon:
> > When C#8 is available in Unity (note the version of Unity which ECS currently supports which at the time of writing is 2020.1.9 instead of the latest 2022 version), this could be done once per thread.
> > https://forum.unity.com/threads/is-it-possible-to-allocate-a-new-temp-nativelist-per-thread-in-entities-foreach.885940/
> 
> Let's capture this in an issue: #3 Do you want to add a reference to the issue in code too?

Yep, added to the code as a TODO and referenced this issue.

jkeon:
> Not to drag this work out further but while you're in the head space can you take a look at the Latios implementation if you haven't already?
> 
>     * https://github.com/Dreaming381/Latios-Framework/blob/master/Documentation~/Core/README.md#rng-and-rngtoolkit
> 
>     * https://github.com/Dreaming381/Latios-Framework/blob/master/Documentation%7E/Core/Rng%20and%20RngToolkit.md
> 
> 
> Just make sure we aren't overlooking something obvious. From the Latios summary I think I like what we have better but we're not aiming for the determinism required for networked experiences.

We are essentially doing the same thing but ours is a bit lighter on the housekeeping because we don't care about determinism as much.

For us to support determinism it would be pretty trivial. 

- Instead of calling `RandomUtil.CreateSeeded` we'd want to call `RandomUntil.CreateSeeded("SystemName")`. 
    - Essentially just have each system always request the same seed before hitting it's loop.
- Instead of adjusting the state in each thread via the threadindex, adjust the state in each thread via the entityInQueryIndex. Each entity would then always get the same random numbers.

If we run into needing that, we can add it.

mbaker3 (approved):

Scripts/Runtime/Util/Random/RandomUtil.cs
jkeon: We talked about it and we agree. All good here.

PR #5, opened 2022-03-04
Create mechanism to manage value access dependencies in jobified code

mbaker3:
- Manage access to arbitrary values that are not stored in ECS's chunk system (Ex: Values hosted on systems).
- Provides and maintains job handles to manage safe job access to the value

Create concrete implementation for `NativeArray` (most common use case)

#### Tag Along
 - Added `anvil-csharp-core` and `anvil-unity-core` to the `asmdef`.

Managing values that aren't attached to an Entity is a pain.

Developers can wrap their variable in this type to allow any part of their code to schedule access in a way that's compatible with the Job system.

😞  and 😢 and 😭  when managing access dependencies.

 - None

 - [ ] Yes
 - [x] No

### DIFF
--- a/Scripts/Runtime/Jobs/ValueAccessController/AbstractValueAccessController.cs
+++ b/Scripts/Runtime/Jobs/ValueAccessController/AbstractValueAccessController.cs
@@ -11,7 +11,7 @@ namespace Anvil.Unity.DOTS.Jobs
     /// A wrapper class for managing async access to a value.
     /// </summary>
     /// <remarks>
-    /// Take care that value type implementations propogate state through copies (usually by wrapping a pointer).
+    /// Take care that value type implementations propagate state through copies (usually by wrapping a pointer).
     /// For Example: <see cref="NativeArray{T}" />
     /// </remarks>
     public abstract class AbstractValueAccessController<T> : AbstractAnvilBase
@@ -79,7 +79,7 @@ namespace Anvil.Unity.DOTS.Jobs
         public void ReleaseAsync(JobHandle releaseAccessDependency)
         {
             Debug.Assert(!IsDisposed);
-            Debug.Assert(m_State != AcquisitionState.Unacquired, "There is no outstanding acquisition to realease.");
+            Debug.Assert(m_State != AcquisitionState.Unacquired, "There is no outstanding acquisition to release.");
 
             switch (m_State)
             {
@@ -101,10 +101,12 @@ namespace Anvil.Unity.DOTS.Jobs
         /// <summary>
         /// Get the value of the <see cref="AbstractValueAccessController{T}"/> immediately.
         /// This blocks the calling thread until the value is available.
-        /// <see cref="Release"/> must be called before any other calls to <see cref="AcquireAsync" /> or <see cref="Acquire" /> are made for this value.
+        /// <see cref="Release"/> must be called before any other calls to <see cref="AcquireAsync" /> 
+        /// or <see cref="Acquire" /> are made for this value.
         /// </summary>
         /// <remarks>
-        /// This method and its compliment <see cref="Release" /> are intended to be used for syncronous work on the main thread.
+        /// This method and its compliment <see cref="Release" /> are intended to be used for synchronous work 
+        /// on the main thread.
         /// </remarks>
         public T Acquire(bool isReadOnly)
         {
@@ -123,12 +125,14 @@ namespace Anvil.Unity.DOTS.Jobs
         /// Paired with the use of <see cref="Acquire"/>.
         /// </summary>
         /// <remarks>
-        /// This method can be called after <see cref="AcquireAsync"/> has been called but it will block the calling thread until the value's access dependency is resolved. <see cref="ReleaseAsync" /> is typically the better option.
+        /// This method can be called after <see cref="AcquireAsync"/> has been called but it will 
+        // block the calling thread until the value's access dependency is resolved. <see cref="ReleaseAsync" /> 
+        /// is typically the better option.
         /// </remarks>
         public void Release()
         {
             Debug.Assert(!IsDisposed);
-            Debug.Assert(m_State != AcquisitionState.Unacquired, "There is no outstanding acquisition to realease.");
+            Debug.Assert(m_State != AcquisitionState.Unacquired, "There is no outstanding acquisition to release.");
             m_State = AcquisitionState.Unacquired;
 
             m_ReadAccessDependency.Complete();
--- a/Scripts/Runtime/Jobs/ValueAccessController/AbstractValueAccessController.cs
+++ b/Scripts/Runtime/Jobs/ValueAccessController/AbstractValueAccessController.cs
@@ -1,0 +1,165 @@
+using System;
+using Anvil.CSharp.Core;
+using Unity.Collections;
+using Unity.Entities;
+using Unity.Jobs;
+using UnityEngine;
+
+namespace Anvil.Unity.DOTS.Jobs
+{
+    /// <summary>
+    /// A wrapper class for managing async access to a value.
+    /// </summary>
+    /// <remarks>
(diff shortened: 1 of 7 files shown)


Scripts/Runtime/Jobs/ValueAccessController/AbstractValueAccessController.cs
```diff
@@ -0,0 +1,165 @@
...
+using Unity.Jobs;
+using UnityEngine;
+
+namespace Anvil.Unity.DOTS.Jobs
+{
+    /// <summary>
+    /// A wrapper class for managing async access to a value.
+    /// </summary>
+    /// <remarks>
+    /// Take care that value type implementations propogate state through copies (usually by wrapping a pointer).
```
jkeon, line 14: propogate -> propagate

Scripts/Runtime/Jobs/ValueAccessController/AbstractValueAccessController.cs
```diff
@@ -0,0 +1,165 @@
...
+        {
+            m_Value = initialValue;
+        }
+
+        protected override void DisposeSelf()
+        {
+            // NOTE: If these asserts trigger we should think about calling Complete() on these job handles.
+            Debug.Assert(m_ReadAccessDependency.IsCompleted, "The read access dependency is not completed");
+            Debug.Assert(m_WriteAccessDependency.IsCompleted, "The write access dependency is not completed");
+            (m_Value as IDisposable)?.Dispose();
```
jkeon, line 51: Should we restrict `T` to being `IDisposable`?

Scripts/Runtime/Jobs/ValueAccessController/AbstractValueAccessController.cs
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        /// <summary>
+        /// Schedule the release of the value's ownership after an asynchronous operation.
+        /// </summary>
+        /// <param name="releaseAccessDependency">The JobHandle that describes when work with the value is complete.</param>
+        public void ReleaseAsync(JobHandle releaseAccessDependency)
+        {
+            Debug.Assert(!IsDisposed);
+            Debug.Assert(m_State != AcquisitionState.Unacquired, "There is no outstanding acquisition to realease.");
```
jkeon, line 82: realease -> release

Scripts/Runtime/Jobs/ValueAccessController/AbstractValueAccessController.cs
```diff
@@ -0,0 +1,165 @@
...
+            m_State = AcquisitionState.Unacquired;
+        }
+
+        /// <summary>
+        /// Get the value of the <see cref="AbstractValueAccessController{T}"/> immediately.
+        /// This blocks the calling thread until the value is available.
+        /// <see cref="Release"/> must be called before any other calls to <see cref="AcquireAsync" /> or <see cref="Acquire" /> are made for this value.
+        /// </summary>
+        /// <remarks>
+        /// This method and its compliment <see cref="Release" /> are intended to be used for syncronous work on the main thread.
```
jkeon, line 107: syncronous -> synchronous

Scripts/Runtime/Jobs/ValueAccessController/AbstractValueAccessController.cs
```diff
@@ -0,0 +1,165 @@
...
+        /// Releases access to the value immediately.
+        /// Paired with the use of <see cref="Acquire"/>.
+        /// </summary>
+        /// <remarks>
+        /// This method can be called after <see cref="AcquireAsync"/> has been called but it will block the calling thread until the value's access dependency is resolved. <see cref="ReleaseAsync" /> is typically the better option.
+        /// </remarks>
+        public void Release()
+        {
+            Debug.Assert(!IsDisposed);
+            Debug.Assert(m_State != AcquisitionState.Unacquired, "There is no outstanding acquisition to realease.");
```
jkeon, line 131: realease -> release

Scripts/Runtime/Jobs/ValueAccessController/NativeArrayAccessController.cs
```diff
@@ -0,0 +1,14 @@
+using Unity.Collections;
+using Unity.Entities;
+
+namespace Anvil.Unity.DOTS.Jobs
+{
+    /// <summary>
+    /// A wrapper for NativeArray that provides a way to manage async access to the array.
+    /// </summary>
+    public class NativeArrayAccessController<T> : AbstractValueAccessController<NativeArray<T>> where T : struct, IComponentData
```
jkeon, line 9: Does `T` need to be `IComponentData`? 

What if you want to but in `IBufferElement` or any other `struct` really? I think we just care about making sure it's safe to read/write to the `NativeArray` but what goes in there just needs to be a `struct`

jkeon (dismissed):
Looks good, mostly some typos. 

Couple questions.

Scripts/Runtime/Jobs/ValueAccessController/AbstractValueAccessController.cs
mbaker3: No because not all value types are disposable.

Scripts/Runtime/Jobs/ValueAccessController/NativeArrayAccessController.cs
mbaker3: Yea I don't know why I had that. At one point it fixed an issue but doesn't seem to be required anymore. Removed.

Scripts/Runtime/Jobs/ValueAccessController/AbstractValueAccessController.cs
mbaker3: Fixed

Scripts/Runtime/Jobs/ValueAccessController/AbstractValueAccessController.cs
mbaker3: Fixed

Scripts/Runtime/Jobs/ValueAccessController/AbstractValueAccessController.cs
mbaker3: fixed

Scripts/Runtime/Jobs/ValueAccessController/AbstractValueAccessController.cs
mbaker3: fixed

mbaker3:
Issues addressed and questions answered or addressed.
Ready for review 2

Scripts/Runtime/Jobs/ValueAccessController/AbstractValueAccessController.cs
jkeon: True, but I thought this was for the use of wrapping `NativeXXXX` types? 

Would you ever need a read/write dependency for an `int` for example?

Scripts/Runtime/Jobs/ValueAccessController/NativeArrayAccessController.cs
jkeon: 👍

jkeon (approved):
Looks good, further discussion on one of the comments but if it doesn't spark anything, don't worry about it.

Scripts/Runtime/Jobs/ValueAccessController/AbstractValueAccessController.cs
mbaker3: You could potentially want an `int` wrapped but you'll have to figure out how to deal with the pass by reference issues in the concrete implementation.

In theory you could also use this with managed reference types which wouldn't necessarily be disposable. You couldn't reference the managed type in your job but that managed type may have the data you're passing into your job.

PR #9, opened 2022-03-15
Create MemsetBufferJob

mbaker3:
Sets all elements of a `DynamicBuffer<T>` to a given value.
This is a compliment to Unity's built in `MemsetNativeArray<T>`

New feature

New Feature

None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Entities/MemsetBufferJob.cs
+++ b/Scripts/Runtime/Entities/MemsetBufferJob.cs
@@ -1,0 +1,34 @@
+using Unity.Entities;
+using Unity.Jobs;
+using Unity.Burst;
+using Unity.Collections;
+
+namespace Anvil.Unity.DOTS.Entities
+{
+    /// <summary>
+    /// Sets all elements of a <see cref="DynamicBuffer{T}"/> to a given value.
+    /// </summary>
+    /// <typeparam name="T">The element type of the <see cref="DynamicBuffer{T}"/></typeparam>
+    /// <remarks>The <see cref="DynamicBuffer{T}"/> compliement to <see cref="MemsetNativeArray{T}"/>.</remarks>
+    [BurstCompile]
+    public struct MemsetBufferJob<T> : IJobParallelForBatch where T : struct, IBufferElementData
+    {
+        /// <summary>
+        /// The buffer to write to.
+        /// </summary>
+        [WriteOnly] public BufferFromSingleEntity<T> Source;
+        /// <summary>
+        /// The value to write to the buffer.
+        /// </summary>
+        [ReadOnly] public T Value;
+
+        public void Execute(int startIndex, int count)
+        {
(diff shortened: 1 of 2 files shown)


Scripts/Runtime/Entities/MemsetBufferJob.cs
```diff
@@ -0,0 +1,34 @@
...
+using Unity.Burst;
+using Unity.Collections;
+
+namespace Anvil.Unity.DOTS.Entities
+{
+    /// <summary>
+    /// Sets all elements of a <see cref="DynamicBuffer{T}"/> to a given value.
+    /// </summary>
+    /// <typeparam name="T">The element type of the <see cref="DynamicBuffer{T}"/></typeparam>
+    /// <remarks>The <see cref="DynamicBuffer{T}"/> compliement to <see cref="MemsetNativeArray{T}"/>.</remarks>
```
jkeon, line 12: typo -> compliment

Scripts/Runtime/Entities/MemsetBufferJob.cs
```diff
@@ -0,0 +1,34 @@
...
+        [WriteOnly] public BufferFromSingleEntity<T> Source;
+        /// <summary>
+        /// The value to write to the buffer.
+        /// </summary>
+        [ReadOnly] public T Value;
+
+        public void Execute(int startIndex, int count)
+        {
+            NativeArray<T> buffer = Source.GetBuffer().AsNativeArray();
+            for (int i = startIndex; i < startIndex + count; i++)
```
jkeon, line 28: startIndex + count will get evaluated each iteration.

https://stackoverflow.com/questions/655373/is-the-condition-in-a-for-loop-evaluated-each-iteration

Might as well do a `int len = startIndex + count` and then replace the condition with len

jkeon (changes requested):
Minor changes

jkeon (changes requested):
Not sure what happened, but I had already reviewed?

Scripts/Runtime/Entities/MemsetBufferJob.cs
mbaker3: Huh

The linked example doesn't really address what's going on here but this link reveals that with at least the .NET compiler the value won't get cached because it's not guaranteed to be constant.
https://stackoverflow.com/questions/655373/is-the-condition-in-a-for-loop-evaluated-each-iteration

I thought it could infer that they values would be constant.

Also verified that the add operation remains at the IL level. Not sure if Burst optimizes, I couldn't quite figure out where the check was at the assembly or LLVM level.

mbaker3:
Weird, I didn't re-request a review... I don't know why it pinged you again.

feedback addressed

jkeon (approved):
Odd, but oh well. Looks good!

PR #11, opened 2022-03-24
Feature MemsetBufferJob<T> Batch Calculator

mbaker3:
Add a method to calculate a best effort optimal batch size for `MemsetBufferJob<T>` that aims for minimum real world processing time (not minimum total computation time). If minimum total computation time is desired single thread is almost always the only solution.

The intention of this method is to provide a good default batch size for the job that is pretty good and avoids over provisioning worker threads for small data sets.
**Note:** Frequently a better batch size can be tune for a specific execution/application context but this is better left to optimization passes later in development. If required.

#### Tag Along
 - Documentation omissions and spelling.

Developer is left to define their own batch size.

Developer can call `MemsetBufferJob<T>.ScheduleWithOptimalBatchSize()` to have the batch size calculated for them. Alternatively, developers can call `MemsetBufferJob<T>.CalculateOptimalBatchSize(dataSize)` to get the ideal batch size themselves.

Having to think about defining a reasonable batch size for the workload.

 - None

 - [ ] Yes
 - [x] No - but switching to the calculated batch size is recommended unless there's a reason for the current batch size.

### DIFF
--- a/Scripts/Runtime/Entities/MemsetBufferJob.cs
+++ b/Scripts/Runtime/Entities/MemsetBufferJob.cs
@@ -29,7 +29,7 @@ namespace Anvil.Unity.DOTS.Entities
         /// <summary>
         /// Calculate an ideal batch size per thread.
         /// Aim to spread work across as many threads as possible while satisfying (in order of importance):
-        ///  - Aligning batch size to cache line size
+        ///  - Aligning batch size to chunk size
         ///  - Keeping batch sizes large enough to overcome the cost of splitting.
         ///  - Minimizing the number of batches (there's overhead in each batch run)
         /// </summary>
--- a/Scripts/Runtime/Entities/MemsetBufferJob.cs
+++ b/Scripts/Runtime/Entities/MemsetBufferJob.cs
@@ -22,9 +22,9 @@ namespace Anvil.Unity.DOTS.Entities
         //
         // Ex: We shouldn't split a set of 10 across 5 threads.
         //
-        // Re CacheSize: If generalizing to a 64kb cache size is good enough for Unity it's good 
-        // enough for us.
-        private static readonly float MIN_BATCH_SIZE_PER_THREAD = math.max(1f, JobsUtility.CacheLineSize * 1024 / (float)UnsafeUtility.SizeOf<T>());
+        // Re CacheSize: If generalizing to a 16kb cache size is good enough for Unity it's good 
+        // enough for us. If this isn't good enough #12 will allow us to get the actual cache size.
+        private static readonly float MIN_BATCH_SIZE_PER_THREAD = math.max(1f, 16 * 1024 / (float)UnsafeUtility.SizeOf<T>());
 
         /// <summary>
         /// Calculate an ideal batch size per thread.
--- a/Scripts/Runtime/Entities/MemsetBufferJob.cs
+++ b/Scripts/Runtime/Entities/MemsetBufferJob.cs
@@ -2,6 +2,8 @@ using Unity.Entities;
 using Unity.Jobs;
 using Unity.Burst;
 using Unity.Collections;
+using Unity.Jobs.LowLevel.Unsafe;
+using Unity.Mathematics;
 
 namespace Anvil.Unity.DOTS.Entities
 {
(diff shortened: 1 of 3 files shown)


Scripts/Runtime/Entities/MemsetBufferJob.cs
```diff
@@ -13,6 +16,36 @@ namespace Anvil.Unity.DOTS.Entities
...
     {
+        // The minimum number of elements to process in a single batch.
+        // Attempts to keep batch sizes large enough to prevent multiple cores from copying the same
+        // data into cache. This job's operation is so simple the copy into cache is probably the most expensive part.
+        //
+        // Ex: We shouldn't split a set of 10 across 5 threads.
+        //
+        // Re CacheSize: If generalizing to a 64kb cache size is good enough for Unity it's good 
+        // enough for us.
+        private static readonly float MIN_BATCH_SIZE_PER_THREAD = math.max(1f, JobsUtility.CacheLineSize * 1024 / (float)UnsafeUtility.SizeOf<T>());
```
jkeon, line 27: I don't think this is quite correct. 

The `CacheLineSize` is 64 bytes but we're multiplying arbitrarily by 1024 to get 64 kilobytes. The L1 cache on most CPU's is at least 64 kilobytes but could be different as you and I have discussed. 

If we feel it's important to take up L1 cache we could calculate the real L1 cache size via:
https://stackoverflow.com/questions/6995787/how-to-determine-cpu-cache-size-in-net

But that might be overkill. 

If we feel it should be 64 kilobytes, then I think we should change the calculation to be `64 * 1024` and eliminate the dependency on the `JobsUtility.CacheLineSize`. They aren't really related for what we're doing it's just a happy accident. 

In either case, the comment about 64kb being good enough for Unity isn't quite right because Unity uses 16kb chunk sizes. 

This thread has more info, starting with the direct linked comment: https://forum.unity.com/threads/ok-ecs-works-in-16k-chunks-but-what-is-the-instruction-size-limit-in-ecs.577894/#post-3860092

In the code you can see in `Chunk.kChunkSize` where it's declared as a const of `16 * 1024`. Sadly it's not public because `Chunk` is internal where we can just grab it. 

Looking for `JobsUtility.CacheLineSize` that's used internally by Unity primarily for sizing structs in the collections so that as you iterate over buckets, the data for each bucket occupies it's own cacheline. (I'll confess I'm not 100% sure the reason for this) But it has no relation to Unity `Chunk` size or the actual CPU `L1 Cache` size.

Scripts/Runtime/Entities/MemsetBufferJob.cs
```diff
@@ -13,6 +16,36 @@ namespace Anvil.Unity.DOTS.Entities
...
+
+        /// <summary>
+        /// Calculate an ideal batch size per thread.
+        /// Aim to spread work across as many threads as possible while satisfying (in order of importance):
+        ///  - Aligning batch size to cache line size
```
jkeon, line 32: Based on comment above, might want to revisit this point, if we're trying to align the batches to the `Chunk` or `L1 Cache` size rather than line size.

Scripts/Runtime/Entities/MemsetBufferJob.cs
```diff
@@ -13,6 +16,36 @@ namespace Anvil.Unity.DOTS.Entities
...
+        /// This approach minimizes frame time consumed not total computation time. 
+        /// Single thread will always consume the fewest CPU cycles.
+        /// </remarks>
+        /// <param name="length">The total length of the data set</param>
+        /// <returns>The ideal batch size</returns>
+        public static int CalculateOptimalBatchSize(int length)
+        {
+            float maxBatches = length / MIN_BATCH_SIZE_PER_THREAD;
+            // Spread the max batches across all threads but round up to the nearest MIN_BATCH_SIZE_PER_THREAD.
+            return (int)(math.ceil(maxBatches / (float)JobsUtility.JobWorkerCount) * MIN_BATCH_SIZE_PER_THREAD);
```
jkeon, line 46: I ran into this just recently, but you might want to use `JobsUtility.JobWorkerCount + 1`

The value returned will give you the number of worker threads that you have (9 on your M1 system, 15 on my older intel 8 core) but you also have the main thread which jobs can still be scheduled on. 

Looking further at the actual code for that getter, I was surprised to see that it is a setter too... which seems dangerous? 

But apparently there are some use cases for it albeit much older: https://www.gamedev.net/tutorials/programming/general-and-gameplay-programming/unity-job-system-excessive-multithreading-might-hurt-your-performance-r5258/

jkeon (changes requested):
I approve in principle but I added some info/questions/context that might cause some changes.

Scripts/Runtime/Entities/MemsetBufferJob.cs
mbaker3: Sticking with chunk size for now. Seems to be the best estimate until we feel the need to actually fetch the CPU's cache size.

mbaker3: Fixed comment
```suggestion
        ///  - Aligning batch size to chunk size
```

PR #17, opened 2022-04-12
UnsafeTypedStream Improvements

jkeon:
Added some convenience helpers to get `LaneWriter`s and `LaneReader`s from a `NativeThreadIndex` to handle unity's way of returning thread indexes from 1 to X instead of from 0 to (X-1).

Added a way to specify the number of lanes the collection should use.

Fixed some docs spacing issues.

It did not have these helpers.

It now has these helpers.

None

 - None

 - [ ] Yes 
 - [x] No


mbaker3 (approved):
⛵ ⛵

PR #18, opened 2022-04-19
Busy Work Jobs and DebugUtil

jkeon:
Added two jobs that perform busy work plus a static function in a `DebugUtil` class to access the busy work function anywhere. 

This is useful to create work for the CPU so that you can better understand scheduling problems with jobs or just general profiling.

Currently, trying to add meaningless busy work gets optimized out by the compiler.

Easy, clean way to make the CPU cores work harder.

 - None

 - None

 - [ ] Yes 
 - [ x ] No

### DIFF
--- a/Scripts/Runtime/Jobs/Debug/BusyWorkJob.cs
+++ b/Scripts/Runtime/Jobs/Debug/BusyWorkJob.cs
@@ -10,7 +10,7 @@ namespace Anvil.Unity.DOTS.Jobs
     /// <seealso cref="DebugUtil.FindPrimeNumber"/>
     /// </summary>
     [BurstCompile]
-    public struct BusyWorkJob : IJob
+    public struct BusyWorkJob : IJobFor
     {
         private readonly int m_NthPrimeNumberToFind;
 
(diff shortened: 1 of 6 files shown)


Scripts/Runtime/Jobs/Debug/BusyWorkJob.cs
```diff
@@ -0,0 +1,67 @@
...
+
+namespace Anvil.Unity.DOTS.Jobs
+{
+    /// <summary>
+    /// An <see cref="IJob"/> job that will take up CPU time by calculating the Nth prime number passed in.
+    /// Useful for artificially increasing the time a job takes to help with debugging scheduling or profiling.
+    /// <seealso cref="DebugUtil.FindPrimeNumber"/>
+    /// </summary>
+    [BurstCompile]
+    public struct BusyWorkJob : IJob
```
mbaker3, line 13: Looks good. Might want to consider an `IJobFor` so you can test parallel or single worker using the one job implementation.

mbaker3 (approved):

mbaker3 (approved):

PR #19, opened 2022-04-22
Parallel Collection Util

jkeon:
A nice helper class for sizing collections for writing in parallel and being able to nicely map `NativeThreadIndex` to the index in the collection. 

Currently the defacto way to do this is to size your collection to have 128 buckets and you mostly just fill up 1 to X where X is the number of threads you have. There's a lot of waste there that isn't necessary both from memory usage of the collection and iteration time going through all the buckets.

This didn't exist before so you had the remember the rules and manually size or adjust your indices.

You can just use this helper.

- None

 - None

 - [ ] Yes
 - [x] No

### Tag Along

Updated the `UnsafeTypedStream` to get rid of `FromNativeThreadIndex` functions as they are no longer needed. Cleaner and less maintenance to just use these new utility methods instead.

### DIFF
--- a/Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
+++ b/Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
@@ -119,13 +119,13 @@ namespace Anvil.Unity.DOTS.Util
 #if ENABLE_UNITY_COLLECTIONS_CHECKS
     internal static class ThreadHelper
     {
+        private static ConcurrentBag<int> s_ThreadIndicesSeen = new ConcurrentBag<int>();
+        
         [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.SubsystemRegistration)]
         private static void Init()
         {
             s_ThreadIndicesSeen = new ConcurrentBag<int>();
         }
-
-        private static ConcurrentBag<int> s_ThreadIndicesSeen = new ConcurrentBag<int>();
 
         [BurstDiscard]
         public static void DetectMultipleXThreads(int nativeThreadIndex, int maxSize)
--- a/Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
+++ b/Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
@@ -32,6 +32,7 @@ namespace Anvil.Unity.DOTS.Util
         /// <seealso cref="https://docs.unity3d.com/Packages/com.unity.burst@1.7/manual/docs/AdvancedUsages.html#shared-static"/>
         /// </remarks>
         private static readonly SharedStatic<int> JOB_WORKER_MAXIMUM_COUNT = SharedStatic<int>.GetOrCreate<InternalClassJobWorkerMaximumCount>();
+
         private class InternalClassJobWorkerMaximumCount
         {
         }
@@ -45,7 +46,7 @@ namespace Anvil.Unity.DOTS.Util
             Debug.Assert(JOB_WORKER_MAXIMUM_COUNT.Data > 0);
         }
 
-    
+
         /// <summary>
         /// Returns an ideal size for the number of buckets a collection should have
         /// to account for all possible threads that could write to it at once.
@@ -105,21 +106,27 @@ namespace Anvil.Unity.DOTS.Util
             Debug.Assert(nativeThreadIndex > 0 && nativeThreadIndex <= JobsUtility.MaxJobThreadCount);
             return math.min(nativeThreadIndex - 1, JOB_WORKER_MAXIMUM_COUNT.Data);
         }
-
-
+        
         [Conditional("ENABLE_UNITY_COLLECTIONS_CHECKS")]
         [BurstDiscard]
         private static void DetectMultipleXThreads(int nativeThreadIndex)
         {
             ThreadHelper.DetectMultipleXThreads(nativeThreadIndex, CollectionSizeForMaxThreads);
         }
+
     }
 
-    
+#if ENABLE_UNITY_COLLECTIONS_CHECKS
     internal static class ThreadHelper
     {
-        private static readonly ConcurrentBag<int> s_ThreadIndicesSeen = new ConcurrentBag<int>();
-        
+        [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.SubsystemRegistration)]
+        private static void Init()
+        {
+            s_ThreadIndicesSeen = new ConcurrentBag<int>();
+        }
+
+        private static ConcurrentBag<int> s_ThreadIndicesSeen = new ConcurrentBag<int>();
+
         [BurstDiscard]
         public static void DetectMultipleXThreads(int nativeThreadIndex, int maxSize)
         {
@@ -134,4 +141,5 @@ namespace Anvil.Unity.DOTS.Util
             return output;
         }
     }
+#endif
 }
--- a/Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
+++ b/Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
@@ -1,7 +1,11 @@
-using System;
-using System.Reflection;
+using System.Collections.Concurrent;
+using System.Diagnostics;
+using System.Linq;
+using Unity.Burst;
 using Unity.Jobs.LowLevel.Unsafe;
+using Unity.Mathematics;
 using UnityEngine;
+using Debug = UnityEngine.Debug;
 
 namespace Anvil.Unity.DOTS.Util
 {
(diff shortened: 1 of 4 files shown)


Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
```diff
@@ -0,0 +1,103 @@
...
+        ///
+        /// To avoid this, we have a <see cref="RuntimeInitializeOnLoadMethodAttribute"/> on the
+        /// <see cref="ParallelCollectionUtil.Init"/> method which uses reflection to assign the static readonly
+        /// variable.
+        ///
+        /// While a bit hacky, it is ensured to run before any potential jobs execute. The external call is allowed
+        /// and we get the correct value and inject it into the static readonly field. Subsequent bursted jobs are then
+        /// allowed to access it.
+        /// </remarks>
+        private static readonly int JOB_WORKER_MAXIMUM_COUNT = -1;
```
mbaker3, line 30: Can you add some assertions that this value was actually set? Easier than tracking down the resulting bug if/when this reflected set doesn't work in the future.

Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
```diff
@@ -0,0 +1,103 @@
...
+        /// <see cref="NativeStream"/> has foreachCount
+        /// <see cref="UnsafeTypedStream"/> has lanes
+        /// etc
+        /// It's the number of separate "buckets" that can be written to in parallel.
+        /// </remarks>
+        /// <returns>The number of buckets a collection should have.</returns>
+        public static int CollectionSizeForMaxThreads()
+        {
+            return JOB_WORKER_MAXIMUM_COUNT + 1;
+        }
```
mbaker3, line 55: Should just be a getter or maybe a `public static readonly` value?

Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
```diff
@@ -0,0 +1,103 @@
...
+        /// </remarks>
+        /// <param name="nativeThreadIndex">
+        /// The native thread index to figure out the collection index from. Must be greater than 0.
+        /// </param>
+        /// <returns>The collection index to use</returns>
+        public static int CollectionIndexForThread(int nativeThreadIndex)
+        {
+            // JobsUtility.JobWorkerMaximumCount = 15
+            // index  0, 1, 2, 3, 4, 5, 6, 7, 8, 9,  10, 11, 12, 13, 14, 15
+            // thread 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 17
```
mbaker3, line 97: Remove since the example exists in the comment above?

Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
```diff
@@ -0,0 +1,103 @@
...
+            return (nativeThreadIndex <= JOB_WORKER_MAXIMUM_COUNT)
+                ? nativeThreadIndex - 1
+                : JOB_WORKER_MAXIMUM_COUNT;
```
mbaker3, line 100:
```suggestion
            return math.min(nativeThreadIndex - 1, JOB_WORKER_MAXIMUM_COUNT);
```

Would this work better?

Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
```diff
@@ -0,0 +1,103 @@
...
+        /// <param name="nativeThreadIndex">
+        /// The native thread index to figure out the collection index from. Must be greater than 0.
+        /// </param>
+        /// <returns>The collection index to use</returns>
+        public static int CollectionIndexForThread(int nativeThreadIndex)
+        {
+            // JobsUtility.JobWorkerMaximumCount = 15
+            // index  0, 1, 2, 3, 4, 5, 6, 7, 8, 9,  10, 11, 12, 13, 14, 15
+            // thread 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 17
+            return (nativeThreadIndex <= JOB_WORKER_MAXIMUM_COUNT)
```
mbaker3, line 98: Since we're really just handling a special case should we assert that the index is between `1` and `JOB_WORKER_MAXIMUM_COUNT-1` OR `JOB_WORKER_MAXIMUM_COUNT+1`. That way we're immediately notified on a situation where that assumption changes (Unity update, untested architecture, etc...)

mbaker3 (commented):
Nice changes. Added some recommendations to make sure these hacks don't come back to bite us later....or at least that they'll be easy to identify at the source.

Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
jkeon: Welp, turns out that if Burst is enabled this doesn't work at all. 

Had to change it around to use a `SharedStatic` as per the Burst docs. But yes, I added an assert to catch if it's not set.

Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
jkeon: Can't do that because of when I get to assign `JOB_WORKER_MAXIMUM_COUNT` but I did move to a getter/private setter and init in the `Init` function instead.

Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
jkeon: Ooops, yep, gone.

Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
jkeon: Added an assert to ensure we're between 1 and the highest expected thread index.

Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
jkeon: Yes good call

Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
jkeon: Actually, I changed this because it turns out that this time Unity was returning 16 instead of 17 for the special cases. 

So now we're valid inside Unity's bounds and anything over our bounds we remap to the last bucket. 

But this could be unsafe. What if we get 16 AND 17 AND 28 once? I don't think we can, but I obviously can't prove it. If we do, the Jobs on 16, 17 and 28 will all clobber each other.

So I setup a burst compatible way to make sure we're logging what thread indexes we have seen and ensuring we don't ever go over the number of buckets. 

If this ever gets triggered, then it will be solid evidence to just use the 128 default Unity uses.

Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
```diff
@@ -92,12 +101,37 @@ public static int CollectionSizeForMaxThreads()
...
+        private static void DetectMultipleXThreads(int nativeThreadIndex)
+        {
+            ThreadHelper.DetectMultipleXThreads(nativeThreadIndex, CollectionSizeForMaxThreads);
+        }
+    }
+
+    
+    internal static class ThreadHelper
+    {
+        private static readonly ConcurrentBag<int> s_ThreadIndicesSeen = new ConcurrentBag<int>();
```
mbaker3, line 121: Does this need to get re-initialized on subsystem registration?

Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
```diff
@@ -92,12 +101,37 @@ public static int CollectionSizeForMaxThreads()
...
-            return (nativeThreadIndex <= JOB_WORKER_MAXIMUM_COUNT)
-                ? nativeThreadIndex - 1
-                : JOB_WORKER_MAXIMUM_COUNT;
+            DetectMultipleXThreads(nativeThreadIndex);
+            Debug.Assert(nativeThreadIndex > 0 && nativeThreadIndex <= JobsUtility.MaxJobThreadCount);
+            return math.min(nativeThreadIndex - 1, JOB_WORKER_MAXIMUM_COUNT.Data);
+        }
+
+
+        [Conditional("ENABLE_UNITY_COLLECTIONS_CHECKS")]
```
mbaker3, line 110: Could probably just move this conditional onto the `ThreadHelper` class and its `DetectMultipleXThreads` method then call directly. I don't think this private method adds any value.

mbaker3 (dismissed):
Sorry for the delay. I misunderstood the order of the PRs to be completed.
Looks good. A good take em' or leave em' comments. Up to you.

Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
jkeon: Good point, fixed.

Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
jkeon: I tried but Burst gets mad. 

We're not allowed to call a static method from inside a bursted function. So we call the private function which is BurstDiscarded which is then allowed to call the static function.

mbaker3: Huh, maybe you'd have to bust discard the whole `ThreadHelper` class for it to work. In any case, given the restrictions, this method is fine.

Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
```diff
@@ -0,0 +1,145 @@
...
+#if ENABLE_UNITY_COLLECTIONS_CHECKS
+    internal static class ThreadHelper
+    {
+        [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.SubsystemRegistration)]
+        private static void Init()
+        {
+            s_ThreadIndicesSeen = new ConcurrentBag<int>();
+        }
+
+        private static ConcurrentBag<int> s_ThreadIndicesSeen = new ConcurrentBag<int>();
```
mbaker3, line 128: 🚲 Preferred order?
```suggestion
        private static ConcurrentBag<int> s_ThreadIndicesSeen = new ConcurrentBag<int>();

        [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.SubsystemRegistration)]
        private static void Init()
        {
            s_ThreadIndicesSeen = new ConcurrentBag<int>();
        }
```

mbaker3 (changes requested):

Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
jkeon: Fixed

Scripts/Runtime/Util/Jobs/ParallelCollectionUtil.cs
jkeon: It seems it has to do with how Burst under the hood works. A static class gets created two times. Once for managed land and once for Burst. 

https://forum.unity.com/threads/burst-static-restrictions.536928/#post-4871255

Regardless, `BurstDiscard` only works on methods, not classes :(

mbaker3 (approved):

PR #22, opened 2022-05-02
Add `WorldUtils`

mbaker3:
Add a collection of utilities to manipulate and augment worlds
 - `PlayerLoop` optimization for multi-world applications
 - Move system between groups
 - Add top level system groups to `PlayerLoop`
 - Add additional top level `PlayerLoop` phases `PostInitialization` and `PostUpdate`
 - Bulk system group sorting.

This is an Anvil compliment to `ScriptBehaviourUpdateOrder`

Developer's are left to their own implementations to perform somewhat common tasks on World instances.

Utilities help developers get on with their intent.

Tedium.

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Entities/Util/WorldUtil.cs
+++ b/Scripts/Runtime/Entities/Util/WorldUtil.cs
@@ -15,7 +15,7 @@ namespace Anvil.Unity.DOTS.Entities
     /// </summary>
     /// <remarks>Kept outside <see cref="WorldUtil"/> so their names read better in the editor.</remarks>
     [DisableAutoCreation]
-    internal class EndInitializationCommandBufferSystemGroup : ComponentSystemGroup { }
+    internal class EndInitializationCommandBufferSystemGroup_Anvil : ComponentSystemGroup { }
 
     /// <summary>
     /// A system group to house a <see cref="World"/>s <see cref="EndSimulationEntityCommandBufferSystem"/>.
@@ -23,13 +23,13 @@ namespace Anvil.Unity.DOTS.Entities
     /// </summary>
     /// <remarks>Kept outside <see cref="WorldUtil"/> so their names read better in the editor.</remarks>
     [DisableAutoCreation]
-    internal class EndSimulationCommandBufferSystemGroup : ComponentSystemGroup { }
+    internal class EndSimulationCommandBufferSystemGroup_Anvil : ComponentSystemGroup { }
 
     /// <summary>
     /// A <see cref="PlayerLoop"/> phase inserted immediately after <see cref="Update"/>.
     /// Add via <see cref="WorldUtil.AddCustomPhasesToCurrentPlayerLoop" />.
     /// </summary>
-    internal class PostUpdate
+    internal class PostUpdate_Anvil
     {
         /// <summary>
         /// Provides a <see cref="PlayerLoopSystem" /> instance to to identify this phase.
@@ -38,7 +38,7 @@ namespace Anvil.Unity.DOTS.Entities
         {
             get => new PlayerLoopSystem()
             {
-                type = typeof(PostUpdate)
+                type = typeof(PostUpdate_Anvil)
             };
         }
     }
@@ -47,7 +47,7 @@ namespace Anvil.Unity.DOTS.Entities
     /// A <see cref="PlayerLoop"/> phase inserted immediately after <see cref="Initialization"/>.
     /// Add via <see cref="WorldUtil.AddCustomPhasesToCurrentPlayerLoop" />.
     /// </summary>
-    internal class PostInitialization
+    internal class PostInitialization_Anvil
     {
         /// <summary>
         /// Provides a <see cref="PlayerLoopSystem" /> instance to to identify this phase.
@@ -56,7 +56,7 @@ namespace Anvil.Unity.DOTS.Entities
         {
             get => new PlayerLoopSystem()
             {
-                type = typeof(PostInitialization)
+                type = typeof(PostInitialization_Anvil)
             };
         }
     }
@@ -73,7 +73,7 @@ namespace Anvil.Unity.DOTS.Entities
 
         /// <summary>
         /// Add custom phases to the <see cref="PlayerLoop"/>.
-        /// <see cref="PostInitialization"/> and <see cref="PostUpdate"/>.
+        /// <see cref="PostInitialization_Anvil"/> and <see cref="PostUpdate_Anvil"/>.
         /// </summary>
         public static void AddCustomPhasesToCurrentPlayerLoop()
         {
@@ -88,13 +88,13 @@ namespace Anvil.Unity.DOTS.Entities
 
             int initializationPhaseIndex = topLevelPhases.FindIndex((phase) => phase.type == typeof(Initialization));
             Debug.Assert(initializationPhaseIndex != -1, $"{nameof(Initialization)} phase not found");
-            Debug.Assert(!topLevelPhases.Any((phase) => phase.type == typeof(PostInitialization)), $"{nameof(PostInitialization)} phase already added");
-            topLevelPhases.Insert(initializationPhaseIndex + 1, PostInitialization.PlayerLoopSystem);
+            Debug.Assert(!topLevelPhases.Any((phase) => phase.type == typeof(PostInitialization_Anvil)), $"{nameof(PostInitialization_Anvil)} phase already added");
+            topLevelPhases.Insert(initializationPhaseIndex + 1, PostInitialization_Anvil.PlayerLoopSystem);
 
             int updatePhaseIndex = topLevelPhases.FindIndex((phase) => phase.type == typeof(Update));
             Debug.Assert(updatePhaseIndex != -1, "Update phase not found");
-            Debug.Assert(!topLevelPhases.Any((phase) => phase.type == typeof(PostUpdate)), $"{nameof(PostUpdate)} phase already added");
-            topLevelPhases.Insert(updatePhaseIndex + 1, PostUpdate.PlayerLoopSystem);
+            Debug.Assert(!topLevelPhases.Any((phase) => phase.type == typeof(PostUpdate_Anvil)), $"{nameof(PostUpdate_Anvil)} phase already added");
+            topLevelPhases.Insert(updatePhaseIndex + 1, PostUpdate_Anvil.PlayerLoopSystem);
 
             playerLoop.subSystemList = topLevelPhases.ToArray();
             PlayerLoop.SetPlayerLoop(playerLoop);
@@ -110,7 +110,7 @@ namespace Anvil.Unity.DOTS.Entities
         /// A collection of value pairs where the first value is the PlayerLoop phase to place the system in and the second value is the 
         /// system to create.
         /// </param>
-        /// <returns></returns>
+        /// <returns>A collection of the system groups created.</returns>
         /// <remarks>
         /// Groups must not already exist in the world. This is a limitation of our ability to detect whether a system has already been 
         /// added to the palyer loop.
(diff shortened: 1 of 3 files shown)


Scripts/Runtime/Entities/Util/WorldUtil.cs
```diff
@@ -0,0 +1,230 @@
...
+    [DisableAutoCreation]
+    internal class EndInitializationCommandBufferSystemGroup : ComponentSystemGroup { }
+
+    /// <summary>
+    /// A system group to house a <see cref="World"/>s <see cref="EndSimulationEntityCommandBufferSystem"/>.
+    /// Used for mutli-world playerloop optimization.
+    /// </summary>
+    /// <remarks>Kept outside <see cref="WorldUtil"/> so their names read better in the editor.</remarks>
+    [DisableAutoCreation]
+    internal class EndSimulationCommandBufferSystemGroup : ComponentSystemGroup { }
```
jkeon, line 26: Minor but I'm wondering if there's value in having these names have something in them to denote that they are custom? Or at least not Unity SystemGroups? 

Right now if I saw these in the Entities Debugger or Profiler, I would think they were default Unity groups. 

Maybe `AnvilEndInitializationCommandBufferSystemGroup` etc?

Scripts/Runtime/Entities/Util/WorldUtil.cs
```diff
@@ -0,0 +1,230 @@
...
+
+    /// <summary>
+    /// A <see cref="PlayerLoop"/> phase inserted immediately after <see cref="Update"/>.
+    /// Add via <see cref="WorldUtil.AddCustomPhasesToCurrentPlayerLoop" />.
+    /// </summary>
+    internal class PostUpdate
```
jkeon, line 32: Similar naming question if we should denote between what Unity gives you by default and what we've added custom.

Scripts/Runtime/Entities/Util/WorldUtil.cs
```diff
@@ -0,0 +1,230 @@
...
+    {
+        /// <summary>
+        /// Provides a <see cref="PlayerLoopSystem" /> instance to to identify this phase.
+        /// </summary>
+        public static PlayerLoopSystem PlayerLoopSystem
```
jkeon, line 37: Similar naming question if we should denote between what Unity gives you by default and what we've added custom.

Scripts/Runtime/Entities/Util/WorldUtil.cs
```diff
@@ -0,0 +1,230 @@
...
+                type = typeof(PostUpdate)
+            };
+        }
+    }
+
+    /// <summary>
+    /// A <see cref="PlayerLoop"/> phase inserted immediately after <see cref="Initialization"/>.
+    /// Add via <see cref="WorldUtil.AddCustomPhasesToCurrentPlayerLoop" />.
+    /// </summary>
+    internal class PostInitialization
```
jkeon, line 50: Similar naming question if we should denote between what Unity gives you by default and what we've added custom.

Scripts/Runtime/Entities/Util/WorldUtil.cs
```diff
@@ -0,0 +1,230 @@
...
+        }
+    }
+
+    /// <summary>
+    /// A collection of utilities to manipulate and augments <see cref="World"/>s.
+    /// The Anvil compliment to <see cref="ScriptBehaviourUpdateOrder"/>.
+    /// </summary>
+    public static class WorldUtil
+    {
+        // Not need to reset between play sessions because PlayerLoop systems are stateless and
```
jkeon, line 70:
```suggestion
        // No need to reset between play sessions because PlayerLoop systems are stateless and 
```

Scripts/Runtime/Entities/Util/WorldUtil.cs
```diff
@@ -0,0 +1,230 @@
...
+            {
+                return;
+            }
+
+            // Create new top level player loop phases
+            PlayerLoopSystem playerLoop = PlayerLoop.GetCurrentPlayerLoop();
+            List<PlayerLoopSystem> topLevelPhases = playerLoop.subSystemList.ToList();
+
+            int initializationPhaseIndex = topLevelPhases.FindIndex((phase) => phase.type == typeof(Initialization));
+            Debug.Assert(initializationPhaseIndex != -1, "Initialization phase not found");
```
jkeon, line 90:
```suggestion
            Debug.Assert(initializationPhaseIndex != -1, $"{typeof(Initialization)} phase not found");
```

Scripts/Runtime/Entities/Util/WorldUtil.cs
```diff
@@ -0,0 +1,230 @@
...
+            Debug.Assert(!topLevelPhases.Any((phase) => phase.type == typeof(PostInitialization)), $"{nameof(PostInitialization)} phase already added");
+            topLevelPhases.Insert(initializationPhaseIndex + 1, PostInitialization.PlayerLoopSystem);
+
+            int updatePhaseIndex = topLevelPhases.FindIndex((phase) => phase.type == typeof(Update));
+            Debug.Assert(updatePhaseIndex != -1, "Update phase not found");
+            Debug.Assert(!topLevelPhases.Any((phase) => phase.type == typeof(PostUpdate)), $"{nameof(PostUpdate)} phase already added");
+            topLevelPhases.Insert(updatePhaseIndex + 1, PostUpdate.PlayerLoopSystem);
```
jkeon, line 97: Seems like this and lines 89 to 92 could be made into a function where you pass in the types?

Scripts/Runtime/Entities/Util/WorldUtil.cs
```diff
@@ -0,0 +1,230 @@
...
+
+        /// <summary>
+        /// Create a collection of top level <see cref="ComponentSystemGroup" />s and add them to the current <see cref="PlayerLoop"/>.
+        /// </summary>
+        /// <param name="world">The world to create top level groups in.</param>
+        /// <param name="topLevelGroupTypes">
+        /// A collection of value pairs where the first value is the PlayerLoop phase to place the system in and the second value is the 
+        /// system to create.
+        /// </param>
+        /// <returns></returns>
```
jkeon, line 113: Forgotten doc?

Scripts/Runtime/Entities/Util/WorldUtil.cs
```diff
@@ -0,0 +1,230 @@
...
+
+            // Create and Add top level system groups to the player loop
+            ComponentSystemGroup[] topLevelGroups = new ComponentSystemGroup[topLevelGroupTypes.Length];
+            for (int i = 0; i < topLevelGroupTypes.Length; i++)
+            {
+                (Type playerLoopSystemType, Type systemGroupType) = topLevelGroupTypes[i];
+                Debug.Assert(systemGroupType.IsSubclassOf(typeof(ComponentSystemGroup)));
+
+                // If we had access to ScriptBehaviourUpdateOrder.DummyDelegateWrapper we could detect if the system has already been added to the PlayerLoop.
+                Debug.Assert(world.GetExistingSystem(systemGroupType) == null, $"System group cannot already exists in world {world.Name}. There is no way of knowing if a top level group for an individual world has been added to the player loop already.");
```
jkeon, line 130:
```suggestion
                Debug.Assert(world.GetExistingSystem(systemGroupType) == null, $"System group cannot already exist in world {world.Name}. There is no way of knowing if a top level group for an individual world has been added to the player loop already.");
```

Scripts/Runtime/Entities/Util/WorldUtil.cs
```diff
@@ -0,0 +1,230 @@
...
+
+        /// <summary>
+        /// Sort all groups in a <see cref="World"/>.
+        /// </summary>
+        public static void SortAllGroupsInWorld(World world)
+        {
+            // This implementation is a bit overkill since calling ComponentSystemGroup.SortSystems() calls recursively down to subgroups
+            // but the way ScriptBehaviourUpdateOrder adds systems to the PlayerLoop prevents us from inspecting instances and any logic 
+            // to calculate which systems are top level (check if not a child of other systems) is likely more expensive than just overcalling SortSystems().
+            // SortSystems exits early of no sorting is required.
```
jkeon, line 149:
```suggestion
            // SortSystems exits early if no sorting is required.
```

Scripts/Runtime/Entities/Util/WorldUtil.cs
```diff
@@ -0,0 +1,230 @@
...
+        /// <summary>
+        /// Optimizes a <see cref="World"/>'s <see cref="ComponentSystemGroup"/>s for multi-world applications.
+        /// Groups end command buffers into their own <see cref="PlayerLoop"/> phase just after their default phase. 
+        /// (Ex: <see cref="Update"/> -> <see cref="PostUpdate"/>)
+        /// This allows end command buffers for all worlds to be evaluated after all worlds have scheduled their work for the phase.
+        /// The result is that other worlds can compute their jobified work while one world is executing its end command buffer on the main thread.
+        /// </summary>
+        /// <param name="world">The world instance to optimize.</param>
+        /// <remarks>
+        /// This method doesn't move <see cref="EndFixedStepSimulationEntityCommandBufferSystem" because it's embedded in the <see cref="SimulationSystemGroup"/>.
```
jkeon, line 169:
```suggestion
        /// This method doesn't move <see cref="EndFixedStepSimulationEntityCommandBufferSystem"/> because it's embedded in the <see cref="SimulationSystemGroup"/>.
```

Scripts/Runtime/Entities/Util/WorldUtil.cs
```diff
@@ -0,0 +1,230 @@
...
+        /// making it impractical to group with the other worlds. There shouldn't be any expensive work in that command buffer anyway.
+        /// </remarks>
+        public static void OptimizeWorldForMultiWorldInCurrentPlayerLoop(World world)
```
jkeon, line 172: This is a super smart way to handle this. I really like the thought in knowing we have main thread work we must do but figuring out how to arrange it so that everyone else gets a chance to load up the workers before the main thread work goes on.

Scripts/Runtime/Entities/Util/WorldUtil.cs
```diff
@@ -0,0 +1,230 @@
...
+            ComponentSystemBase system = world.GetExistingSystem<System>();
+            ComponentSystemGroup srcGroup = world.GetExistingSystem<SrcGroup>();
+            ComponentSystemGroup destGroup = world.GetExistingSystem<DestGroup>();
+
+            // Skip if there are missing elements
+            if (system == null || srcGroup == null || destGroup == null)
+            {
+                throw new ArgumentNullException($"{nameof(MoveSystemFromToGroup)}: One or more of the provided system and groups do not exist. {nameof(System)}:{system}, {nameof(SrcGroup)}:{srcGroup}, {nameof(DestGroup)}:{destGroup}");
+            }
+
```
jkeon, line 223: Should we add a check to ensure that `destGroup` and `srcGroup` are not the same?

jkeon (approved):
Overall approved, looks great!

A bunch of minor things as you see fit.

Scripts/Runtime/Entities/Util/WorldUtil.cs
mbaker3: Sort of, In this case they're both "AddAfter X" operations but there's any number of ways to decide where a phase lands. It didn't feel necessary to create a method just because these two phases happened to get positioned in the same way.

Scripts/Runtime/Entities/Util/WorldUtil.cs
mbaker3: edited to `nameof` instead of `typeof`...you still get the credit :)

Scripts/Runtime/Entities/Util/WorldUtil.cs
mbaker3: yea, I guess I can debug assert that. Really, it's not a big deal if your source and destination systems are the same. You're just doing work for nothing.

Scripts/Runtime/Entities/Util/WorldUtil.cs
mbaker3: Good point. I added the `_Anvil` as a suffix to all of these.
That way it's easy to identify as non-default but you don't constantly have to parse the word Anvil when hunting for things in the `PlayerLoop` list. Much easier to read and recognize the group for what it really is.

Scripts/Runtime/Entities/Util/WorldUtil.cs
mbaker3: addressed

Scripts/Runtime/Entities/Util/WorldUtil.cs
mbaker3: addressed

Scripts/Runtime/Entities/Util/WorldUtil.cs
mbaker3: addressed

Scripts/Runtime/Entities/Util/WorldUtil.cs
mbaker3: Good catch, fixed.

Scripts/Runtime/Entities/Util/WorldUtil.cs
mbaker3: fixed

mbaker3:
Issues addressed. Thanks for the review

Scripts/Runtime/Entities/Util/WorldUtil.cs
jkeon: Cool perfect

Scripts/Runtime/Entities/Util/WorldUtil.cs
jkeon: My reasoning was that maybe there would be an issue with OnCreated, OnDestroyed, OnStartRunning, OnStopRunning etc being called that we didn't want. 

But you added the assert so all good.

jkeon (approved):

Scripts/Runtime/Entities/Util/WorldUtil.cs
jkeon: Sounds good

PR #25, opened 2022-05-10
Entities 0.50 upgrade

mbaker3:
Upgrades required for Entities 0.50 compatibility.

No Entities 0.50 support

Entities 0.50 support

Incompatibility with Entities 0.50

 - None

 - [x] Yes - Not backwards compatible. Stay on the [entities-0.17 tag](https://github.com/decline-cookies/anvil-unity-dots/releases/tag/entities-0.17) if you don't want to upgrade to Entities 0.50
 - [ ] No

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,14 +12,15 @@ The code is reasonably clean but documentation and examples are sparse. Feel fre
 ⚠️ We welcome PRs and bug reports but making this repo a public success is not our priority. No promises on when it will be addressed!
 
 # Dependencies
-- [Unity (2021.1.9)](https://unity.com/)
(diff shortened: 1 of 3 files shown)


jkeon:
What's our plan with this and the outstanding PR's/Branches? 

Should all those PRs get merged into main first and then we upgrade?

mbaker3:
I say we merge this into main and then bring all the other branches up to date.
The owners of the outstanding branches are going to be more familiar with upgrade changes than I will be.

It helps that all of my open branches are all based off the head of this branch :)

jkeon (approved):
As discussed, we're upgrading.

PR #26, opened 2022-05-10
Add `NativeArrayExtension`

mbaker3:
Add `NativeArrayExtension`. A collection of extension methods to help developers work with `NativeArray<T>`s

#### Tag Alongs
 - Updated `RandomExtension` to match anvil extension class convention.
 - `CopyFromSingletonBufferJob` - Fixed spelling mistake in docs.

There are no existing extension methods provided by Anvil.

Developers now have access to `FloodClear()` and `FloodSet()` methods to clear and set the value of a range of elements in a `NativeArray<T>` efficiently.

Some quick tests show that this implementation is ~8x faster than using a `MemsetNativeArray` job.
```
Rough metrics on an array of 262,144 byte elements (ticks)
 - MemsetNativeArray.Schedule: 868
 - FloodSet: 103
 - FloodClear: 72
```
If needed, in the future, we could create a job that executes these more efficient methods so that multiple clears/sets could be done in parallel but we'll build that feature if/when the need comes up. You'd have to be setting values on many truly massive arrays at once to see benefits.

**Discussion:** Open to method naming suggestions. `Flood*` works but I'm not 💯  on it.

Convenience and performance in flood filling `NativeArray<T>`s

 - None

 - [ ] Yes 
 - [x] No


Scripts/Runtime/Collections/NativeArrayExtension.cs
```diff
@@ -0,0 +1,62 @@
+using UnityEngine;
+using Unity.Collections;
+using Unity.Collections.LowLevel.Unsafe;
+
+
+/// <summary>
+/// Extension methods for use with a <see cref="NativeArray{T}"/> instance.
+/// </summary>
+public static class NativeArrayExtension
```
jkeon, line 9: Is it `Extension` or `Extensions`?

We seem to be using both? Doesn't matter to me, just clarifying.

Scripts/Runtime/Util/Random/RandomExtension.cs
```diff
@@ -0,0 +1,36 @@
+using System.Runtime.CompilerServices;
+using Unity.Jobs.LowLevel.Unsafe;
+using Unity.Mathematics;
+
+/// <summary>
+/// Extension methods for use with a <see cref="Random"/> instance.
+/// </summary>
+public static class RandomExtension
```
jkeon, line 8: Is it `Extension` or `Extensions`?

We seem to be using both? Doesn't matter to me, just clarifying.

jkeon (commented):
Looks good, just waiting on the resolution of https://github.com/decline-cookies/anvil-csharp-core/pull/82 to approve.

I see what you mean on Flood but I can't think of anything else good and I think this works.
Makes me think of FloodFill from Photoshop and the analogy works well.

Scripts/Runtime/Collections/NativeArrayExtension.cs
mbaker3: Where are we using Extensions? We aren't as far as I know.
Some 3rd party libraries that we depend on do add the `s`

Scripts/Runtime/Util/Random/RandomExtension.cs
mbaker3: See other comment for discussion

Scripts/Runtime/Collections/NativeArrayExtension.cs
```diff
@@ -0,0 +1,62 @@
+using UnityEngine;
+using Unity.Collections;
+using Unity.Collections.LowLevel.Unsafe;
+
+
```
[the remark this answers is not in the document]
mbaker3: Will move to `namespace Anvil.Unity.DOTS.Collections`

Scripts/Runtime/Util/Random/RandomExtension.cs
```diff
@@ -0,0 +1,36 @@
+using System.Runtime.CompilerServices;
+using Unity.Jobs.LowLevel.Unsafe;
+using Unity.Mathematics;
+
```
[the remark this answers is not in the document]
mbaker3: Will move to `namespace Anvil.Unity.DOTS.Collections`

mbaker3:
Merging in. Will deal with namespace fixes in a separate PR.

PR #30, opened 2022-05-17
Logging - Add Burst Compatible Logger

mbaker3:
Add a burst compatible logger that can be brought into bursted contexts.

`Log.Logger` is not burst compatible so the developer can only emit log messages via `Debug.Log` and friends.

#### Tag Alongs
 - Add a lazy initialized `Logger` property to `AbstractAnvilSystemBase`. The same pattern used in `AbstractAnvilBase`

A `BurstableLogger` instance can be created either manually (within a bursted or managed context) or derived from a `Log.Logger` using the `.AsBurstable()` extension method.

At the moment the Burstable logger mainly:
 - Provides a familiar `Log.Logger` API
 - Allows for a message prefix to be defined which is prepended to all messages emitted through the logger.

However, this setup lays the groundwork for more features as they are added to the managed `Log.Logger` and as Burst adds more features to allow for more robust logging.

Allowing familiar `Log.Logger` functionality and API within a Burst compiled context.
Provides a wrapper that can be used to further decorate log messages emitted from burst compiled code with minimal updates at the call site.

 - https://github.com/decline-cookies/anvil-csharp-core/pull/84
 - https://github.com/decline-cookies/anvil-unity-core/pull/62

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Logging/BurstableLogger.cs
+++ b/Scripts/Runtime/Logging/BurstableLogger.cs
@@ -1,0 +1,87 @@
+using System;
+using Anvil.CSharp.Logging;
+using Unity.Collections;
+
+namespace Anvil.Unity.DOTS.Logging
+{
+    [BurstCompatible]
+    public readonly struct BurstableLogger<PrefixStringType> where PrefixStringType : struct, INativeList<byte>, IUTF8Bytes
+    {
+        public readonly PrefixStringType MessagePrefix;
+
+        [NotBurstCompatible]
+        public BurstableLogger(Log.Logger logger, string appendToMessagePrefix)
+        {
+            string managedMessagePrefix = logger.MessagePrefix + appendToMessagePrefix;
+            MessagePrefix = default;
+            CopyError error = FixedStringMethods.CopyFrom(ref MessagePrefix, managedMessagePrefix);
+
+            switch (error)
+            {
+                case CopyError.None:
+                    break;
+
+                case CopyError.Truncation:
+                    logger.Error($"The message prefix is too long for this {nameof(BurstableLogger<PrefixStringType>)} and will be truncated (MaxLength:{MessagePrefix.Capacity} PrefixLength:{managedMessagePrefix.Length}). Select a different larger prefix string type for this logger.");
+                    break;
+
+                default:
+                    throw new Exception("Unknown message prefix error: " + error);
+            }
+        }
+
+        public BurstableLogger(in PrefixStringType messagePrefix)
+        {
+            MessagePrefix = messagePrefix;
+        }
+
+        public void Debug(in FixedString64Bytes message)
+        {
+            Debug<FixedString64Bytes>(message);
+        }
+
+        public void Debug<T>(in T message) where T : struct, INativeList<byte>, IUTF8Bytes
+        {
+            AssertMessageLength(message);
+            UnityEngine.Debug.Log($"{MessagePrefix}{message}");
+        }
+
+        public void Warning(in FixedString64Bytes message)
+        {
+            Warning<FixedString64Bytes>(message);
+        }
+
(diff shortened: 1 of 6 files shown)


Scripts/Runtime/Logging/LoggerExtension.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// <see cref="Log.Logger" /> instance.
+        /// </summary>
+        /// <typeparam name="PrefixStringType"></typeparam>
+        /// <param name="logger">The <see cref="Log.Logger" /> to copy configuration from.</param>
+        /// <param name="appendToMessagePrefix">
+        /// A string to append to the <see cref="Log.Logger" />'s existing prefix.
+        /// (max length: <see cref="PrefixStringType.Capacity"/>)
+        /// </param>
+        /// <returns>A <see cref="BurstableLogger{PrefixStringType}" /> instance.</returns>
+        public static BurstableLogger<PrefixStringType> AsBurstable<PrefixStringType>(this in Log.Logger logger, string appendToMessagePrefix = null) where PrefixStringType : struct, INativeList<byte>, IUTF8Bytes
```
jkeon, line 37:
```suggestion
        public static BurstableLogger<TPrefixStringType> AsBurstable<PrefixStringType>(this in Log.Logger logger, string appendToMessagePrefix = null) where TPrefixStringType : struct, INativeList<byte>, IUTF8Bytes
```

jkeon (changes requested):
Looks good, one minor generic naming refactor.

Scripts/Runtime/Logging/BurstableLogger.cs
```diff
@@ -0,0 +1,182 @@
...
+    /// In the future, the goal of this type is to provide all the same information that 
+    /// <see cref="Log.Logger" /> provides. That is not possible with Burst today but should be 
+    /// possible in the future without making any changes at the consuming end. At the moment, the 
+    /// contextual information is added when the message passes through <see cref="UnityLogListener"/>.
+    /// This information will appear in the output of all active <see cref="ILogHandler" /> implementations.
+    /// except for the Editor console (<see cref="UnityLogHandler"/>).
+    /// This is a Burst limitation.
+    /// </remarks>
+    [BurstCompatible]
+    public readonly struct BurstableLogger<PrefixStringType> where PrefixStringType : struct, INativeList<byte>, IUTF8Bytes
```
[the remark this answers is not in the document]
mbaker3: Fixed

Scripts/Runtime/Logging/LoggerExtension.cs
mbaker3: Fixed

Scripts/Runtime/Logging/BurstableLogger.cs
```diff
@@ -46,6 +52,7 @@ namespace Anvil.Unity.DOTS.Logging
         [NotBurstCompatible]
         public BurstableLogger(Log.Logger logger, string appendToMessagePrefix)
         {
+            m_ThreadIndex = UNSET_THREAD_INDEX;
```
jkeon, line 55: This is probably just being super pedantic but I'll mention it anyway. 

We should probably have line 39 set the thread index.

`private readonly int m_ThreadIndex = UNSET_THREAD_INDEX;`

Today, it will only get set to the `UNSET_THREAD_INDEX` if you use the specific constructors. 

Any `BurstableLogger` that is `default` or created in bulk for an array will use the default constructor and the thread index will be 0. 

The odds are super low this would happen and if it did then you miss out on your message prefix anyway but still, could be a very random weird bug in a specific case where you wanted it to run on the main thread without a prefix and now your log is telling you that you're off thread on thread 0. 

Bonus, you don't have to assign the `m_ThreadIndex` on line 80 anymore either.

jkeon (approved):
I'm approving because it's fine, but one comment that's worth doing a quick fix for.

PR #33, opened 2022-05-20
Refactoring Access Controller

### LINKED ISSUE #31: Remove Static Lookup/Management Collection Access Instances
Now that we've been able to take a step back and see how all the parts of the new access controller system work we can see that it doesn't make sense to have a global (or world scope) lookup of collection access controllers.

This functionality was a holdover from when we were trying to manage the access of data that was held in both entities and one off data stores (Systems, Monobehaviours, etc..).

By removing the global lookup we:
 - Simplify usage
 - Remove any dependency on `Unity.Entities`
 - Reduce the amount of code

jkeon:
Refactoring of the `CollectionAccessController` into `AbstractAccessController`

- Static lookups
- Assumption that we're wrapping a collection
- Not hiding the data inside the access controller

- New `AbstractAccessController` that provides better functionality for owning the data in subclasses.
- Added simple `AccessController<T>` for wrapping a singular element of data (NativeArray etc)
- Removed `AbstractValueAccessController`, it is no longer needed.
- Removed `NativeArrayAccessController`, this can be accomplished via `AccessController<NativeArray<T>>` instead. 

See: https://github.com/decline-cookies/twisted-dots/commit/0f4967bebe0a448008a546893df8add015d1face for how the `TimerSubSystem` was reworked to hold a new subclass called `TimerAccessController` which wraps the three different lists of Timers. 

Resolves #31 

 - None

 - [x] Yes - API changes, but no one else is using this right now.
 - [ ] No


Scripts/Runtime/Jobs/AccessControl/AccessController.cs
```diff
@@ -0,0 +1,35 @@
+using System;
+
+namespace Anvil.Unity.DOTS.Jobs
+{
+    /// <summary>
+    /// A concrete implementation of <see cref="AbstractAccessController{T}"/> that assumes
+    /// wrapping a singular unchanging piece of data which is the most common case.
+    /// </summary>
+    /// <typeparam name="T">The type of data to wrap access control around.</typeparam>
+    public class AccessController<T> : AbstractAccessController<T>
```
mbaker3, line 10: Proposal:
 - Remove `<T>` and value delivery from `AbstractAccessController`
 - Consider making `AccessController` non-abstract
 - Rename `AccessController<T>` to `AccessControlledValue<T>` and compose in the `AccessController`
 - For any more pieces of data that a developer wants to wrap into one control block they can either:
    - Wrap their data in an object and use `AccessControlledValue<T>`
    - Create their own class that composes in `AccessController`. With this option we'll just have to manually ensure the APIs stay as similar as possible between implementations. Docs should probably suggest that the `AccessControlledValue<T>` API is matched as closely as possible.

jkeon: As discussed, yes this is a good idea.

All changes made. 
https://github.com/decline-cookies/twisted-dots/commit/7aee552653a95a1fdc6ed857f863e30399636709 also updated to reflect the new changes.

mbaker3 (approved):
Great change!

PR #34, opened 2022-05-20
Adding clear ability to UnsafeTypedStream plus bug fixes

jkeon:
Adding a Clear function the `UnsafeTypedStream` 
Plus a few bug fixes

- Currently you have to dispose and create a new `UnsafeTypedStream` anytime you want to reuse a variable each frame. 
- `UnsafeTypedStream` uses the same allocator for the structure itself and the content.
- Safety checks in `AccessController` didn't go far enough in the stack frame to get the right calling function.

- Now there is a `Clear` function which either clears immediately on the calling thread or can be scheduled in a job to clear the data but keep the structure intact allowing it to be reused.
- `UnsafeTypedStream` can optionally have different types of memory allocation. One for the structure itself and one for the content. 
    - This allows for a `UnsafeTypedStream` to be created using `Persistent` memory but then for all the content to be `TempJob` memory which results in much faster allocations.
- Safety checks in `AccessController` now go far enough in the stack frame to get the right calling function.

- None

 - None

 - [ ] Yes
 - [x] No

### Tag-Alongs

- Added a `CollectionIndexForMainThread`  function to `ParallelAccessUtil` 
- Added a `ConsolidateToNativeArrayJob` to turn `UnsafeTypedStream<T>` into `NativeArray<T>` via a `DeferredNativeArray`

### DIFF
--- a/Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
+++ b/Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
@@ -171,6 +171,7 @@ namespace Anvil.Unity.DOTS.Data
         /// <summary>
         /// Disposes the collection
         /// </summary>
+        [WriteAccessRequired]
         public void Dispose()
         {
             if (!IsCreated)
@@ -180,10 +181,32 @@ namespace Anvil.Unity.DOTS.Data
 
             Allocator allocator = m_BufferInfo->Allocator;
 
+            Clear();
+
+            UnsafeUtility.Free(m_Lanes, allocator);
+            UnsafeUtility.Free(m_BufferInfo, allocator);
+
+            m_Lanes = null;
+            m_BufferInfo = null;
+        }
+        
+        /// <summary>
+        /// Clears all data in the collection
+        /// </summary>
+        [WriteAccessRequired]
+        public void Clear()
+        {
+            if (!IsCreated)
+            {
+                return;
+            }
+
+            Allocator allocator = m_BufferInfo->Allocator;
+            
             for (int i = 0; i < m_BufferInfo->LaneCount; ++i)
             {
                 LaneInfo* lane = m_Lanes + i;
-
+                
                 BlockInfo* block = lane->FirstBlock;
                 while (block != null)
                 {
@@ -193,20 +216,20 @@ namespace Anvil.Unity.DOTS.Data
                     UnsafeUtility.Free(currentBlock->Data, allocator);
                     UnsafeUtility.Free(currentBlock, allocator);
                 }
+                
+                lane->FirstBlock = null;
+                lane->CurrentWriterBlock = null;
+                lane->WriterHead = null;
+                lane->WriterEndOfBlock = null;
+                lane->Count = 0;
             }
-
-            UnsafeUtility.Free(m_Lanes, allocator);
-            UnsafeUtility.Free(m_BufferInfo, allocator);
-
-            m_Lanes = null;
-            m_BufferInfo = null;
         }
 
         /// <summary>
         /// Schedules the disposal of the collection.
         /// </summary>
         /// <param name="inputDeps">The <see cref="JobHandle" /> to wait on before disposing</param>
-        /// <returns>A <see cref="JobHandle" /> for when disposal is complete</returns>
+        /// <returns>A <see cref="JobHandle"/> for when disposal is complete</returns>
         public JobHandle Dispose(JobHandle inputDeps)
         {
             DisposeJob disposeJob = new DisposeJob(this);
@@ -214,6 +237,17 @@ namespace Anvil.Unity.DOTS.Data
             m_Lanes = null;
 
             return disposeJobHandle;
+        }
+        
+        /// <summary>
+        /// Schedules the clearing of the collection.
+        /// </summary>
+        /// <param name="inputDeps">The <see cref="JobHandle"/> to wait on before clearing.</param>
+        /// <returns>A <see cref="JobHandle"/> for when clearing is complete</returns>
+        public JobHandle Clear(JobHandle inputDeps)
+        {
+            ClearJob clearJob = new ClearJob(this);
+            return clearJob.Schedule(inputDeps);
         }
 
         /// <summary>
@@ -615,6 +649,22 @@ namespace Anvil.Unity.DOTS.Data
             public void Execute()
             {
                 m_UnsafeTypedStream.Dispose();
+            }
+        }
+
+        [BurstCompile]
+        private struct ClearJob : IJob
+        {
+            private UnsafeTypedStream<T> m_UnsafeTypedStream;
+
+            public ClearJob(UnsafeTypedStream<T> unsafeTypedStream)
+            {
+                m_UnsafeTypedStream = unsafeTypedStream;
+            }
+
+            public void Execute()
+            {
+                m_UnsafeTypedStream.Clear();
             }
         }
     }
--- a/Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
+++ b/Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
@@ -37,6 +37,7 @@ namespace Anvil.Unity.DOTS.Data
             public int BlockSize;
             public int LaneCount;
             public Allocator Allocator;
+            public Allocator BlockAllocator;
         }
 
         /// <summary>
@@ -112,15 +113,36 @@ namespace Anvil.Unity.DOTS.Data
         {
             get => m_BufferInfo->BlockSize / ELEMENT_SIZE;
         }
+        
+        /// <summary>
+        /// Convenience constructor that fits as many elements into a 16kb block size and allows you to specify the
+        /// number of lanes.
+        /// </summary>
+        /// <param name="allocator">The <see cref="Allocator"/> to use when allocating memory for the collection and the content</param>
+        /// <param name="laneCount">The number of lanes to allow reading from/writing to.</param>
+        public UnsafeTypedStream(Allocator allocator, int laneCount) : this(math.max(CHUNK_SIZE / UnsafeUtility.SizeOf<T>(), 1), allocator, allocator, laneCount)
+        {
+        }
 
 
         /// <summary>
         /// Convenience constructor that fits as many elements into a 16kb block size and allows you to specify the
         /// number of lanes.
         /// </summary>
-        /// <param name="allocator">The <see cref="Allocator" /> to use when allocating memory.</param>
+        /// <param name="allocator">The <see cref="Allocator"/> to use when allocating memory for the collection but not the content</param>
+        /// <param name="blockAllocator">The <see cref="Allocator"/> to use when allocating memory for the content in the collection.</param>
         /// <param name="laneCount">The number of lanes to allow reading from/writing to.</param>
-        public UnsafeTypedStream(Allocator allocator, int laneCount) : this(math.max(CHUNK_SIZE / UnsafeUtility.SizeOf<T>(), 1), allocator, laneCount)
+        public UnsafeTypedStream(Allocator allocator, Allocator blockAllocator, int laneCount) : this(math.max(CHUNK_SIZE / UnsafeUtility.SizeOf<T>(), 1), allocator, blockAllocator, laneCount)
+        {
+        }
+        
+        /// <summary>
+        /// Convenience constructor that fits as many elements into a 16kb block size.
+        /// Sets the number of lanes to be the maximum amount of worker threads available plus the main thread.
+        /// <see cref="JobsUtility.JobWorkerMaximumCount"/> + 1
+        /// </summary>
+        /// <param name="allocator">The <see cref="Allocator"/> to use when allocating memory for the collection and the content</param>
+        public UnsafeTypedStream(Allocator allocator) : this(math.max(CHUNK_SIZE / UnsafeUtility.SizeOf<T>(), 1), allocator, allocator, ParallelAccessUtil.CollectionSizeForMaxThreads)
         {
         }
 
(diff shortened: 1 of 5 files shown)


Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
```diff
@@ -112,15 +114,36 @@ public int ElementsPerBlock
         {
             get => m_BufferInfo->BlockSize / ELEMENT_SIZE;
         }
+        
+        /// <summary>
+        /// Convenience constructor that fits as many elements into a 16kb block size and allows you to specify the
+        /// number of lanes.
```
mbaker3, line 120:
```suggestion
        /// Convenience constructor that fits as many elements into a 16kb block size as possible and allows you
        /// to specify the number of lanes.
```

mbaker3: This typo appears in all the constructor docs.

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
```diff
@@ -400,10 +473,18 @@ public LaneWriter AsLaneWriter(int laneIndex)
             [NativeDisableUnsafePtrRestriction] private readonly BufferInfo* m_BufferInfo;
             [NativeDisableUnsafePtrRestriction] private readonly LaneInfo* m_Lane;
 
-            internal LaneWriter(LaneInfo* lane, BufferInfo* bufferInfo)
+            /// <summary>
+            /// Reports if this <see cref="LaneWriter"/> has been created and is valid to use.
+            /// </summary>
+            public bool IsCreated
+            {
+                get => m_BufferInfo != null;
```
mbaker3, line 481: Some comment as on `Writer` about isCreated. Although in this case you'll get the error when the constructor tries to populate `m_Lane`

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
```diff
@@ -462,8 +543,15 @@ private void CheckForNewBlock()
...
             [NativeDisableUnsafePtrRestriction] private readonly BufferInfo* m_BufferInfo;
-            [NativeDisableUnsafePtrRestriction] private readonly LaneInfo* m_Lanes;
-
+            
+            /// <summary>
+            /// Reports if this <see cref="Reader"/> has been created and is valid to use.
+            /// </summary>
+            public bool IsCreated
+            {
+                get => m_BufferInfo != null;
```
mbaker3, line 552: same comment about adding an assert to the constructor.

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
```diff
@@ -526,6 +611,14 @@ public struct LaneReader
             private BlockInfo* m_CurrentReadBlock;
             private byte* m_EndOfCurrentBlock;
             private byte* m_ReaderHead;
+            
+            /// <summary>
+            /// Reports if this <see cref="LaneReader"/> has been created and is valid to use.
+            /// </summary>
+            public bool IsCreated
+            {
+                get => m_BufferInfo != null;
```
mbaker3, line 620: 💿 same comment about adding an assert to the constructor.

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
```diff
@@ -617,5 +710,21 @@ public void Execute()
                 m_UnsafeTypedStream.Dispose();
             }
         }
+
+        [BurstCompile]
+        private struct ClearJob : IJob
+        {
+            private UnsafeTypedStream<T> m_UnsafeTypedStream;
```
mbaker3, line 717: I don't think it does anything today but `[WriteOnly]`?

Scripts/Runtime/Entities/Util/ChunkUtil.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// <summary>
+    /// Helper class for dealing with optimal numbers of elements in a chunk
+    /// </summary>
+    public static class ChunkUtil
+    {
+        //See Chunk.kChunkSize - Can't access so we redefine here
+        private const int CHUNK_SIZE = 16 * 1024;
+        
+        /// <summary>
+        /// Gets the maximum number of elements that would fit into a chunk of memory.
```
mbaker3, line 15:
```suggestion
        /// Gets the maximum number of elements that will fit into a chunk of memory.
```

Scripts/Runtime/Entities/Util/ChunkUtil.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// At minimum it will return 1 if the element is larger than a chunk.
```
mbaker3, line 16: I wonder if we should emit a warning if this is the case.
I guess you still allocate the right size if `type > CHUNK_SIZE` so it's not going to cause problems.

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
```diff
@@ -34,9 +34,14 @@ public unsafe struct UnsafeTypedStream<T> : INativeDisposable
         [BurstCompatible]
         internal struct BufferInfo
         {
+            public static readonly int SIZE = UnsafeUtility.SizeOf<BufferInfo>();
+            public static readonly int ALIGNMENT = UnsafeUtility.AlignOf<BufferInfo>();
```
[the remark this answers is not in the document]
jkeon: `BufferInfo` was an outside type at first but we wanted to bring it inside because then we'd have to qualify the name for any other collections that also have `BufferInfo`s. It would become `UnsafeTypedStreamBufferInfo` etc so as not to conflict. 

The addition of two ints to the global memory of the app per variant I think is ok for the niceties in developing with it being an inner type. 

Valid point though.

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
jkeon: Fixed.

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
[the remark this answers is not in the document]
mbaker3: Makes sense. I'll pay the int costs for the better names.

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
```diff
@@ -129,8 +152,20 @@ public UnsafeTypedStream(Allocator allocator, int laneCount) : this(math.max(CHU
...
+        }
+
+        /// <summary>
+        /// More explicit constructor that allows for specifying how many elements to put into each block. Useful for
+        /// smaller counts so that large blocks aren't allocated if not needed.
+        /// </summary>
+        /// <param name="elementsPerBlock">The number of elements to allocate space for in each block.</param>
+        /// <param name="allocator">The <see cref="Allocator"/> to use when allocating memory for the collection and the content</param>
+        /// <param name="laneCount">The number of lanes to allow reading from/writing to.</param>
+        public UnsafeTypedStream(int elementsPerBlock, Allocator allocator, int laneCount) : this (elementsPerBlock, allocator,allocator, laneCount)
```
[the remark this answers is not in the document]
jkeon: I played around with it and I can default the allocators but not the elementCount or laneCount because those aren't compile time constants.

But.... because this a struct, I can't have a parameterless constructor so I need the dev to pass in the allocators explicitly like they do on all the Unity built in `NativeCollectionX` so that I can call `this()` with the calculations for elementCount and laneCount.

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
```diff
@@ -34,9 +34,14 @@ public unsafe struct UnsafeTypedStream<T> : INativeDisposable
...
         internal struct BufferInfo
         {
+            public static readonly int SIZE = UnsafeUtility.SizeOf<BufferInfo>();
+            public static readonly int ALIGNMENT = UnsafeUtility.AlignOf<BufferInfo>();
+            
             public int BlockSize;
             public int LaneCount;
             public Allocator Allocator;
+            public Allocator BlockAllocator;
+            [NativeDisableUnsafePtrRestriction] public LaneInfo* LanesInfo;
```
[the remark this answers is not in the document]
jkeon: Changed to `LaneInfos`

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
```diff
@@ -365,10 +433,17 @@ public int LaneCount
             {
                 get => m_BufferInfo->LaneCount;
             }
+            
+            /// <summary>
+            /// Reports if this <see cref="Writer"/> has been created and is valid to use.
+            /// </summary>
+            public bool IsCreated
+            {
+                get => m_BufferInfo != null;
```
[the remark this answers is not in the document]
jkeon: The only way to create a `Writer`, `LaneWriter`, `Reader` or `LaneReader`  is through the methods `AsWriter` etc. 

Most of those had an assert there for `IsCreated` but I've added them where they were missing.

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
jkeon: I made sure these constructors can't be called without going through an `IsCreated` assert first now.

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
```diff
@@ -139,27 +174,29 @@ public UnsafeTypedStream(Allocator allocator) : this(math.max(CHUNK_SIZE / Unsaf
         /// smaller counts so that large blocks aren't allocated if not needed.
         /// </summary>
         /// <param name="elementsPerBlock">The number of elements to allocate space for in each block.</param>
-        /// <param name="allocator">The <see cref="Allocator" /> to use when allocating memory.</param>
+        /// <param name="allocator">The <see cref="Allocator"/> to use when allocating memory for the collection but not the content</param>
+        /// <param name="blockAllocator">The <see cref="Allocator"/> to use when allocating memory for the content in the collection.</param>
         /// <param name="laneCount">The number of lanes to allow reading from/writing to.</param>
-        public UnsafeTypedStream(int elementsPerBlock, Allocator allocator, int laneCount)
+        public UnsafeTypedStream(int elementsPerBlock, Allocator allocator, Allocator blockAllocator, int laneCount)
```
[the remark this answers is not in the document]
jkeon: Agreed. Added a nice assert that ensures good combinations.

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
jkeon: Ensured that constructor can't be called without passing through the assert.

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
jkeon: Ensured that constructor can't be called without passing through the assert.

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
jkeon: Added

Scripts/Runtime/Entities/Util/ChunkUtil.cs
jkeon: Fixed.

Scripts/Runtime/Entities/Util/ChunkUtil.cs
jkeon: Yeah, it won't break and I feel like you're probably going to know if you have a struct that is that massive?

mbaker3: That's fair. Would be annoying if you, for some reason, had to work with a massive struct.

Scripts/Runtime/Jobs/ConsolidateToNativeArrayJob.cs
```diff
@@ -0,0 +1,53 @@
...
+            int arrayIndex = 0;
+            for (int laneIndex = 0; laneIndex < m_Reader.LaneCount; ++laneIndex)
+            {
+                UnsafeTypedStream<T>.LaneReader laneReader = m_Reader.AsLaneReader(laneIndex);
+                int elementCount = laneReader.Count;
+                for (int elementIndex = 0; elementIndex < elementCount; ++elementIndex)
+                {
+                    array[arrayIndex] = laneReader.Read();
+                    arrayIndex++;
+                }
```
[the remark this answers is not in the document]
jkeon: Yes, this is a good idea. Done.

jkeon:
@mbaker3 Good for review again.

See the `ReadInto` method for the block memcpy

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
```diff
@@ -674,6 +691,21 @@ private void CheckForEndOfBlock()
                 return ref UnsafeUtility.AsRef<T>(readPointer);
             }
 
+            internal void ReadBlock(byte* dstStart, int dstStartIndex, int elementsToRead)
+            {
+#if ENABLE_UNITY_COLLECTIONS_CHECKS
+                Assert.IsTrue(Count > 0 && m_CurrentReadBlock != null && m_ReaderHead != null);
```
mbaker3, line 697: Should we add a condition to make sure m_ReaderHead + elementsToRead stays within the bounds of the lane?

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
```diff
@@ -589,12 +586,32 @@ public int Count()
...
                 {
-                    LaneInfo* lane = m_BufferInfo->LanesInfo + i;
+                    LaneInfo* lane = m_BufferInfo->LaneInfos + i;
                     count += lane->Count;
                 }
 
                 return count;
             }
+
+            public void ReadInto(ref NativeArray<T> array)
```
mbaker3, line 596: - We should probably rename as `CopyTo` for consistency. `Read`, to me, implies stateful reading where the read head will be advanced to the end of the content delivered.
 - Might make sense to move this method up to `UnsafeTypedStream`

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
jkeon: Yeah that's a good idea. Done.

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
jkeon: > Might make sense to move this method up to UnsafeTypedStream

I don't want to move it up because the most common use case is have the `Reader` only in your job and not the full structure. You bring up a  good point though so I exposed the ability to do it from the full structure which drills down through a reader instance.

> We should probably rename as CopyTo for consistency. Read, to me, implies stateful reading where the read head will be advanced to the end of the content delivered.

I agree but this is stateful reading. The reader head will be moved to the end for that instance of the reader. I need to do this so that we can detect when we're at the end of a block and change to the next one.

Thoughts?

mbaker3: Maybe I'm missing something but this method doesn't change the state of the reader does it?
Your LaneReaders move through a state but from what I understand they're aligned to the first element of the first block with each instance you get back from an `AsLaneReader()` call.

Calling `ReadInto()` doesn't change the state of your `Reader` instance. Meaning if you call `ReadInto()` on this instance and then try to perform other other read operations with this instance it behaves the same way. The ReadInto doesn't consider an existing reader state or affect the reader state. It just dumps all of the streams data into an array from the first element of the first lane to the last element of the last lane.

mbaker3 (approved):

PR #35, opened 2022-06-03
Updated Job Scheduling to Support Init on Thread

jkeon:

Jobs with `DeferredNativeArrays` often needed to use the `NativeThreadIndex` and then write to a certain "lane" on a wide collection like `UnsafeTypedStream`. 

Today we have to remember to do some initialization work before we iterate through the batch making `IJobDeferredNativeArrayForBatch` the only real choice. 

The downside is that depending on your batch size, if the same job is executed multiple times on the same thread, the same initialization logic will happen multiple times. In the below example `InitializeSelf` can be called multiple times per thread.

```
            public void Execute(int startIndex, int count)
            {
                //Initialize first - Uses a native thread index via [NativeSetThreadIndex] attribute.
                m_Random.UpdateSeedWithNativeThreadIndex(m_NativeThreadIndex);
                m_ContinueLaneWriter = m_ContinueWriter.AsLaneWriter(ParallelAccessUtil.CollectionIndexForThread(m_NativeThreadIndex));
                
                //Now iterate and do the work
                int endIndex = startIndex + count;
                for (int i = 0; i < endIndex; ++i)
                {
                    int value = m_Array[i];
                    Debug.Assert(value == SNASpawnJob.INT_TO_WRITE);
                }
            }
```

We also always need to write the loop and we also need to ensure our job has the native thread index on it via:

```
[NativeSetThreadIndex] private readonly int m_NativeThreadIndex;
```

The job scheduling now handles getting the native thread index for you and will track whether a thread has received an init call. Because the init call is now a separate function, the `IJobDeferredNativeArrayFor` is now a viable job type to use and in most cases preferable. 

The `InitForThread` function is called once per thread regardless of how many batches run on the thread.
The native thread index is passed in so you don't need to deal with the attribute and injection.

And our execute can smaller because we don't need to write the loop anymore.

```
            public void InitForThread(int nativeThreadIndex)
            {
                m_Random.UpdateSeedWithNativeThreadIndex(nativeThreadIndex);
                m_ContinueLaneWriter = m_ContinueWriter.AsLaneWriter(ParallelAccessUtil.CollectionIndexForThread(nativeThreadIndex));
            }

            public void Execute(int index)
            {
                int value = m_Array[index];
                Debug.Assert(value == SNASpawnJob.INT_TO_WRITE);
            }
```

- None

 - None

 - [x] Yes - Any jobs that used these interfaces will need to adhere to the new `InitForThread` function. Should be minimal work to convert.
 - [ ] No


Scripts/Runtime/Jobs/Deferred/JobDeferredNativeArrayForBatchExtension.cs
```diff
@@ -111,10 +116,20 @@ internal static void Initialize()
 #if ENABLE_UNITY_COLLECTIONS_CHECKS
                     JobsUtility.PatchBufferMinMaxRanges(bufferRangePatchData, UnsafeUtility.AddressOf(ref jobData), beginIndex, endIndex - beginIndex);
 #endif
-
                     jobData.Execute(beginIndex, endIndex - beginIndex);
                 }
             }
+
+            private const int DEFAULT_NATIVE_THREAD_INDEX = -1;
```
mbaker3, line 123: 🚲 Consider renaming to UNSET_NATIVE_THREAD_INDEX. Otherwise, it reads a bit like this is a valid thread index.

Scripts/Runtime/Jobs/Deferred/JobDeferredNativeArrayForBatchExtension.cs
```diff
@@ -111,10 +116,20 @@ internal static void Initialize()
...
+
+            private TJob m_JobData;
+            [NativeSetThreadIndex] private readonly int m_NativeThreadIndex;
+            
+            public JobDeferredNativeArrayForBatchProducer(ref TJob jobData)
+            {
+                m_JobData = jobData;
+                m_NativeThreadIndex = DEFAULT_NATIVE_THREAD_INDEX;
+            }
```
mbaker3, line 132: What was this struct used for before? Does it make sense that we're tacking this functionality onto the producer?

It looks like `JobDeferredNativeArrayForBatchProducer` was just a struct with a bunch of static methods?

mbaker3 (commented):
Looking good. Just a couple of questions

Scripts/Runtime/Jobs/Deferred/JobDeferredNativeArrayForBatchExtension.cs
jkeon: Done

jkeon:
@mbaker3 Reworked to make things nicer and adjusted to your feedback. 

I also added similar job types for:

`IJob` -> `IAnvilJob`
`IJobFor` -> `IAnvilJobFor`
`IJobParallelForBatch` -> `IAnvilJobBatch`
`IJobParallelForDefer` -> `IAnvilJobForDefer` and `IAnvilJobBatchDefer`

I would have liked to have made one for

`IJobEntityBatch` 

But it uses a ton of internal chuck aspects that we don't have access to :(

Scripts/Runtime/Jobs/JobTypes/AnvilJobBatchExtension.cs
```diff
@@ -0,0 +1,136 @@
...
+
+        /// <summary>
+        /// This method is only to be called by automatically generated setup code.
+        /// Called by Unity automatically
+        /// </summary>
+        /// <typeparam name="TJob"></typeparam>
+        public static void EarlyJobInit<TJob>()
+            where TJob : struct, IAnvilJobBatch
+        {
+            WrapperJobProducer<TJob>.Initialize();
```
mbaker3, line 48: Worth doing an assert on the stack for this to make sure there is only ever the one place that this is called?

mbaker3 (approved):
Minor assert suggestion.

Otherwise this looks good assuming we will be using these job types as the default rather than Unity's built in. If we won't be then we should probably use something more descriptive than `IAnvil*` for the jobs.

jkeon:
> Otherwise this looks good assuming we will be using these job types as the default rather than Unity's built in. If we won't be then we should probably use something more descriptive than IAnvil* for the jobs.

Correct, we'll use these as the default job types rather than Unity's built in.

PR #36, opened 2022-06-03
Adding the ability to clear a DeferredNativeArray

jkeon:

Currently a `DeferredNativeArray` must be `Disposed` and recreated if you want to use it again similar to the other built in Unity native collections.

There's often a pattern of having a persistent collection existing in the system and the temp job allocating the data in a job per frame similar to the `UnsafeTypedStream`.

This allows for keeping a collection around and just clearing the data each frame instead of having to dispose and recreate.

`DeferredNativeArray` now supports this pattern.

- None

 - None

 - [ ] Yes 
 - [x] No

### Tag-Alongs

- Fixed a bug where the debugger info would throw an error when inspecting in the debugger.
- Fixed a bug where a scheduled `Dispose` would clear the pointer of an inflight iteration resulting in a runtime error in the job. Now properly clearing the pointer where we should be.

### DIFF
--- a/Scripts/Runtime/Data/Collections/DeferredNativeArray.cs
+++ b/Scripts/Runtime/Data/Collections/DeferredNativeArray.cs
@@ -1,14 +1,40 @@
 using System;
-using System.Diagnostics;
 using System.Runtime.InteropServices;
 using Unity.Assertions;
 using Unity.Burst;
 using Unity.Collections;
 using Unity.Collections.LowLevel.Unsafe;
 using Unity.Jobs;
+using UnityEngine;
 
 namespace Anvil.Unity.DOTS.Data
 {
+    /// <summary>
+    /// Scheduling information for a <see cref="DeferredNativeArray{T}"/>
+    /// </summary>
+    /// <remarks>
+    /// Provides access to internal data necessary for job scheduling but avoids
+    /// boxing if done with interfaces.
+    /// </remarks>
+    public unsafe struct DeferredNativeArrayScheduleInfo
+    {
+        internal void* SafetyHandlePtr
+        {
+            get;
+        }
+
+        internal void* BufferPtr
+        {
+            get;
+        }
+
+        internal DeferredNativeArrayScheduleInfo(void* safetyHandlePtr, void* bufferPtr)
+        {
+            SafetyHandlePtr = safetyHandlePtr;
+            BufferPtr = bufferPtr;
+        }
+    }
+
     /// <summary>
     /// A native collection similar to <see cref="NativeArray{T}"/> but intended for use in a deferred context.
     /// Useful for cases where a job that hasn't finished yet will determine the length of the array.
@@ -41,7 +67,7 @@ namespace Anvil.Unity.DOTS.Data
 
         [BurstCompatible]
         [StructLayout(LayoutKind.Sequential)]
-        internal unsafe struct BufferInfo
+        private unsafe struct BufferInfo
         {
             // ReSharper disable once MemberHidesStaticFromOuterClass
             public static readonly int SIZE = UnsafeUtility.SizeOf<BufferInfo>();
@@ -52,6 +78,7 @@ namespace Anvil.Unity.DOTS.Data
             [NativeDisableUnsafePtrRestriction] public void* Buffer;
             public int Length;
             public Allocator DeferredAllocator;
+            public bool IsPending;
         }
 
         //*************************************************************************************************************
@@ -100,6 +127,7 @@ namespace Anvil.Unity.DOTS.Data
             array.m_BufferInfo->Length = 0;
             array.m_BufferInfo->Buffer = null;
             array.m_BufferInfo->DeferredAllocator = deferredAllocator;
+            array.m_BufferInfo->IsPending = true;
 
             array.m_Allocator = allocator;
 
@@ -125,6 +153,7 @@ namespace Anvil.Unity.DOTS.Data
             {
                 return;
             }
+
             ClearBufferInfo(bufferInfo);
             UnsafeUtility.Free(bufferInfo, allocator);
         }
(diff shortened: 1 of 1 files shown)


Scripts/Runtime/Data/Collections/DeferredNativeArray.cs
```diff
@@ -85,12 +86,11 @@ private static void AssertValidElementType()
...
         }
 
-        private static unsafe void Allocate(Allocator allocator, out DeferredNativeArray<T> array)
+        private static unsafe void Allocate(Allocator allocator, Allocator deferredAllocator, out DeferredNativeArray<T> array)
         {
-            if (allocator <= Allocator.None)
-            {
-                throw new ArgumentException("Allocator must be Temp, TempJob or Persistent", nameof(allocator));
-            }
+            //Ensures that deferred allocator can only be Temp, TempJob or Persistent and that the allocator is at the same or higher level.
```
mbaker3, line 91:
```suggestion
            //Ensures that deferred allocator can only be Temp, TempJob or Persistent and that the allocator is at the same or more persistent level.
```

mbaker3 (commented):
Lookin good. Comments to discuss

Scripts/Runtime/Data/Collections/DeferredNativeArray.cs
jkeon: Done

Scripts/Runtime/Data/Collections/DeferredNativeArray.cs
```diff
@@ -285,11 +331,36 @@ public void Execute()
                 //when this job was scheduled because it requires main thread access
                 if (m_BufferInfo->Buffer != null)
                 {
-                    UnsafeUtility.Free(m_BufferInfo->Buffer, m_Allocator);
+                    UnsafeUtility.Free(m_BufferInfo->Buffer, m_BufferInfo->DeferredAllocator);
+                    m_BufferInfo->Buffer = null;
```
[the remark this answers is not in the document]
jkeon: No longed as relevant since they both call the static functions so the repetition is reduced.

jkeon:
@mbaker3 Ready for re-review.

I also found a bug with the asserts when clearing so those have been removed since it is now possible to get the job array then deferred create and then clear and get the job array again for a new array and then deferred create again in a subsequent job.

Scripts/Runtime/Data/Collections/DeferredNativeArray.cs
```diff
@@ -243,8 +254,6 @@ public unsafe JobHandle Clear(JobHandle inputDeps)
         /// <param name="nativeArrayOptions">The <see cref="NativeArrayOptions"/> for initializing the array memory.</param>
         public unsafe NativeArray<T> DeferredCreate(int newLength, NativeArrayOptions nativeArrayOptions = NativeArrayOptions.ClearMemory)
         {
-            AssertForDeferredCreate();
```
mbaker3, line 200: Potential issues now that this assert is gone.
1. Calling DeferredCreate multiple times in a row will leak memory
2. There is no tracking whether a clear has been scheduled between the deferred create calls so you could be leaking memory even though you think you're not calling deferred created multiple times in a row
3. `ClearJob` holds a pointer to `m_BufferInfo` instead of making a copy of it so right now it will clear the most recently allocated memory not the currently allocated memory at the time of schedule. This is another opportunity to leak memory.

mbaker3 (changes requested):
Identified some memory leak opportunities now that the assert has been removed. Will need to address before we merge in.

Scripts/Runtime/Data/Collections/DeferredNativeArray.cs
jkeon: 1. Added an assert and flag to ensure you can't call this multiple times in a row now.
2. Via the flag the tracking is here so you can't leak because you can't allocate again until you've cleared.
3. This would be awesome but we can't actually do it because the copy of the buffer info wouldn't have the inner data exist until later. With the flag and assert we're protected against leaks.

Scripts/Runtime/Data/Collections/DeferredNativeArray.cs
```diff
@@ -52,6 +78,7 @@ internal unsafe struct BufferInfo
             [NativeDisableUnsafePtrRestriction] public void* Buffer;
             public int Length;
             public Allocator DeferredAllocator;
+            public bool IsPending;
```
mbaker3, line 81: Naming is a bit misleading. Maybe `CanBufferAllocate`, `IsClearPending`, etc...?

`IsPending` just makes it seem like an action is imminent which isn't really the case unless you actually call createDeferred again.

I wonder too if this should be compiler defined out since it's really just a safety check. Although, if serialized, that could lead to incompatible saves between debug and release mode.

Scripts/Runtime/Data/Collections/DeferredNativeArray.cs
```diff
@@ -1,14 +1,40 @@
...
 namespace Anvil.Unity.DOTS.Data
 {
+    /// <summary>
+    /// Scheduling information for a <see cref="DeferredNativeArray{T}"/>
+    /// </summary>
+    /// <remarks>
+    /// Provides access to internal data necessary for job scheduling but avoids
+    /// boxing if done with interfaces.
+    /// </remarks>
+    public unsafe struct DeferredNativeArrayScheduleInfo
```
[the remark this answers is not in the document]
jkeon: No, we need it to pass into the job scheduling. The inner values are all internal so you don't mess with the pointers though.

Scripts/Runtime/Data/Collections/DeferredNativeArray.cs
```diff
@@ -156,6 +188,17 @@ public unsafe int Length
                     : 0;
         }
 
+        /// <summary>
+        /// Scheduling information about this <see cref="DeferredNativeArray{T}"/> for use in job
+        /// scheduling.
+        /// </summary>
+        public unsafe DeferredNativeArrayScheduleInfo ScheduleInfo
```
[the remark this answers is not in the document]
jkeon: ^ See Above

mbaker3 (approved):
To the moon and back.
🚢 🚢 🚢 🚢

PR #37, opened 2022-06-21
Adding Virtual Data

jkeon:
Virtual Data is an abstract way to represent data that doesn't have a persistent lifetime relative to an Entity. It is a good way to represent state and flowing through tasks.

**NOTE** On it's own, this PR doesn't add much for the developer because a lot of it is locked down via `internal`. This is intentional as `VirtualData` will be created and controlled via the upcoming PRs for `TaskDrivers`. However in the interest of smaller PR's this is a good chunk to look at.

This doesn't exist today, you would have to manage collections manually and manage access to those collections manually as well.

Nice wrapped behaviour to take care of
- Shared Writing to add new data
- Consolidated fast NativeArray for iteration reading
- Writing results to different result locations for implicit grouping of data for a context while still having the overall type of data be generically updated. 
    - Ex. All `Timers` can be updated together but completed to different locations depending who spawned them.

- None

 - #36 and #35

 - [ ] Yes
 - [x] No

### DIFF
--- a/Scripts/Runtime/Data/VirtualData/IVirtualData.cs
+++ b/Scripts/Runtime/Data/VirtualData/IVirtualData.cs
@@ -1,0 +1,26 @@
+using Anvil.CSharp.Core;
+using Anvil.Unity.DOTS.Jobs;
+using Unity.Jobs;
+
+namespace Anvil.Unity.DOTS.Data
+{
+    /// <summary>
+    /// Interface for dealing with <see cref="VirtualData{TKey,TInstance}"/> in a generic way without having
+    /// to know the types.
+    /// </summary>
+    public interface IVirtualData : IAnvilDisposable
+    {
+        internal void AddResultDestination(IVirtualData resultData);
+        internal void RemoveResultDestination(IVirtualData resultData);
+        
+        internal JobHandle ConsolidateForFrame(JobHandle dependsOn);
+
+        internal JobHandle AcquireForUpdate();
+        internal void ReleaseForUpdate(JobHandle releaseAccessDependency);
+
+        internal AccessController AccessController
+        {
+            get;
+        }
+    }
+}
--- a/Scripts/Runtime/Data/VirtualData/InstanceData/ILookupData.cs
+++ b/Scripts/Runtime/Data/VirtualData/InstanceData/ILookupData.cs
@@ -1,0 +1,21 @@
+using System;
+
+namespace Anvil.Unity.DOTS.Data
+{
+    /// <summary>
+    /// Data to be used in <see cref="VirtualData{TKey,TInstance}"/>
+    /// that can be looked up with a <typeparamref name="TKey"/>
+    /// </summary>
+    /// <typeparam name="TKey">The type of key to use for lookup.</typeparam>
+    public interface ILookupData<out TKey>
+        where TKey : struct, IEquatable<TKey>
+    {
+        /// <summary>
+        /// The lookup key for this struct
+        /// </summary>
+        public TKey Key
+        {
+            get;
+        }
+    }
+}
--- a/Scripts/Runtime/Data/VirtualData/Jobs/VDJobReader.cs
+++ b/Scripts/Runtime/Data/VirtualData/Jobs/VDJobReader.cs
@@ -1,0 +1,30 @@
+using Unity.Collections;
+
+namespace Anvil.Unity.DOTS.Data
+{
+    /// <summary>
+    /// A struct to be used in jobs that only allows for reading.
+    /// Commonly used to read and trigger something else.
+    /// </summary>
+    /// <typeparam name="TInstance">They type of data to read</typeparam>
+    [BurstCompatible]
+    public readonly struct VDJobReader<TInstance>
+        where TInstance : struct
+    {
+        [ReadOnly] private readonly NativeArray<TInstance> m_Iteration;
+
+        internal VDJobReader(NativeArray<TInstance> iteration)
+        {
+            m_Iteration = iteration;
+        }
+        
+        /// <summary>
+        /// Gets the <typeparamref name="TInstance"/> at the specified index.
+        /// </summary>
+        /// <param name="index">The index into the backing array</param>
+        public TInstance this[int index]
+        {
+            get => m_Iteration[index];
+        }
+    }
+}
--- a/Scripts/Runtime/Data/VirtualData/Jobs/VDJobResultsDestination.cs
+++ b/Scripts/Runtime/Data/VirtualData/Jobs/VDJobResultsDestination.cs
@@ -1,0 +1,26 @@
+using Unity.Collections;
+
+namespace Anvil.Unity.DOTS.Data
+{
+    /// <summary>
+    /// A struct to be used in jobs that acts as a placeholder for results to be written to
+    /// later on. Use with <see cref="IVirtualDataInstance{TResult}"/>
+    /// </summary>
+    /// <typeparam name="TResult">The type of result that can be written</typeparam>
+    [BurstCompatible]
+    public readonly struct VDJobResultsDestination<TResult>
+        where TResult : struct
+    {
+        //Implicit cast to a writer when we need to actually write.
+        //Otherwise this struct doesn't let you do anything with it until we can guarantee you
+        //have access.
+        public static implicit operator VDJobResultsWriter<TResult>(VDJobResultsDestination<TResult> destination) => new VDJobResultsWriter<TResult>(destination.m_ResultWriter);
+
+        [ReadOnly] private readonly UnsafeTypedStream<TResult>.Writer m_ResultWriter;
+
+        internal VDJobResultsDestination(UnsafeTypedStream<TResult>.Writer resultWriter)
+        {
+            m_ResultWriter = resultWriter;
+        }
+    }
+}
--- a/Scripts/Runtime/Data/VirtualData/Jobs/VDJobResultsWriter.cs
+++ b/Scripts/Runtime/Data/VirtualData/Jobs/VDJobResultsWriter.cs
@@ -1,0 +1,38 @@
+using Unity.Collections;
+
+namespace Anvil.Unity.DOTS.Data
+{
+    /// <summary>
+    /// A struct to be used in jobs that writes new results.
+    /// <seealso cref="IVirtualDataInstance{TResult}"/>
+    /// <seealso cref="VDJobResultsDestination{TResult}"/>
+    /// </summary>
+    /// <typeparam name="TResult">The type of result to write</typeparam>
+    [BurstCompatible]
+    public readonly struct VDJobResultsWriter<TResult>
+        where TResult : struct
+    {
+        [ReadOnly] private readonly UnsafeTypedStream<TResult>.Writer m_ResultWriter;
+
+        internal VDJobResultsWriter(UnsafeTypedStream<TResult>.Writer resultWriter)
+        {
+            m_ResultWriter = resultWriter;
+        }
+        
+        /// <summary>
+        /// Adds a new result to the <see cref="VirtualData{TKey,TInstance}"/> of results.
+        /// </summary>
+        /// <param name="result">The result to write</param>
+        /// <param name="laneIndex">The lane index to write to.</param>
+        public void Add(TResult result, int laneIndex)
+        {
+            Add(ref result, laneIndex);
+        }
+        
+        /// <inheritdoc cref="Add(TResult,int)"/>
+        public void Add(ref TResult result, int laneIndex)
+        {
+            m_ResultWriter.AsLaneWriter(laneIndex).Write(ref result);
+        }
+    }
+}
--- a/Scripts/Runtime/Data/VirtualData/Jobs/VDJobUpdater.cs
+++ b/Scripts/Runtime/Data/VirtualData/Jobs/VDJobUpdater.cs
@@ -1,0 +1,121 @@
+using Anvil.Unity.DOTS.Jobs;
+using System;
+using Unity.Collections;
+using UnityEngine;
+
+namespace Anvil.Unity.DOTS.Data
+{
+    /// <summary>
+    /// A struct to be used in jobs that is for updating the <see cref="VirtualData{TKey,TInstance}"/>
+    /// that this <see cref="VDJobUpdater{TKey,TInstance}"/> represents.
+    ///
+    /// Commonly used to iterate through all instances, perform some work and either
+    /// <see cref="Continue(TInstance)"/> if more work needs to be done next frame or
+    /// <see cref="Complete"/> if the work is done.
+    /// </summary>
+    /// <typeparam name="TKey">The type of key to use for lookup of the instance</typeparam>
+    /// <typeparam name="TInstance">The type of instance</typeparam>
+    [BurstCompatible]
+    public struct VDJobUpdater<TKey, TInstance>
+        where TKey : struct, IEquatable<TKey>
+        where TInstance : struct, ILookupData<TKey>
+    {
+        private const int DEFAULT_LANE_INDEX = -1;
+
+        [ReadOnly] private readonly UnsafeTypedStream<TInstance>.Writer m_ContinueWriter;
+        [ReadOnly] private readonly NativeArray<TInstance> m_Iteration;
+
+        private UnsafeTypedStream<TInstance>.LaneWriter m_ContinueLaneWriter;
+
+#if ENABLE_UNITY_COLLECTIONS_CHECKS
+        private enum UpdaterState
+        {
+            Uninitialized,
+            Ready,
+            Modifying
+        }
+
+        private UpdaterState m_State;
+#endif
+
+        public int LaneIndex
+        {
+            get;
+            private set;
+        }
+
+        internal VDJobUpdater(UnsafeTypedStream<TInstance>.Writer continueWriter,
+                              NativeArray<TInstance> iteration)
+        {
+            m_ContinueWriter = continueWriter;
+            m_Iteration = iteration;
+
+            m_ContinueLaneWriter = default;
+            LaneIndex = DEFAULT_LANE_INDEX;
+
+#if ENABLE_UNITY_COLLECTIONS_CHECKS
+            m_State = UpdaterState.Uninitialized;
+#endif
+        }
+        
+        /// <summary>
+        /// Initializes the struct based on the thread it's being used on.
+        /// This must be called before doing anything else with the struct.
+        /// </summary>
+        /// <param name="nativeThreadIndex">The native thread index</param>
+        public void InitForThread(int nativeThreadIndex)
+        {
+#if ENABLE_UNITY_COLLECTIONS_CHECKS
+            Debug.Assert(m_State == UpdaterState.Uninitialized);
+            m_State = UpdaterState.Ready;
+#endif
+
+            LaneIndex = ParallelAccessUtil.CollectionIndexForThread(nativeThreadIndex);
+            m_ContinueLaneWriter = m_ContinueWriter.AsLaneWriter(LaneIndex);
+        }
+        
+        /// <summary>
+        /// Gets a <typeparamref name="TInstance"/> at the specified index.
+        /// </summary>
+        /// <param name="index">The index to the backing array</param>
+        public TInstance this[int index]
+        {
+            get
+            {
+#if ENABLE_UNITY_COLLECTIONS_CHECKS
+                Debug.Assert(m_State == UpdaterState.Ready);
+                m_State = UpdaterState.Modifying;
+#endif
+
+                return m_Iteration[index];
+            }
+        }
+        
+        /// <summary>
+        /// Signals that this instance should be updated again next frame.
+        /// </summary>
+        /// <param name="instance">The instance to continue</param>
+        public void Continue(TInstance instance)
+        {
+            Continue(ref instance);
+        }
+        
+        /// <inheritdoc cref="Continue(TInstance)"/>
+        public void Continue(ref TInstance instance)
+        {
+#if ENABLE_UNITY_COLLECTIONS_CHECKS
+            Debug.Assert(m_State == UpdaterState.Modifying);
+            m_State = UpdaterState.Ready;
+#endif
+            m_ContinueLaneWriter.Write(ref instance);
+        }
+
+        internal void Complete()
+        {
+#if ENABLE_UNITY_COLLECTIONS_CHECKS
+            Debug.Assert(m_State == UpdaterState.Modifying);
+            m_State = UpdaterState.Ready;
+#endif
+        }
+    }
+}
--- a/Scripts/Runtime/Data/VirtualData/Jobs/VDJobWriter.cs
+++ b/Scripts/Runtime/Data/VirtualData/Jobs/VDJobWriter.cs
@@ -1,0 +1,84 @@
+using Anvil.Unity.DOTS.Jobs;
+using Unity.Collections;
+using UnityEngine;
+
+namespace Anvil.Unity.DOTS.Data
+{
+    /// <summary>
+    /// A struct to be used in jobs that is for writing new <typeparamref name="TInstance"/> to
+    /// <see cref="VirtualData{TKey,TInstance}"/>.
+    ///
+    /// Commonly used to add new instances. 
+    /// </summary>
+    /// <typeparam name="TInstance">The type of instance to add</typeparam>
+    [BurstCompatible]
+    public struct VDJobWriter<TInstance>
+        where TInstance : struct
+    {
+        private const int DEFAULT_LANE_INDEX = -1;
+
+        [ReadOnly] private readonly UnsafeTypedStream<TInstance>.Writer m_InstanceWriter;
+
+        private UnsafeTypedStream<TInstance>.LaneWriter m_InstanceLaneWriter;
+        private int m_LaneIndex;
+
+#if ENABLE_UNITY_COLLECTIONS_CHECKS
+        private enum WriterState
+        {
+            Uninitialized,
+            Ready
+        }
+
+        private WriterState m_State;
+#endif
+
+
+        internal VDJobWriter(UnsafeTypedStream<TInstance>.Writer instanceWriter) : this()
+        {
+            m_InstanceWriter = instanceWriter;
+
+            m_InstanceLaneWriter = default;
+            m_LaneIndex = DEFAULT_LANE_INDEX;
+
+#if ENABLE_UNITY_COLLECTIONS_CHECKS
+            m_State = WriterState.Uninitialized;
+#endif
+        }
+        
+        /// <summary>
+        /// Initializes the struct based on the thread it's being used on.
+        /// This must be called before doing anything else with the struct.
+        /// </summary>
+        /// <param name="nativeThreadIndex">The native thread index</param>
+        public void InitForThread(int nativeThreadIndex)
+        {
+#if ENABLE_UNITY_COLLECTIONS_CHECKS
+            Debug.Assert(m_State == WriterState.Uninitialized);
+            m_State = WriterState.Ready;
+#endif
+
+            m_LaneIndex = ParallelAccessUtil.CollectionIndexForThread(nativeThreadIndex);
+            m_InstanceLaneWriter = m_InstanceWriter.AsLaneWriter(m_LaneIndex);
+        }
+        
+        /// <summary>
+        /// Adds the instance to the <see cref="VirtualData{TKey,TInstance}"/>'s
+        /// underlying pending collection to be added the next time the virtual data is
+        /// consolidated.
+        /// </summary>
+        /// <param name="instance">The instance to add</param>
+        public void Add(TInstance instance)
+        {
+            Add(ref instance);
+        }
+        
+        /// <inheritdoc cref="Add(TInstance)"/>
+        public void Add(ref TInstance instance)
+        {
+#if ENABLE_UNITY_COLLECTIONS_CHECKS
+            Debug.Assert(m_State == WriterState.Ready);
+#endif
+            m_InstanceLaneWriter.Write(ref instance);
+        }
+    }
+}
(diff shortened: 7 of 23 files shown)


Scripts/Runtime/Data/VirtualData/Jobs/VDJobResultsDestination.cs
```diff
@@ -0,0 +1,26 @@
+using Unity.Collections;
+
+namespace Anvil.Unity.DOTS.Data
+{
+    /// <summary>
+    /// A struct to be used in jobs that acts as a placeholder for results to be written to
+    /// later on. Use with <see cref="IVirtualDataInstance{TResult}"/>
+    /// </summary>
```
mbaker3, line 8:
```suggestion
    /// <summary>
    /// A container to write Virtual Data job results to.
    /// Use with <see cref="IVirtualDataInstance{TResult}"/>
    /// </summary>
```

Scripts/Runtime/Data/VirtualData/Jobs/VDJobResultsWriter.cs
```diff
@@ -0,0 +1,38 @@
+using Unity.Collections;
+
+namespace Anvil.Unity.DOTS.Data
+{
+    /// <summary>
+    /// A struct to be used in jobs that writes new results.
```
mbaker3, line 6:
```suggestion
    /// Represents a writable reference to VirtualData.
```

Scripts/Runtime/Data/VirtualData/Jobs/VDJobUpdater.cs
```diff
@@ -0,0 +1,121 @@
...
+        /// Signals that this instance should be updated again next frame.
+        /// </summary>
+        /// <param name="instance">The instance to continue</param>
+        public void Continue(TInstance instance)
+        {
+            Continue(ref instance);
+        }
+        
+        /// <inheritdoc cref="Continue(TInstance)"/>
+        public void Continue(ref TInstance instance)
```
mbaker3, line 104: Should this be `in` instead? Are you mutating the struct at all?

Scripts/Runtime/Data/VirtualData/Jobs/VDJobWriter.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// <see cref="VirtualData{TKey,TInstance}"/>.
+    ///
+    /// Commonly used to add new instances. 
+    /// </summary>
+    /// <typeparam name="TInstance">The type of instance to add</typeparam>
+    [BurstCompatible]
+    public struct VDJobWriter<TInstance>
+        where TInstance : struct
+    {
+        private const int DEFAULT_LANE_INDEX = -1;
```
mbaker3, line 18: Is it valid to use a lane index of -1? If not, then this should probably be `UNSET_LANE_INDEX`
(Same goes for any other places in this PR `DEFAULT_*` was used.)

Scripts/Runtime/Data/VirtualData/VirtualData.cs
```diff
@@ -0,0 +1,296 @@
...
+    /// <summary>
+    /// This class represents wrapped collections of data and manages them for use in Jobs.
+    /// </summary>
+    /// <remarks>
+    /// In Unity's ECS, data is stored on <see cref="Entity"/>'s via <see cref="IComponentBase"/> structs.
+    /// Unity's <see cref="SystemBase"/>'s handle the dependencies on the different sorts of data that are needed
+    /// for a given update call depending on the <see cref="EntityQuery"/>s used and Jobs scheduled.
+    ///
+    /// There are often cases where the overhead of using Entities doesn't make sense. The data is not persistent
+    /// and should exist for a period of time and then cease to exist. While this can be accomplished by Adding
```
mbaker3, line 22:
```suggestion
    /// There are often cases where the overhead of using Entities doesn't make sense. The data is not persistent
    /// and has a short lifecycle (<1-10 frames). While this can be accomplished by Adding
```

Scripts/Runtime/Data/VirtualData/VirtualData.cs
```diff
@@ -0,0 +1,296 @@
...
+    /// Instead, using <see cref="VirtualData{TKey,TInstance}"/> can alleviate these issues while working in a similar
+    /// manner. Additional benefits are:
+    /// 
+    /// - Allowing for Shared Write
+    /// - - Multiple different jobs can write to the Pending collection at the same time.
+    /// - Fast reading via iteration or lookup
+    /// - The ability for instances of the data to write a result to different result destinations.
+    /// - - This gives implicit grouping of the data while still allowing for processing the overall set of data
+    ///     as one large set.
+    /// - - Getting access to writing to the results is handled automatically.
```
mbaker3, line 35:
```suggestion
    /// - Allowing for parallel writing
    ///    - Multiple different jobs can write to the Pending collection at the same time.
    /// - Fast reading via iteration or individual lookup
    /// - The ability for each instance of the data to write its result to a different destination.
    ///    - This gives implicit grouping of the data while still allowing for processing the overall set of data
    ///     as one large set. (Ex: Timers update as one set but complete back to different destinations)
    ///    - Getting write to result destinations is handled automatically.
```

Scripts/Runtime/Data/VirtualData/VirtualData.cs
```diff
@@ -0,0 +1,296 @@
...
+    /// </remarks>
+    /// <typeparam name="TKey">The type of key to use to lookup data. Usually <see cref="Entity"/></typeparam>
```
mbaker3, line 37:
```suggestion
    /// <typeparam name="TKey">The type of key to use to lookup data. Usually <see cref="Entity"/> if this is being used as an alternative to adding component data to an <see cref="Entity"/>.</typeparam>
```

Scripts/Runtime/Data/VirtualData/VirtualData.cs
```diff
@@ -0,0 +1,296 @@
...
+            {
+                virtualData.AddSource(source);
+                source.AddResultDestination(virtualData);
+            }
+
+            return virtualData;
+        }
+
+        private readonly HashSet<IVirtualData> m_Sources = new HashSet<IVirtualData>();
+        private readonly HashSet<IVirtualData> m_ResultDestinations = new HashSet<IVirtualData>();
```
mbaker3, line 64: Can we be consistent about what gets initialized in constructor vs declaration?

Scripts/Runtime/Data/VirtualData/VirtualData.cs
```diff
@@ -0,0 +1,296 @@
...
+
+        //*************************************************************************************************************
+        // SERIALIZATION
+        //*************************************************************************************************************
+
+        //TODO: Add Serialization and Deserialization functions to hook into our serialization framework
+
+        //*************************************************************************************************************
+        // RELATIONSHIPS
+        //*************************************************************************************************************
```
mbaker3, line 119: mayyyybe. Hopefully we can do it all from the outside, with extension methods or wrapper objects.

Scripts/Runtime/Data/VirtualData/VirtualData.cs
```diff
@@ -0,0 +1,296 @@
...
+
+        void IVirtualData.AddResultDestination(IVirtualData resultData)
+        {
+            AddResultDestination(resultData);
+        }
+
+        private void AddResultDestination(IVirtualData resultData)
+        {
+            m_ResultDestinations.Add(resultData);
+        }
```
mbaker3, line 129: I don't understand. Why not just make this internal to satisfy the interface and omit the proxy method?

Same applies to the rest of times this happens

mbaker3, line 129: How often do these destinations and sources change? 
I get the impression you'd add/remove a source and destination maybe once per session?

But you'll iterate constantly. Should we instead optimize the collections for iteration? (NativeArrays or List)

I could very well be misunderstanding the use case though

Scripts/Runtime/Data/VirtualData/VirtualDataExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// <typeparam name="TInstance">The type of the instance struct.</typeparam>
+        /// <typeparam name="TResult">The type of the result struct</typeparam>
+        public static void Complete<TKey, TInstance, TResult>(this TInstance instance, ref TResult result, ref VDJobUpdater<TKey, TInstance> jobUpdater)
+            where TKey : struct, IEquatable<TKey>
+            where TInstance : struct, ILookupData<TKey>, IVirtualDataInstance<TResult>
+            where TResult : struct
+        {
+            VDJobResultsWriter<TResult> resultsWriter = instance.ResultsDestination;
+            resultsWriter.Add(ref result, jobUpdater.LaneIndex);
+            jobUpdater.Complete();
```
mbaker3, line 29: Now that I see this in context I wonder if we should come up with a different name than `Complete()` for the method. Particularly since the word job is in there it makes me think this is a sync point.

Scripts/Runtime/Data/VirtualData/IVirtualData.cs
```diff
@@ -0,0 +1,26 @@
...
+        
+        internal JobHandle ConsolidateForFrame(JobHandle dependsOn);
+
+        internal JobHandle AcquireForUpdate();
+        internal void ReleaseForUpdate(JobHandle releaseAccessDependency);
+
+        internal AccessController AccessController
+        {
+            get;
+        }
```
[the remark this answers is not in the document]
jkeon: In my mind, no because you as the developer will never use this directly. 

If it gets to the point where we do expose it, then yeah docs will be needed

Scripts/Runtime/Data/VirtualData/InstanceData/ILookupData.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+
+namespace Anvil.Unity.DOTS.Data
+{
+    /// <summary>
+    /// Data to be used in <see cref="VirtualData{TKey,TInstance}"/>
+    /// that can be looked up with a <typeparamref name="TKey"/>
+    /// </summary>
+    /// <typeparam name="TKey">The type of key to use for lookup.</typeparam>
+    public interface ILookupData<out TKey>
```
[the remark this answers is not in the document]
jkeon: I went with `IKeyedData`

Scripts/Runtime/Data/VirtualData/Jobs/VDJobReader.cs
```diff
@@ -0,0 +1,30 @@
+using Unity.Collections;
+
+namespace Anvil.Unity.DOTS.Data
+{
+    /// <summary>
+    /// A struct to be used in jobs that only allows for reading.
+    /// Commonly used to read and trigger something else.
+    /// </summary>
```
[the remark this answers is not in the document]
jkeon: Updated

Scripts/Runtime/Data/VirtualData/Jobs/VDJobReader.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <typeparam name="TInstance">They type of data to read</typeparam>
+    [BurstCompatible]
+    public readonly struct VDJobReader<TInstance>
```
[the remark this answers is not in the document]
jkeon: Fair point and open to discussion. 

For me, it's a consistency and intended usage thing. The `VDJob*ers` are specific structs that are given to your jobs to allow you to interact with the `VirtualData` in a context that reflects the intention and access control you have over the data. 

I suppose you don't have to do it via a job but we don't have main thread API's yet... but ok, I can see dropping the "Job" part. Changed.

I could use the `NativeArray<T>.ReadOnly` but then that brings along a bunch of other cruft we don't need and makes it inconsistent with the other `VD*ers` structs. I'd rather you had a familiar api and types that all felt like they go together in your jobs.

Scripts/Runtime/Data/VirtualData/Jobs/VDJobReader.cs
```diff
@@ -0,0 +1,30 @@
...
+        where TInstance : struct
+    {
+        [ReadOnly] private readonly NativeArray<TInstance> m_Iteration;
```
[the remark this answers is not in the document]
jkeon: Sure, changed

Scripts/Runtime/Data/VirtualData/Jobs/VDJobResultsDestination.cs
jkeon: I reworked this.

Scripts/Runtime/Data/VirtualData/Jobs/VDJobResultsDestination.cs
```diff
@@ -0,0 +1,26 @@
...
+    /// <typeparam name="TResult">The type of result that can be written</typeparam>
+    [BurstCompatible]
+    public readonly struct VDJobResultsDestination<TResult>
```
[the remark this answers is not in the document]
jkeon: This holds the reference to the writer but doesn't allow you to do anything with it. You cannot read. You cannot write. You can only pass the reference around.

Behind the scenes, no access control has been acquired for it either. 

This means that as the developer, you can't shoot yourself in the foot by trying to write when you get this in your job. The only thing you can do is assign it to an `IVirtualDataInstance` as the bucket to write it's result to. 

When you call `Complete` on a `IVirtualDataInstance` we implicit cast to the `VDResultsWriter` which we are now allowed to write to. 

We can only call `Complete` when we have a `VDUpdater` and that ensures that we have gotten the access control to actually write to the backing `VirtualData` the `VDResultsWriter` references. 

In short, we're just passing around a `UnsafeTypedStream<TResult>.Writer` but having it naked is dangerous. The two different structs represent and enforce what you can do with it and ensure that it is safe to do and access has been acquired.

Scripts/Runtime/Data/VirtualData/Jobs/VDJobResultsWriter.cs
jkeon: Reworked.

Scripts/Runtime/Data/VirtualData/Jobs/VDJobUpdater.cs
```diff
@@ -0,0 +1,121 @@
...
+            }
+        }
+        
+        /// <summary>
+        /// Signals that this instance should be updated again next frame.
+        /// </summary>
+        /// <param name="instance">The instance to continue</param>
+        public void Continue(TInstance instance)
+        {
+            Continue(ref instance);
```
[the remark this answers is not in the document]
jkeon: I'm not sure how Burst factors into this, but a lot of the internal Unity code is using `ref` when passing around structs.

Additionally, because of our layers of abstraction, `Continue` then calls `Write` which then calls `UnsafeUtility.CopyStructureToPtr`. So instead of multiple copies, we can just pass the same pointer all the way down which is what it wants at the end anyway. 

Agreed that in C# if the struct is small, then it's better to just copy, but multiple copies? Seems like that might get into the tiny allocate/deallocate cost on the stack mentioned here: https://stackoverflow.com/questions/46291828/when-is-it-more-efficient-to-pass-structs-by-value-and-when-by-ref-in-c

Scripts/Runtime/Data/VirtualData/Jobs/VDJobUpdater.cs
jkeon: We are not, but `in` is only useful if the struct is a `readonly` struct. We can't guarantee that so `in` will end up doing a defensive copy. 

The end api of `UnsafeUtility.CopyStructureToPtr` wants a ref so it makes sense to pass it all along the chain.

Scripts/Runtime/Data/VirtualData/Jobs/VDJobWriter.cs
jkeon: Yep good point, fixed.

Scripts/Runtime/Data/VirtualData/VirtualData.cs
jkeon: I don't think that's accurate. 

A Timer could exist for 1 frame or 10 minutes.
Following a path could exist for a few seconds or a few minutes.

My point that I was trying to make is that it doesn't make sense to have it be on the Actor Entity because you won't ALWAYS have a path to follow or a timer going on. (Also in the timer case, you might want multiple timers so a component wouldn't work anyway). 

And you don't want to be a standalone entity that is somehow linked due to chunk thrashing as these overall collections blink in and out of existence.

Scripts/Runtime/Data/VirtualData/VirtualData.cs
```diff
@@ -0,0 +1,296 @@
...
+    /// This class represents wrapped collections of data and manages them for use in Jobs.
+    /// </summary>
+    /// <remarks>
+    /// In Unity's ECS, data is stored on <see cref="Entity"/>'s via <see cref="IComponentBase"/> structs.
+    /// Unity's <see cref="SystemBase"/>'s handle the dependencies on the different sorts of data that are needed
+    /// for a given update call depending on the <see cref="EntityQuery"/>s used and Jobs scheduled.
+    ///
+    /// There are often cases where the overhead of using Entities doesn't make sense. The data is not persistent
+    /// and should exist for a period of time and then cease to exist. While this can be accomplished by Adding
+    /// or Removing Components from an Entity or spawning/destroying new entities with those components, it results
```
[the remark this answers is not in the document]
jkeon: Fixed.

Scripts/Runtime/Data/VirtualData/VirtualData.cs
```diff
@@ -0,0 +1,296 @@
...
+    /// in a structural change which gets resolved at a sync point and happens on the main thread.
```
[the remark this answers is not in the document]
jkeon: Fixed.

Scripts/Runtime/Data/VirtualData/VirtualData.cs
```diff
@@ -0,0 +1,296 @@
...
+    ///
+    /// Instead, using <see cref="VirtualData{TKey,TInstance}"/> can alleviate these issues while working in a similar
+    /// manner. Additional benefits are:
+    ///
```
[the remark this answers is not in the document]
jkeon: Fixed.

Scripts/Runtime/Data/VirtualData/VirtualData.cs
jkeon: Fixed.

Scripts/Runtime/Data/VirtualData/VirtualData.cs
jkeon: Fixed.

Scripts/Runtime/Data/VirtualData/VirtualData.cs
jkeon: Yes.... I've moved to the constructor. 

I don't know why but I always do Collections as declarations and everything else in the constructor. Weird habit.

Scripts/Runtime/Data/VirtualData/VirtualData.cs
jkeon: Is this comment meant to be for the Serialization?

Scripts/Runtime/Data/VirtualData/VirtualData.cs
jkeon: Not allowed. 

https://stackoverflow.com/questions/68592353/why-cannot-implicitly-implement-a-non-public-interface-member

I need the interface because I need to store and operate on `VirtualData<TKey,TInstance>` as a collection. 

I suppose I could make an `AbstractVirtualData` instead that has the internal methods on it marked as abstract and then implement and seal them if you think that's better?

jkeon: Yes you are correct. Switched to `List`

Scripts/Runtime/Data/VirtualData/VirtualData.cs
```diff
@@ -0,0 +1,296 @@
...
+            }
+
+            public void Execute()
+            {
+                m_Lookup.Clear();
+                m_Iteration.Clear();
+
+                NativeArray<TInstance> iterationArray = m_Iteration.DeferredCreate(m_Pending.Count());
+                m_Pending.CopyTo(ref iterationArray);
+                m_Pending.Clear();
```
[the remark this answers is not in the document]
jkeon: Sure let's chat tomorrow. 

Here's what's happening:

- The previous frame's data is out of date so we clear both our Lookup and our Iteration collections. 
- Now we allocate the memory needed for the total amount of elements by looking at the count in Pending. 
    - Pending represents any new elements AND any elements from last frame that were `Continue`ed.
- With the array sized, we use the fast `CopyTo` which blits the blocks of memory to the array instead of going one by one.
- We then clear Pending because we'll want to write to it again this frame.
- We then need to populate the Lookup. But the lookup doesn't have a fast blit method and we can't bypass it either because it stores things in blocks. 
    - Array iteration is super fast and we already have our array so we loop through it and use the fastest add function for adding to the lookup.
- We now have a Lookup and an Iteration that contain all the same elements and an empty primed Pending collection. We are ready for any jobs this frame that need to read or update or add new elements or continue existing ones to the next frame.

Scripts/Runtime/Data/VirtualData/VirtualDataExtensions.cs
jkeon: Let's chat about it. I agree the naming has been the hardest part. 

Here's the context for coming up with the name - see `ContinueOn` and `Complete`

```
            public void Execute(int index)
            {
                TimerInstance timerInstance = m_Timers[index];
                timerInstance.RemainingTime -= m_DeltaTime;
                if (timerInstance.RemainingTime > 0.0f)
                {
                    timerInstance.ContinueOn(ref m_Timers);
                    return;
                }

                timerInstance.Complete(new TimerResult(timerInstance.Key), ref m_Timers);
            }
```

jkeon:
I updated the docs to remove instances of "this struct" or "this class". 

I think in my mind I was trying to alert that "hey watch yourself, this is going to be copied" but yeah you can just see that by the definition.

jkeon:
@mbaker3 Everything addressed except for the ones that require a call.

Scripts/Runtime/Data/VirtualData/IVirtualData.cs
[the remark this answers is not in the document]
mbaker3: Yea just didn't know if developer was potentially a library developer but I'm ok with the rationale.

Scripts/Runtime/Data/VirtualData/Jobs/VDJobResultsDestination.cs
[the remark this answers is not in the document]
mbaker3: ok, with the implicit cast isn't it just as dangerous that someone casts it? Do we need to assert a condition be met to actually convert to a `VDJobResultsWriter`?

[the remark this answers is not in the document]
jkeon: It is, which is why I removed the implicit cast. It's now a method to convert that can only happen when safe to do so.

Scripts/Runtime/Data/VirtualData/Jobs/VDJobUpdater.cs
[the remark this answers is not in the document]
mbaker3: That's fair. You lose data locality passing refs but unless you're storing that ref in a big collection and iterating later it's not a big deal.

Scripts/Runtime/Data/VirtualData/VirtualData.cs
mbaker3: The decision on lifecyle length isn't about covering all cases just the most common one. If your timer's are usually 10mins then it probably does make more sense to use components on entities. Obviously we don't build the timer system to assume this since most timers are indeed short. The point of VirtualData is to remove the overhead of potentially short lived data right?

Or are you saying that Entity+Component should never have a structural change and the entity should exist permanently in a session. If any aspect of an entity is going to change throughout its lifetime it should be VD. 
If that's the case then what in a game would actually be Entity + Component? Nearly everything has a lifecycle. 

Should a crew member be totally VirtualData and not bound to an entity since they can die?

I think what I'm getting at is that this documentation needs to specify when VD makes sense. Reading "should exist for a period of time and then cease to exist" doesn't tell me anything as a developer. Everything in an application meets that definition.

> And you don't want to be a standalone entity that is somehow linked due to chunk thrashing as these overall collections blink in and out of existence.

If it's not a short lifecycle what's the issue with thrashing the chunk? It's not like it will become fragmented if you have a reasonable amount of churn.

Scripts/Runtime/Data/VirtualData/VirtualData.cs
mbaker3: Yes

Scripts/Runtime/Data/VirtualData/Jobs/VDJobUpdater.cs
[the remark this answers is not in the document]
mbaker3: Seems right to me.

Scripts/Runtime/Data/VirtualData/VirtualData.cs
jkeon: > The point of VirtualData is to remove the overhead of potentially short lived data right?

Somewhat. You should use VirtualData when the data is shortlived or when you have multiple different "completion" avenues or when you want to avoid large entities etc.

> Or are you saying that Entity+Component should never have a structural change and the entity should exist permanently in a session. If any aspect of an entity is going to change throughout its lifetime it should be VD.
> If that's the case then what in a game would actually be Entity + Component? Nearly everything has a lifecycle.
> 
> Should a crew member be totally VirtualData and not bound to an entity since they can die?

In my mind, I would use Entity+Component to represent an Actor and it's corresponding render information because that's what the URP is expecting. 

I'm not sure I see a benefit to using it for anything else right now. Maybe the DynamicBuffer consolidation of Attributes. 

> I think what I'm getting at is that this documentation needs to specify when VD makes sense. Reading "should exist for a period of time and then cease to exist" doesn't tell me anything as a developer. Everything in an application meets that definition.

Sure, ok. I'll try and reword a bit. In essence, use `VirtualData` first unless it is easier to use Entity+Components for your specific use case. I would argue that most cases are not.

> If it's not a short lifecycle what's the issue with thrashing the chunk? It's not like it will become fragmented if you have a reasonable amount of churn.

I observed a ton of chunk thrashing when the lifespan was variable. Take timers and following paths for example. 
When both were Entity+Component some timers are short (1.0 seconds) and some are longer (3.0 seconds). Some paths are a few tiles away, some are the whole map away. 

I would have say 10 chunks of these entities and on frame 1 all of them are full which makes sense. Then as Timers complete or paths complete, pieces of the chunks will fall out. 

What you would expect is that as new ones come into existence it would fill up the empty space in the first chunk and then fill up the second chunk etc and eventually reach a point where the longest lasting Timers or Paths in the 10th chunk are all gone and on average we have let's say 6 chunks in flight. And that last chunk would ebb and flow as things came in and out of existence. 

But in practice that's not what happened. All 10 chunks remained and all 10 ebbed and flowed throughout execution. I'm not sure how Unity is handling the recycling of Entities but it seemed to keep the chunks open and have the churn.

Scripts/Runtime/Data/VirtualData/VirtualData.cs
jkeon: I'll update the TODO to reference the hopeful aspect.

Scripts/Runtime/Data/VirtualData/VirtualData.cs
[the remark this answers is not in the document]
mbaker3: hmmm, probably not worth testing this micro optimization but since you have to perform a for iteration anyway I wonder if Blit Copy is actually faster than assigning one element at a time.

Also for another day, `UnsafeHashMap` doesn't have a bulk add method but I bet one could be implemented that mem copied the entire value array and then did a mem copy with a stride to extract the keys.

Scripts/Runtime/Data/VirtualData/VirtualDataExtensions.cs
mbaker3: Some naming ideas:
 - Resolve (stolen from Promises)
 - Conclude
 - Finalize (probably not good either given existing meanings)
 - Fulfill
 - Done
 - FInish

Alternatively if we want to change continue you could come up with a different analogy:
Instead of ContinueOn/Complete
 - RetainWith/Release
 - PersistOn/Release
 - ScheduleUpdateWith/Conclude

I think my fave is just changing Complete to Resolve or doing Retain/Release.

Scripts/Runtime/Data/VirtualData/VirtualData.cs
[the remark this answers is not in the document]
mbaker3: ah ok, makes sense.

Yea hold off on HashMap until some other day. Maybe make an issue if you really want to get into it some day.

Scripts/Runtime/Data/VirtualData/VirtualDataExtensions.cs
jkeon: Cool, changing it to `Resolve` and keeping `Continue`

Scripts/Runtime/Data/VirtualData/VirtualData.cs
mbaker3: > > The point of VirtualData is to remove the overhead of potentially short lived data right?
> 
> Somewhat. You should use VirtualData when the data is shortlived or when you have multiple different "completion" avenues or when you want to avoid large entities etc.

OK, the docs need to define that as well as what you mean by "shortlived"


> I observed a ton of chunk thrashing when the lifespan was variable. Take timers and following paths for example.

That's strange because the forum posts seem to indicate that chunks will have elements backfill their gaps.

jkeon: > OK, the docs need to define that as well as what you mean by "shortlived"

Yep docs are updated.

> That's strange because the forum posts seem to indicate that chunks will have elements backfill their gaps.

This is what I thought too.

mbaker3 (approved):

PR #41, opened 2022-06-30
Entities 0.51-preview.32 API updates

mbaker3:
This PR covers updates to the API that were made to be compatible with Entities 0.51-preview.32

The library does not compile

The library compiles

None

 - None

 - [x] Yes - Update your project to Entities 0.51-preview.32
 - [ ] No


jkeon (approved):
Looks fine but maybe we should upgrade TwistedDots at the same time?

mbaker3:
> Looks fine but maybe we should upgrade TwistedDots at the same time?

Go for it? Up to you. I'm not working in that project right now.

PR #43, opened 2022-07-12
DRAFT: Cancel Task Drivers

jkeon:
This is a preview of the Cancel functionality for TaskDrivers/Systems and how we handle it. 

I've called out in line specific discussion points. 

For everything else, it's good to read but don't worry too much about in depth review since our discussions may change it. 

For an overview of how everything works, it's probably best to walk you through the updated Figma diagrams which I can include here in the actual PR once we get to that.

The whole thing is a bit taxing on the mind but the diagrams really help.


Scripts/Runtime/Data/VirtualData/InstanceData/IKeyedData.cs
```diff
@@ -1,21 +1,20 @@
...
+    public interface IKeyedData
     {
         /// <summary>
         /// The lookup key for this data
         /// </summary>
-        public TKey Key
+        public VDContextID ContextID
         {
             get;
+            //TODO: DISCUSS - Way to hide this set. Might not be possible.
```
jkeon, line 16: I don't think there is a way to do this, but I'll struggled long enough that it's worth talking about. 

I don't want you to be able to set the `ContextID` from outside. Only internally when I'm doing something rewrite the context part out from underneath you as the data flows across different TaskDrivers. 

I can't make it an `internal set` because then whoever implements this interface in the game code is no longer internal and can't compile.

I've tried some pointer magic to get the memory for the ContextID and then reach in and change the context but the problem is that because it's a getter, it returns a copy of the backing field. My changes work but only on that copy, the original backing field is still the same. 

We could get around that with a ref return, but of course structs aren't allowed to do that. 

So our options are this:

1. Add a set. You shouldn't use it, you'll probably break things if you do and there's no way we can guard against it. However similarly `Entity` lets you modify the `Index` and `Version` fields and you really shouldn't ever do that either.

2. Make the entire game assembly allow unsafe code. Have this interface have a function that returns a pointer to the backing field. You the developer would have to copy paste that code to return the proper pointer into every VirtualData instance struct you ever create... I think it's gross. But it does work.

Scripts/Runtime/Data/VirtualData/VirtualData.cs
```diff
@@ -1,11 +1,15 @@
...
 using Anvil.Unity.DOTS.Jobs;
 using System;
+using System.Diagnostics;
 using Unity.Burst;
 using Unity.Collections;
 using Unity.Collections.LowLevel.Unsafe;
+using Unity.Entities;
 using Unity.Jobs;
 
+//TODO: DISCUSS - Namespace
```
jkeon, line 11: VirtualData is now pretty tied into Entities. It's only used by TaskDrivers which are only used by TaskDriverSystems and we only ever use `Entity` as part of the key for VirtualData. 

Thoughts on namespace re-org?

Scripts/Runtime/Data/VirtualData/VirtualData.cs
```diff
@@ -143,35 +158,156 @@ internal sealed override JobHandle ConsolidateForFrame(JobHandle dependsOn)
...
-        private struct ConsolidateLookupJob : IJob
+        private struct ConsolidateLookupJob : IAnvilJob
         {
             private UnsafeTypedStream<TInstance> m_Pending;
-            private DeferredNativeArray<TInstance> m_Iteration;
-            private UnsafeParallelHashMap<TKey, TInstance> m_Lookup;
+            private DeferredNativeArray<TInstance> m_IterationTarget;
+            private UnsafeParallelHashMap<VDContextID, TInstance> m_Lookup;
+            private VDLookupReader<bool> m_CancelledLookup;
+            //TODO: DISCUSS - Split this out into separate data
```
jkeon, line 167: This one requires a bit of forward thinking to actual use cases rather than my contrived Wander example. 

In the Wander, we have two competing flows. One where the Actor waits a bit, chooses a spot on the map, pathfinds to that spot and then walks to it and when they get there, repeats and starts waiting again.

The other flow is just a timer where once it expires, it cancels that Actor's Wander. Which means that the following structs get wiped (if they exist) for that Actor's Entity.

- WanderTaskInstance
- TimerInstance/Result
- PathfindRequest/Result
- FollowpathInstance/Result

That's all fine and expected. The WanderTaskDriver has a job to handle when it gets cancelled where it wants to restart itself. 

```
ConfigureCancelJob(HandleCancelJob.Schedule)
               .ScheduleOnForCancel(System.WanderInstanceData, BatchStrategy.MaximizeChunk)
               .RequireCancelledDataForIterateAsync(System.WanderInstanceData)
               .RequireDataForAddAsync(WanderStartData)
               .RequireDataForAddAsync(System.WanderInstanceData);
```

Here's we're using the cancelled instances of `WanderInstanceData` as the trigger for the job and what we want to iterate through. 

However since we want to restart, we need to create a new instance of `WanderInstanceData` and now we have a conflict in trying to access the same data in two different contexts in the same job. 

To get around it we can do an intermediate job to trigger the restart like so:

```
ConfigureCancelJob(HandleCancelJob.Schedule)
               .ScheduleOnForCancel(System.WanderInstanceData, BatchStrategy.MaximizeChunk)
               .RequireCancelledDataForIterateAsync(System.WanderInstanceData)
               .RequireDataForAddAsync(RestartData);
```

With a second job to handle the restart data to ultimately restart

```
ConfigureUpdateJob(RestartWanderJob.Schedule)
               .ScheduleOn(RestartData, BatchStrategy.MaximizeChunk)
               .RequireDataForIterateAsync(RestartData)
               .RequireDataForAddAsync(System.WanderInstanceData)
               .RequireDataForAddAsync(WanderStartData);
```

It works, but it's annoying and tedious. 

There's no reason the first job schedule where we read and write couldn't be done since they are different underlying data structures, they just share the same AccessController right now as we've grouped them together into a `VirtualData`. 

The question is "Should we split out the Cancelled aspects into it's own separate data structure".

We'd be introducing another job to handle consolidating that structure but we'd be saving on the ease of use when we want to use Cancelled Instances as a trigger to spawn new instances of the same type. 

This might dove tail into the below comment about consolidation in general.

Scripts/Runtime/Data/VirtualData/VirtualDataIntent.cs
```diff
@@ -0,0 +1,10 @@
+namespace Anvil.Unity.DOTS.Data
+{
+    public enum VirtualDataIntent
+    {
+        //TODO: DISCUSS - Data usage intent
```
jkeon, line 5: Data comes in three flavours. 

1. Default - This is data like a TimerInstance or FollowPathInstance where you create it and it exists for as long as there is work to be done on it. It gets updated every frame and then eventually gets Resolved to a result.
2. OneShot - This is like an event. It exists for one frame so someone else can pick it up and spawn whatever they need to. Then this data is gone. No results, just poof.
3. Persistent - This is data that represents State. It exists and maybe gets updated with some new info and looked up when needed but it's not "processed" per frame. It just needs to stay alive. 

I'm not sure about the names. I'm also not sure about using an enum and manually opting into one of these. 
I could make them interfaces instead and you just tag your data so that we can handle the cases. 

`public struct TimerInstance : IKeyedData, IDefaultData`
`public struct WanderStart: IKeyedData, IOneShotData`
`public struct WanderTaskInstance: IKeyedData, IPersistentData`

Thoughts?

Scripts/Runtime/Entities/Tasks/AbstractTaskDriverSystem.cs
```diff
@@ -12,89 +12,139 @@ namespace Anvil.Unity.DOTS.Entities
...
+
             //Note: We don't dispose TaskDrivers here because their parent or direct reference will do so. 
             m_TaskDrivers.Clear();
 
             base.OnDestroy();
         }
 
+        internal int GetNextID()
+        {
+            //TODO: DISCUSS - Make this robust
```
jkeon, line 59: While very unlikely since this ID represents over 2 billion TaskDrivers tied to a specific System and TaskDrivers don't really get created/destroyed at runtime.... there's still a tiny chance of overflow. 

Should be bother making an IDProvider that keeps track of open ids? Or just keep like so and add an issue to make one in the future?

Scripts/Runtime/Entities/Tasks/AbstractTaskDriverSystem.cs
```diff
@@ -12,89 +12,139 @@ namespace Anvil.Unity.DOTS.Entities
...
         {
             Debug_EnsureTaskDriverSystemRelationship(taskDriver);
             m_TaskDrivers.Add(taskDriver);
         }
-        
+
         protected override void OnUpdate()
         {
+            JobTaskWorkConfig.Debug_EnsureNoDataLoss(this, m_InstanceDataLookup, m_UpdateJobData, ref m_HasCheckedUpdateJobsForDataLoss);
+            //TODO: DISCUSS - Should we ensure no CancelData loss?
```
jkeon, line 113: If you have Data on your TaskDriver or TaskDriverSystem that they own and it's not a `OneShot`, it needs to be updated and either resolved or written back to the Pending so it sticks around.

Otherwise you have data loss. 

We can trap against this with the Debug function which ensures we have a job scheduled to take care of it. If the data is Persistent we automatically create a `KeepPersistentJob` for it.

When data is cancelled though, we're currently not do anything about it from a safety perspective. 

You're welcome to schedule a HandleCancelJob so you can do something with the instances if you want but otherwise they just "poof" out of existence which is normally what you want. 

The question is whether we should enforce the developer to opt into a behaviour to explicitly deal with cancelled instances.

Scripts/Runtime/Entities/Tasks/Data/JobTaskWorkConfig.cs
```diff
@@ -40,12 +41,20 @@ internal JobTaskWorkConfig(ScheduleJobDelegate scheduleJobDelegate, AbstractTask
...
-            where TInstance : unmanaged, IKeyedData<TKey>
+        public JobTaskWorkConfig ScheduleOn<TInstance>(VirtualData<TInstance> data, BatchStrategy batchStrategy)
+            where TInstance : unmanaged, IKeyedData
+        {
+            Debug_EnsureNoDuplicateScheduleInfo();
+            m_ScheduleInfo = new VirtualDataScheduleInfo<TInstance>(data, batchStrategy, false);
+            return this;
+        }
+        
+        //TODO: DISCUSS - CancelTaskWorkConfig vs JobTaskWorkConfig and abstract
```
jkeon, line 52: Should we separate into a CancelTaskWorkConfig? The more I'm working with this stuff on the developer end, the more it feels like we should lock down the types to only the functionality they should be used for.

`CancelTaskWorkConfig` -> `AbstractJobTaskWorkConfig` -> `AbstractTaskWorkConfig`
`JobTaskWorkConfig` -> `AbstractJobTaskWorkConfig` -> `AbstractTaskWorkConfig`
`MainTaskWorkConfig` -> `AbstractTaskWorkConfig`

Scripts/Runtime/Entities/Tasks/Data/JobTaskWorkConfig.cs
```diff
@@ -40,12 +41,20 @@ internal JobTaskWorkConfig(ScheduleJobDelegate scheduleJobDelegate, AbstractTask
         /// <typeparam name="TKey">The type of the key</typeparam>
         /// <typeparam name="TInstance">The type of the data</typeparam>
         /// <returns>This <see cref="JobTaskWorkConfig"/> for chaining additional configuration.</returns>
-        public JobTaskWorkConfig ScheduleOn<TKey, TInstance>(VirtualData<TKey, TInstance> data, BatchStrategy batchStrategy)
-            where TKey : unmanaged, IEquatable<TKey>
-            where TInstance : unmanaged, IKeyedData<TKey>
+        public JobTaskWorkConfig ScheduleOn<TInstance>(VirtualData<TInstance> data, BatchStrategy batchStrategy)
```
jkeon, line 44:
```
ConfigureCancelJob(HandleCancelJob.Schedule)
               .ScheduleOnForCancel(System.WanderInstanceData, BatchStrategy.MaximizeChunk)
               .RequireCancelledDataForIterateAsync(System.WanderInstanceData)
               .RequireDataForAddAsync(RestartData);
```

Should we optimize this and automatically do the `ScheduleOnForCancel` part and `RequireCancelledDataForIterateAsync` for you?

```
ConfigureCancelJob(HandleCancelJob.Schedule, System.WanderInstanceData, BatchStrategy.MaximizeChunk)
               .RequireDataForAddAsync(RestartData);
```

It's a bit nicer/easier and you won't forget, but you don't have as many hints for writing your job about what you need to get to use.

jkeon:
@mbaker3 All comments inline added for discussion. 

Ping me when you want to go through it all together.

Scripts/Runtime/Data/VirtualData/VDContextID.cs
```diff
@@ -0,0 +1,78 @@
...
+        public override bool Equals(object compare)
+        {
+            return compare is VDContextID id && Equals(id);
+        }
+
+        public override int GetHashCode()
+        {
+            //Taken from ValueTuple.cs GetHashCode. 
+            //https://github.com/dotnet/roslyn/blob/main/src/Compilers/Test/Resources/Core/NetFX/ValueTuple/ValueTuple.cs
+            //Unfortunately we can't use directly because it has a static Random class it creates which doesn't jive with Burst
```
mbaker3, line 58: - [ ] double check the license on this code.

Scripts/Runtime/Data/VirtualData/VirtualData.cs
mbaker3: Hmmm, given what we discussed on the call with the ability to progress through a graceful cancel before informing your parent that you have cancelled I'm wondering if we treat cancellation as a form of failed resolution in the same way we resolve as failed when a task can't be completed (Ex: no path exists).

We'd just need to come up with a way to decern between failure to complete because impossible vs failure to complete because of cancel. If the parent could already be aware (some flag set) that it is within a cancellation flow then we don't need to write a reason for task failure (cancel vs impossible) into the result.


If you don't think what we discussed changes the nature of this problem then are we able to wrap types? Instead of a two phase job what if you had a generic struct that could hold the task data? 
Ex: `Cancelled<System.WanderInstanceData>`
That way you are dealing with different buckets without competing on the access controller?

Scripts/Runtime/Data/VirtualData/VirtualDataIntent.cs
mbaker3: Interfaces are troublesome because you could stack multiple interfaces on a type and then what would you do?

Are there actual optimizations behind these flavours or is it purely dev convenience?

Scripts/Runtime/Entities/Tasks/AbstractTaskDriverSystem.cs
mbaker3: Fine for now, make a TODO but

 - Make sure an overflow exception will actually get thrown so we catch this in the moment not as weird behaviour emerges. I DO NOT want to debug that!
 - If there is no overflow exception then throw your own exception
 - I don't see where `m_TaskDriverID` is getting initialized but shouldn't you be using `uint` and initializing to 0?

In a pinch we can always kick the can down the road and switch to `long` ids 😆

Scripts/Runtime/Entities/Tasks/AbstractTaskDriverSystem.cs
mbaker3: Developer should be forced to explicitly define what the cancel behaviour is. Even if annoying it's better than forgetting to think through a cancel situation.

Back to my other question though, is there an actual optimization path to OneShot? If not, then let's just have every task required to write a result. Just like in the application code there's always a return code. I don't see the purpose to a task that you fire and forget but then for some reason want to be able to cancel.

Scripts/Runtime/Entities/Tasks/Data/JobTaskWorkConfig.cs
mbaker3: I'm going to need you to refresh me on what other types of `Configure*Job`s there are.
I like the first option better because it's more consistent. However having a unique `.ScheduleOnForCancel` method seems wrong. Why can't we use `ScheduleOn` for this?

Scripts/Runtime/Entities/Tasks/Data/JobTaskWorkConfig.cs
mbaker3: What's unique about cancel. Why can't it be a JobTaskWorkConfig?

Scripts/Runtime/Data/VirtualData/InstanceData/IKeyedData.cs
jkeon: In your DotFiddle you are returning a ref to a struct from a class. 
I need to return a ref to a struct from a struct. https://dotnetfiddle.net/TYUo3D


The core issue is that you as the Developer need to create structs to represent your data. This data can be and will be anything `TimerInstance`, `WanderInstance`, `PickupTarget` etc etc etc. 

We need to store that data in a lookup and we need to associate it with an `Entity` and we need to know which `TaskDriver` it came from. 

If it was a class, no problem it all just inherits from `AbstractData` and off we go. But since it's a struct best we can do is make you implement an interface to return an inner struct for an ID. In this case a `VDContextID`. 

Because it's an interface, it must be a method/getter. Which means it always returns a copy which makes it hard to inject a new context. 

I can't make any of the data internal because you as the dev need to actually implement. 

The constructor that takes the context is internal but I need one public one so that you can seed it with your Entity. 

The only way I can lock it down nicely is to make the developer implement the VDContextID as a field on the struct and then also implement a method that returns a pointer to that field. There's no way to enforce that they did it correctly, so I just have to hope they did it right for every single data structure they create. It's a lot of boilerplate.

Scripts/Runtime/Data/VirtualData/VirtualData.cs
jkeon: I don't believe so anymore because the benefit in using VirtualData is via the TaskDrivers which are run by Systems. 

IF a usecase comes up in the future, we could generic VirtualData again so that the Key is not an Entity but I don't think it ever will.

Scripts/Runtime/Data/VirtualData/VirtualData.cs
```diff
@@ -143,35 +158,156 @@ internal sealed override JobHandle ConsolidateForFrame(JobHandle dependsOn)
...
                 m_Lookup.Clear();
-                m_Iteration.Clear();
+                m_IterationTarget.Clear();
+                m_CancelledIterationTarget.Clear();
 
-                NativeArray<TInstance> iterationArray = m_Iteration.DeferredCreate(m_Pending.Count());
-                m_Pending.CopyTo(ref iterationArray);
+                //Get the new counts
+                int pendingCancelledCount = m_CancelledLookup.Count();
+                int pendingCount = m_Pending.Count();
```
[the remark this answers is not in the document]
jkeon: > Why do this? If you have to iterate anyway why not just interate and do the work?

We don't have to iterate, the add/removes are what are iterated. We just need to copy all the contents of the lookup into a flat array after for load balancing and this provides a fast way to do that.

> Can you use removeSwapBack to cull all the cancelled tasks right off the update list and then perform your pending adds to minimize growth/shrink required?

Yes, we'll always process removes first and then adds. The collection will only grow, but never shrink capacity.

> Can you record the index of each task to be cancelled and pass that into the iterator for the Job. On each iteration it checks to see if a given index should be skipped and skips that job for update?

Possibly, I'll explore that.

Scripts/Runtime/Entities/Tasks/AbstractTaskDriverSystem.cs
jkeon: Cool, sounds good.

Scripts/Runtime/Entities/Tasks/AbstractTaskDriverSystem.cs
jkeon: > Developer should be forced to explicitly define what the cancel behaviour is. Even if annoying it's better than forgetting to think through a cancel situation.

Ok sounds good.

> Back to my other question though, is there an actual optimization path to OneShot? If not, then let's just have every task required to write a result. Just like in the application code there's always a return code. I don't see the purpose to a task that you fire and forget but then for some reason want to be able to cancel.

There was an optimization but I think our other comments cover how the changes will negate or improve this.

Scripts/Runtime/Entities/Tasks/Data/JobTaskWorkConfig.cs
jkeon: It can be, but then you are polluted with a bunch of functions that won't do anything/work

Scripts/Runtime/Data/VirtualData/InstanceData/IKeyedData.cs
mbaker3: Ah I see,
Discussed on phone about maybe just composing the VDContextID on demand internally.

Scripts/Runtime/Data/VirtualData/CancelData.cs
```diff
@@ -0,0 +1,151 @@
...
+using Anvil.Unity.DOTS.Jobs;
+using System;
+using Unity.Burst;
+using Unity.Collections;
+using Unity.Collections.LowLevel.Unsafe;
+using Unity.Jobs;
+
+namespace Anvil.Unity.DOTS.Data
+{
+    //TODO: DISCUSS - Further Abstraction or just like so?
```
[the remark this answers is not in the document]
mbaker3: What is in `VirtualDataBase` vs `AbstractVirtualData`?

I'm assuming `VirtualData` isn't actually a type. That's an example of a specific subclass that a developer would implement?

Scripts/Runtime/Entities/Tasks/Data/JobTaskWorkConfig.cs
mbaker3: ah ok, that seems fine I think

Scripts/Runtime/Data/VirtualData/InstanceData/IKeyedData.cs
jkeon: I reworked this so that you as the developer deal only with your instances of `IKeyedData` which operate on `Entity` as the key. 

It's readonly and can't be changed so that's good.

Internally, `VirtualData` handles wrapping your instance in a `VDWrapper` struct with assigns the proper context for you. You never see it and it's internal so you can't. 

I profiled and the difference was negligible.

Scripts/Runtime/Data/VirtualData/VirtualData.cs
jkeon: I reworked this so that we can read from cancelled data and write to the live data of the same type in the same job. No need to have this random `RestartData` struct to ping off of anymore.

Scripts/Runtime/Data/VirtualData/VirtualData.cs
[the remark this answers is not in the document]
jkeon: This was reworked to be more efficient. We explored a bunch of different options but ran into safety/usability issues on the fastest approach and the others were too slow. 

We've gone from about 10 to 13ms for 100,000 elements down to 3ms so that's pretty decent.

jkeon:
Closing this out as #77 is the culmination of this effort.

PR #47, opened 2022-08-02
Component Referencable - Managed instances in ECS Component Data

mbaker3:
This PR adds the ability to store a managed reference in an `IComponentData` container. Effectively allowing you to attach managed instances to entities.

There is no convenient way to reference a managed instance in `IComponentData`. This is particularly painful when trying to interface between traditional OO code and ECS.

Existing approaches are:
 - Store managed references on a system (Ex: `AbstractDataSystem`) and have all systems query the system.
    - Can't easily associate instance with a particular entity
    - Instance doesn't travel with entities between worlds
    - Can't be referenced in jobs or burst compiled code.

 - Use Managed IComponentData
    - Not performant
    - Can't be referenced in jobs or burst compiled code.

 - Create a managed wrapper around an unsafe type
    - Pain to create and manage
    - Not useful for 3rd party instances that aren't written in this way.

The Component Referencable feature allows developers to easily attach managed instance references to entities. The main pieces are:
 - `IComponentReferencable`/`AbstractComponentReferencable - Types that implement or subclass can be referenced by hashcode using a `ManagedReference<T>` instance 
 - `ManagedReferenceStore` - Maintains a map between hashcode and managed instance
 - `ManagedReference<T>` - Holds the hashcode of a managed instance it represents. This instance is `IComponentData` compatible and can be resolved into the managed instance it represents on the main thread.

##### Example usage
The most common current usage is to map a world scoped singleton to a managed instance.

**A managed instance:**
```csharp
public class MyWorldSingleton : AbstractComponentReferencable
{
   //.... any managed fields, properties, etc...
   // Just a normal IDisposable class
}
```

**Adding the instance**
```csharp
   // In some managed context
   EntityManager entityManager = myWorld.EntityManager;
   Entity entity = entityManager.CreateEntity();
   entityManager.AddManagedRefTo(entity, new MyWorldSingleton())
```

**A consuming system:**
```csharp
public partial class MySystem : SystemBase
{
   protected override void OnCreate()
   {
      base.OnCreate();

      this.RequireManagedSingletonForUpdate<MyWorldSingleton>();
   }

   {
      MyWorldSingleton myWorldSingleton = this.GetManagedSingleton<MyWorldSingleton>();

      // Or if it's not a singleton...
      ManagedReference<MyWorldSingleton> myWorldSingletonRef = EntityManager.GetComponentData<ManagedReference<MyWorldSingleton>>(someEntity);
      MyWorldSingleton myWorldSingleton = myWorldSingletonRef.Resolve();
   }
}
```

However this feature could be used to:
 - Maintain a link between an OO version of a game actor and the ECS version.
 - Provide access to managed configuration data in a world.
 - Easily implement 

##### Benefits
 - `ManagedReference<T>` can be compared against other `ManagedReference<T>` in threaded and burst compiled code
    - Managed instance access can only done from main thread in non-burst code.
 - Can be associated with an Entity like any other `IComponentData`

 - None

 - https://github.com/decline-cookies/anvil-csharp-core/pull/97

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Entities/ManagedReference/ManagedReference.cs
+++ b/Scripts/Runtime/Entities/ManagedReference/ManagedReference.cs
@@ -1,4 +1,5 @@
 using Unity.Entities;
+using UnityEngine;
 
 namespace Anvil.Unity.DOTS.Entities
 {
@@ -9,7 +10,7 @@ namespace Anvil.Unity.DOTS.Entities
     /// <typeparam name="T">The type of the managed instance that is referenced.</typeparam>
     public readonly struct ManagedReference<T> : IComponentData where T : class, IComponentReferencable
     {
-        private readonly int m_ManagedContextHash;
+        private readonly uint m_ManagedReferenceID;
 
         /// <summary>
         /// Creates a new managed reference from an <see cref="IComponentReferencable"/> instance.
@@ -17,7 +18,11 @@ namespace Anvil.Unity.DOTS.Entities
         /// <param name="instance">The instance to hold reference too.</param>
         public ManagedReference(T instance)
         {
-            m_ManagedContextHash = ManagedReferenceStore.GetHash(instance);
+            m_ManagedReferenceID = ManagedReferenceStore.GetID(instance);
+
+            // Shouldn't ever happen but we want to know since there are equality checks that depend on the default
+            // instance being unique to any properly initialized instance.
+            Debug.Assert(!Equals(default(ManagedReference<T>)));
         }
 
         /// <summary>
@@ -26,7 +31,7 @@ namespace Anvil.Unity.DOTS.Entities
         /// <returns>The managed instance that this structure represents.</returns>
         public T Resolve()
         {
-            return ManagedReferenceStore.Get<T>(m_ManagedContextHash);
+            return ManagedReferenceStore.Get<T>(m_ManagedReferenceID);
         }
     }
 }
--- a/Scripts/Runtime/Entities/ManagedReference/ManagedReference.cs
+++ b/Scripts/Runtime/Entities/ManagedReference/ManagedReference.cs
@@ -2,15 +2,28 @@ using Unity.Entities;
 
 namespace Anvil.Unity.DOTS.Data
 {
+    /// <summary>
+    /// Holds reference to a managed instance that is registered with <see cref="ManagedReferenceStore"/> in a valid
+    /// <see cref="IComponentData"/> structure.
+    /// </summary>
+    /// <typeparam name="T">The type of the managed instance that is referenced.</typeparam>
     public readonly struct ManagedReference<T> : IComponentData where T : class, IComponentReferencable
     {
         private readonly int ManagedContextHash;
 
+        /// <summary>
+        /// Creates a new managed reference from an <see cref="IComponentReferencable"/> instance.
+        /// </summary>
+        /// <param name="instance">The instance to hold reference too.</param>
         public ManagedReference(T instance)
         {
             ManagedContextHash = ManagedReferenceStore.GetHash(instance);
         }
 
+        /// <summary>
+        /// Retrieve the managed instance from the <see cref="ManagedReference{T}"/>.
+        /// </summary>
+        /// <returns>The managed instance that this structure represents.</returns>
         public T Resolve()
         {
             return ManagedReferenceStore.Get<T>(ManagedContextHash);
--- a/Scripts/Runtime/Entities/ManagedReference/ManagedReference.cs
+++ b/Scripts/Runtime/Entities/ManagedReference/ManagedReference.cs
@@ -9,7 +9,7 @@ namespace Anvil.Unity.DOTS.Data
     /// <typeparam name="T">The type of the managed instance that is referenced.</typeparam>
     public readonly struct ManagedReference<T> : IComponentData where T : class, IComponentReferencable
     {
-        private readonly int ManagedContextHash;
+        private readonly int m_ManagedContextHash;
 
         /// <summary>
         /// Creates a new managed reference from an <see cref="IComponentReferencable"/> instance.
@@ -17,7 +17,7 @@ namespace Anvil.Unity.DOTS.Data
         /// <param name="instance">The instance to hold reference too.</param>
         public ManagedReference(T instance)
         {
-            ManagedContextHash = ManagedReferenceStore.GetHash(instance);
+            m_ManagedContextHash = ManagedReferenceStore.GetHash(instance);
         }
 
         /// <summary>
@@ -26,7 +26,7 @@ namespace Anvil.Unity.DOTS.Data
         /// <returns>The managed instance that this structure represents.</returns>
         public T Resolve()
         {
-            return ManagedReferenceStore.Get<T>(ManagedContextHash);
+            return ManagedReferenceStore.Get<T>(m_ManagedContextHash);
         }
     }
 }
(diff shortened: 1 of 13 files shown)


Scripts/Runtime/Entities/ManagedReference/ManagedReferenceExtension.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// </param>
+        /// <typeparam name="T">The type of the managed instance.</typeparam>
+        public static void RemoveManagedRefFrom<T>(this EntityManager entityManager, Entity entity, T instance) where T : class, IComponentReferencable
+        {
+            DEBUG_AssertInstanceMatchesRef(entityManager, entity, instance);
+            entityManager.RemoveComponent<ManagedReference<T>>(entity);
+        }
+
+        // Ensure that the instance that's been provided matches the one defined in the component ref on the entity
+        [Conditional("DEBUG")]
```
jkeon, line 57: DEBUG? Or Unity's Collections checks one? We seem to be using the Collections Checks one for any DOTS code.

Scripts/Runtime/Entities/ManagedReference/ManagedReferenceStore.cs
```diff
@@ -0,0 +1,72 @@
...
+
+namespace Anvil.Unity.DOTS.Data
+{
+    /// <summary>
+    /// A static store of all <see cref="IComponentReferencable"/> instances. Used to get instances by hash.
+    /// Used by <see cref="ManagedReference{T}"/>.
+    /// </summary>
+    public static class ManagedReferenceStore
+    {
+        private static readonly Dictionary<int, IComponentReferencable> m_instances = new Dictionary<int, IComponentReferencable>();
```
jkeon, line 12:
```suggestion
        private static readonly Dictionary<int, IComponentReferencable> m_Instances = new Dictionary<int, IComponentReferencable>();
```

Scripts/Runtime/Entities/ManagedReference/ManagedReferenceStore.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        /// <summary>
+        /// Registers a managed instance so it can be resolved by a <see cref="ManagedReference{T}"/>.
+        /// This is usually called when the instance is created.
+        /// </summary>
+        /// <param name="instance">The instance to register.</param>
+        public static void RegisterReference(IComponentReferencable instance)
+        {
+            Debug.Assert(instance.GetType().IsClass, $"Implementations of {nameof(IComponentReferencable)} must be classes");
+            m_instances.Add(instance.GetHashCode(), instance);
```
jkeon, line 30: While it is unlikely, it's possible that two separate instances could generate the same hash code and we'll get a collision. 

This will throw an exception here which is probably fine for alerting us that we now need to deal with it. But just wanted to raise it in case we want to do some sort of assert check before. 

Unity's TypeManager goes WAY into depth with detecting and dealing with Hash Collisions which I don't think we need to do, but if we discover that we're getting hash collisions we might need to handle modifying the hash code.

Scripts/Runtime/Entities/ManagedReference/ManagedReferenceExtension.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// <param name="preventOverwrite">
+        /// (Default: true) When true the method will throw an exception if a <see cref="ManagedReference{T}"/> of the
+        /// same type already exists on the entity.
+        /// </param>
+        /// <typeparam name="T">The type of the managed instance.</typeparam>
+        /// <exception cref="InvalidOperationException">
+        /// Thrown when <see cref="preventOverwrite"/> is true and a <see cref="ManagedReference{T}"/> of the same type
+        /// already exists on the <see cref="Entity"/>.
+        /// </exception>
+        public static void AddManagedRefTo<T>(this EntityManager entityManager, Entity entity, T instance, bool preventOverwrite = true) where T : class, IComponentReferencable
```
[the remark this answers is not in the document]
mbaker3: My singleton use case means that paying this one time performance cost isn't such a big deal. If the API exists on the Unity side (`EntityManager.AddComponentData`) I don't think there's harm in there being a managed equivalent.

This did make me look closer at what `SetComponentData` and `AddComponentData` will actually accept and the `preventOverwrite` param doesn't make any sense for add.

Currently writing a SetManagedRefOn method.

Scripts/Runtime/Entities/ManagedReference/ManagedReferenceExtension.cs
mbaker3: This seems like a good, reasonably low cost check for an infrequent call that is good for debug. You won't want to miss one of these!

Scripts/Runtime/Entities/ManagedReference/ManagedReferenceStore.cs
mbaker3: Yea, I just looked into this for a different reason. Checking if GetHashCode() would ever return 0 to see if I could easily detect a `default(ManagedRef<T>)` and found this:
https://docs.microsoft.com/en-us/dotnet/api/system.object.gethashcode?view=net-6.0#remarks

Soooo, I need to rethink hashcodes. I think if I mix in an ID provider I'll be fine but this is why I had a separate GetHash method to make this change easy!

Scripts/Runtime/Entities/ManagedReference/ManagedReferenceStore.cs
mbaker3: Fixed

Scripts/Runtime/Entities/ManagedReference/ManagedReference.cs
```diff
@@ -0,0 +1,32 @@
...
+namespace Anvil.Unity.DOTS.Data
+{
+    /// <summary>
+    /// Holds reference to a managed instance that is registered with <see cref="ManagedReferenceStore"/> in a valid
+    /// <see cref="IComponentData"/> structure.
+    /// </summary>
+    /// <typeparam name="T">The type of the managed instance that is referenced.</typeparam>
+    public readonly struct ManagedReference<T> : IComponentData where T : class, IComponentReferencable
+    {
+        private readonly int ManagedContextHash;
```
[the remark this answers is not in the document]
mbaker3: Fixed

Scripts/Runtime/Entities/ManagedReference/ManagedReferenceExtension.cs
jkeon: Ok sounds good.

Scripts/Runtime/Entities/ManagedReference/ManagedReferenceExtension.cs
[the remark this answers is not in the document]
jkeon: Sounds good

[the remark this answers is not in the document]
mbaker3: Implemented

Scripts/Runtime/Entities/ManagedReference/ManagedReferenceStore.cs
mbaker3: Reworked. See next PR post for details.

mbaker3:
Ready for review again.

This PR now depends on https://github.com/decline-cookies/anvil-csharp-core/pull/97
(Updated description too)

Also fixed the namespaces to be consistent with the rest of Anvil.

jkeon (approved):

PR #51, opened 2022-08-22
Fix Build Support

mbaker3:
These changes allow projects using Anvil-DOTS to be exported to a build.

Attempting to publish a build when Anvil-DOTS is referenced throws compiler errors.
 - Missing `#if ENABLE_UNITY_COLLECTIONS_CHECKS` guards around `DisposeSentinel` and `AtomicSafetyHandle` uses
 - Issues with the compiler correctly handling methods decorated with `[Conditional]` attributes referencing `#if` wrapped code (when they both reference the same `#define`).
 - Static logic that relies on `[RuntimeInitializeOnLoad]` code being executed fails.

 - Add `#if ENABLE_UNITY_COLLECTIONS_CHECKS` guards around `SafetyHandlePtr`, `DisposeSentinel`, and `AtomicSafetyHandle` uses
 - Methods decorated with `[Conditional]` attributes have their logic referencing `#if` wrapped code now also have their logic `#if` wrapped.
    - This means that when this limitation is resolved by Unity we will automatically get the benefits of having the method removed. We can periodically go back and check if this limitation still exists and remove the new `#if` checks if it has been fixed.
 - Static logic that relies on `[RuntimeInitializeOnLoad]` now also has a static constructor that calls the same decorated `Init()` method.

Ability to export a build that correctly functions.

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Data/Collections/DeferredNativeArray.cs
+++ b/Scripts/Runtime/Data/Collections/DeferredNativeArray.cs
@@ -19,21 +19,31 @@ namespace Anvil.Unity.DOTS.Data
     /// </remarks>
     public unsafe struct DeferredNativeArrayScheduleInfo
     {
+#if ENABLE_UNITY_COLLECTIONS_CHECKS
         internal void* SafetyHandlePtr
         {
             get;
         }
+#endif
 
         internal void* BufferPtr
         {
             get;
         }
 
+#if ENABLE_UNITY_COLLECTIONS_CHECKS
         internal DeferredNativeArrayScheduleInfo(void* safetyHandlePtr, void* bufferPtr)
         {
             SafetyHandlePtr = safetyHandlePtr;
             BufferPtr = bufferPtr;
         }
+#else
+        internal DeferredNativeArrayScheduleInfo(void* bufferPtr)
+        {
+            BufferPtr = bufferPtr;
+        }
+#endif
+
     }
 
     /// <summary>
@@ -88,6 +98,7 @@ namespace Anvil.Unity.DOTS.Data
         private static readonly int SIZE = UnsafeUtility.SizeOf<T>();
         private static readonly int ALIGNMENT = UnsafeUtility.AlignOf<T>();
 
+#if ENABLE_UNITY_COLLECTIONS_CHECKS
         // ReSharper disable once StaticMemberInGenericType
         private static int s_StaticSafetyId;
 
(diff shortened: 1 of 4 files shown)


Scripts/Runtime/Data/Collections/DeferredNativeArray.cs
```diff
@@ -19,21 +19,31 @@ namespace Anvil.Unity.DOTS.Data
...
             get;
         }
+#endif
 
         internal void* BufferPtr
         {
             get;
         }
 
+#if ENABLE_UNITY_COLLECTIONS_CHECKS
```
mbaker3, line 34: This is a bit awkward. 
Alternatively I could have duplicated the type in the different `#if` conditions but that didn't seem any better.

Ideally it would be best to not include the safety handle pointer in the struct and instead request it from the `DeferredNativeArray` instance but I don't know if that's feasible.

jkeon:
I've read through this but I need some time to go through it in depth. Might need to chat with you about it.

Scripts/Runtime/Entities/ManagedReference/ManagedReferenceStore.cs
```diff
@@ -12,20 +12,27 @@ namespace Anvil.Unity.DOTS.Entities
...
+        {
+            s_Logger = Log.GetStaticLogger(typeof(ManagedReferenceStore));
+            s_IDToInstance = new Dictionary<uint, IComponentReferencable>();
+            s_InstanceToID = new Dictionary<IComponentReferencable, uint>();
+
+            Init();
+        }
+
         [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.SubsystemRegistration)]
         private static void Init()
```
jkeon, line 30: Does this not get called twice in Editor? 

It may not matter for this function in particular but I've run into this problem with this pattern. You may want to wrap in a #if check for UNITY_EDITOR or it may just be fine to let it be called twice.

jkeon (approved):
Ok read through in depth and I understand now. 

Annoying but looks fine. One small question.

Scripts/Runtime/Entities/ManagedReference/ManagedReferenceStore.cs
mbaker3: You're right. That's risky to have IDs reset if something was triggering the static constructor before subsystem registration happened. (which was happening!)

I've removed the `Init` call from the constructor and changed to `RuntimeInitializeLoadType.AfterAssembliesLoaded` which works in builds and editor.

mbaker3:
Ready for re-review

jkeon (approved):

PR #53, opened 2022-08-31
Fixing job scheduling

jkeon:
Fixed Job scheduling code to reflect the right way to handle generics 

When using some data that was generic (Wrapper<T>) the jobs couldn't be scheduled due to some internal error with `SharedStatic` and the generic. 

I looked at the internal code for Unity's Job scheduling and it's a mess with different ways to do things across the board from `IJob` to `IJobFor` etc. 

I took the parts that made sense from all of them that fixed the issue so the compiler knows what data we're trying to operate on.

Nothing changes for the developer and no changes in scheduling/performance.

- None

 - #52 

 - [ ] Yes 
 - [x] No

### Tag Alongs
 - `ParallelAccessUtil` was in a file named `ParallelCollectionUtil`. Fixed that.
 - Also encountered a rare bug where Unity put my jobs on the main thread and the profiler thread in the same session. So I expanded the range to account for it.


Scripts/Runtime/Jobs/JobTypes/AnvilJobBatchDeferExtension.cs
```diff
@@ -27,7 +25,8 @@ public static class AnvilJobBatchDeferExtension
             atomicSafetyHandlePtr = scheduleInfo.SafetyHandlePtr;
 #endif
 
-            IntPtr reflectionData = WrapperJobProducer<TJob>.GetReflectionData();
+            IntPtr reflectionData = WrapperJobProducer<TJob>.JOB_REFLECTION_DATA;
+            CheckReflectionDataCorrect(reflectionData);
```
mbaker3, line 29: 🚲 Suggest rename to `ValidateReflectionData()`

Scripts/Runtime/Jobs/Util/ParallelAccessUtil.cs
```diff
@@ -95,10 +97,10 @@ public static int CollectionSizeForMaxThreads
         /// Our goal is to have a tightly packed collection so in the example of an 8 core machine with
         /// <see cref="JobsUtility.JobWorkerMaximumCount"/> equal to 15, we would have the following mapping.
         ///
-        /// thread 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, X
-        /// index  0, 1, 2, 3, 4, 5, 6, 7, 8, 9,  10, 11, 12, 13, 14, 15
+        /// thread 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, X, X2
+        /// index  0, 1, 2, 3, 4, 5, 6, 7, 8, 9,  10, 11, 12, 13, 14, 15, 16
         ///
-        /// We have a tightly packed collection with index 0 through 15 for a total of 16 buckets.
+        /// We have a tightly packed collection with index 0 through 16 for a total of 17 buckets. (15 job workers, one main, one profiler)
```
mbaker3, line 103: 🚲 I think this line is past our margin?

```suggestion
        /// We have a tightly packed collection with index 0 through 16 for a total of 17 buckets. 
        /// (15 job workers, one main, one profiler)
```

mbaker3 (approved):
Minor 🚲 comments

Scripts/Runtime/Jobs/JobTypes/AnvilJobBatchDeferExtension.cs
jkeon: Well that's what I get for copying Unity. Much better name, fixed.

mbaker3: Is that seriously what they named theirs?! 🤦

jkeon: It sure is... it sadly sure is. Surprisingly one of the few things that's actually consistent between all job types too!

PR #55, opened 2022-09-01
Adding enumerator and better clear functions

### LINKED ISSUE #16: UnsafeTypedStream Improvements
- Allow for specifying number of lanes.
- Create enumerators for nicer iteration - https://github.com/decline-cookies/anvil-unity-dots/pull/14/files#r842968034

jkeon:
Improvements to `UnsafeTypedStream`

**Iterating**
In order to iterate through the collection, you have to build a doubly nested for loop. 

```
for (all lanes in the stream)
{
    for (all elements in the current lane)
    {
          //do whatever
    }
}
```

The pseudocode is nicer than the actual implementation. It's annoying and hard to remember. 

**Clearing**
When clearing the stream, it disposes all the memory. This is wasteful for times where you are going to have MANY elements that take up many blocks and you have to reallocate all those blocks again.

**Iterating**
There is now an `Enumerator` tailored to the context of Burst and Native Collections. You can simply do a `foreach` now.

**Clearing**
There is a `Clear` function which will reset the headers so that the collection thinks it is empty but the memory is still there. They can be overwritten and filled up again without the need to re-allocate memory.

The `ClearAndReduceCapacity` function will clear but also free up the allocated memory. 

- Resolves #16

 - #54 

 - [x] Yes but no one is using it right now so it's fine.
 - [ ] No

### Tag-Alongs
- Fixed a burst error with Enum casting.

### DIFF
--- a/Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
+++ b/Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
@@ -1,5 +1,8 @@
 using Anvil.Unity.DOTS.Entities;
 using Anvil.Unity.DOTS.Jobs;
+using System;
+using System.Collections;
+using System.Collections.Generic;
 using System.Runtime.InteropServices;
 using Unity.Assertions;
 using Unity.Burst;
@@ -196,8 +199,7 @@ namespace Anvil.Unity.DOTS.Data
             m_BufferInfo->LaneCount = laneCount;
 
             m_BufferInfo->LaneInfos = (LaneInfo*)UnsafeUtility.Malloc(LaneInfo.SIZE * m_BufferInfo->LaneCount, LaneInfo.SIZE, allocator);
-
-            //TODO: #16 - Look into creating an enumerator like NativeArray to allow for foreach - https://github.com/decline-cookies/anvil-unity-dots/pull/14/files#r842968034
+            
             for (int i = 0; i < m_BufferInfo->LaneCount; ++i)
             {
                 LaneInfo* lane = m_BufferInfo->LaneInfos + i;
@@ -220,7 +222,7 @@ namespace Anvil.Unity.DOTS.Data
                 return;
             }
 
-            Clear();
+            ClearAndReduceCapacity();
 
             Allocator allocator = m_BufferInfo->Allocator;
             UnsafeUtility.Free(m_BufferInfo->LaneInfos, allocator);
@@ -234,7 +236,7 @@ namespace Anvil.Unity.DOTS.Data
         /// Clears all data in the collection
         /// </summary>
         [WriteAccessRequired]
-        public void Clear()
+        public void ClearAndReduceCapacity()
         {
             if (!IsCreated)
             {
@@ -264,6 +266,42 @@ namespace Anvil.Unity.DOTS.Data
                 lane->Count = 0;
             }
         }
+        
+        [WriteAccessRequired]
+        public void Clear()
+        {
+            if (!IsCreated)
+            {
+                return;
+            }
+
+            for (int i = 0; i < m_BufferInfo->LaneCount; ++i)
+            {
+                LaneInfo* lane = m_BufferInfo->LaneInfos + i;
+                //Reset current writer block to the first block and reset writer heads and counts
+                lane->CurrentWriterBlock = lane->FirstBlock;
+                if (lane->CurrentWriterBlock == null)
+                {
+                    lane->WriterHead = null;
+                    lane->WriterEndOfBlock = null;
+                }
+                else
+                {
+                    lane->WriterHead = lane->CurrentWriterBlock->Data;
+                    lane->WriterEndOfBlock = lane->WriterHead + m_BufferInfo->BlockSize;
+                }
+                lane->Count = 0;
+            }
+        }
+        
+        /// <summary>
+        /// Gets an enumerator for iterating through all elements in the stream across all lanes
+        /// </summary>
+        /// <returns>The <see cref="Enumerator"/></returns>
+        public Enumerator GetEnumerator()
+        {
+            return new Enumerator(m_BufferInfo);
+        }
 
         /// <summary>
         /// Schedules the disposal of the collection.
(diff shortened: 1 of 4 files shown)


Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
```diff
@@ -234,7 +236,7 @@ public void Dispose()
         /// Clears all data in the collection
         /// </summary>
         [WriteAccessRequired]
-        public void Clear()
+        public void ClearAndReduceCapacity()
```
mbaker3, line 239: `List<T>` has `TrimExcess()`

Should we follow that lead and have excess trimming its own method?
So you'd call `Clear()` then `TrimExcess()` if you wanted to collapse the capacity down.

Or are there optimizations in assuming that you're doing a clear at the same time as the trim?

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
```diff
@@ -765,6 +825,77 @@ internal void ReadBlock(byte* dstStart, int dstStartIndex, int elementsToRead)
...
+        public struct Enumerator : IEnumerator<T>
+        {
+            [NativeDisableUnsafePtrRestriction] private readonly BufferInfo* m_BufferInfo;
+            private int m_LaneIndex;
+            private LaneReader m_LaneReader;
+            private int m_ArrayIndex;
+            
+            public T Current
+            {
+                get => m_LaneReader.Read();
```
mbaker3, line 843: Might want to cache the current value locally like they do in `List<T>`
https://referencesource.microsoft.com/#mscorlib/system/collections/generic/list.cs,cf7f4095e4de7646

This way you don't get an error if you read `Current` before calling `MoveNext` for the first time. It's at least consistent with .NET. Also seems to handle calling `Current` after you reach the end of the enumerable.

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
```diff
@@ -765,6 +825,77 @@ internal void ReadBlock(byte* dstStart, int dstStartIndex, int elementsToRead)
...
+            {
+                m_LaneIndex = 0;
+                m_ArrayIndex = -1;
+                m_LaneReader = new LaneReader(m_BufferInfo, m_LaneIndex);
+            }
+
+            object IEnumerator.Current
+            {
+                get => throw new NotSupportedException($"Calling generic accessor which will box!");
+            }
```
mbaker3, line 892: I wonder if you could burst discard this and only allow it when burst is disabled?
Not that it's a good idea but just for completion. Maybe for debug purposes you're dealing with these enumerators in a generic way 🤷

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
jkeon: It took me a while to figure out how to do this without adding a bunch of extra data to monitor state but I got it. 

This is a good idea. Pasting here because you won't actually see it until the next PR. 

```
        /// <summary>
        /// Frees memory for blocks of elements that are not being used.
        /// </summary>
        /// <remarks>
        /// Because this is a block based collection, trimming will not be a tight fit.
        /// Ex. 10 blocks allocated, 4 elements per block, 40 elements capacity total.
        /// A <see cref="Clear()"/> is called and then 5 elements are written and then <see cref="TrimExcess"/> is
        /// called.
        /// The first block of 4 elements will stay. The 5th element will keep the second block around. The remaining
        /// 8 blocks will be freed.
        /// In the end, 2 blocks will remain, 5 elements in the collection, 8 elements capacity.
        /// </remarks>
        [WriteAccessRequired]
        public void TrimExcess()
        {
            if (!IsCreated)
            {
                return;
            }

            Allocator blockAllocator = m_BufferInfo->BlockAllocator;

            for (int i = 0; i < m_BufferInfo->LaneCount; ++i)
            {
                LaneInfo* lane = m_BufferInfo->LaneInfos + i;
                
                //If our current writing block is null, we can just skip, there's nothing to free
                if (lane->CurrentWriterBlock == null)
                {
                    continue;
                }
                
                //If we have no elements at all in the lane (Clear was called), then we want to start with clearing
                //the current writer block. If there is anything written, (We're calling this to free up excess memory),
                //then the current writer block is still valid and we only want to free blocks after.
                BlockInfo* block = (lane->Count == 0) ? lane->CurrentWriterBlock : lane->CurrentWriterBlock->Next;
                while (block != null)
                {
                    BlockInfo* currentBlock = block;
                    block = currentBlock->Next;

                    UnsafeUtility.Free(currentBlock->Data, blockAllocator);
                    UnsafeUtility.Free(currentBlock, blockAllocator);
                }
            }
        }
```

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
jkeon: Interesting. Everything I remember reading was that you were really not supposed to call `Current` before `MoveNext` or at the end. `Enumerators` worked on a very specific set of rules that had to be followed and that `foreach` code gen expected. 

That said, I couldn't find those readings anywhere and it does look like .NET ensures you don't error for all its enumerators so I agree, it's best to be consistent. 

I've made the change. 

There is a risk (albeit a small one) that someone comes along and does a `GetEnumerator` manually. Doesn't know the "rules" and calls `Current` first and then `MoveNext` in their loop. They will get back `default` which due to the struct nature of everything here, may look completely normal so they don't notice. They will get one extra iteration in their loop they weren't expecting but unless they are trying to add to a presized array, they won't notice that either. 

Before when it would throw an error, at least they would immediately look into it and learn how `Enumerators` work and ideally just use a `foreach` but still, they'd get the feedback right away and it would be easier to debug. 

I think the risk here is less than being consistent with .NET so it's fine, but just bringing it up for posterity.

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
jkeon: Hmmmn, yeah I think that's a good idea. Changed.

Scripts/Runtime/Data/Collections/UnsafeTypedStream.cs
mbaker3: Agreed on the risk.
We could add a debug only log message warning the developer 🤷

jkeon: Yeah I think it's fine. I haven't seen anyone use the actual `GetEnumerator` call for a long long long time so I feel like that would set off some flags in the PR where we ask "Why no `foreach`"?

PR #58, opened 2022-09-12
Task System Attributes

jkeon:
Some more renaming to provide better clarity of what everything does. Let me know what you think about this.
Added some automatic generation of the backing data and safety checks, plus some optimizations.

Previously we had to manually create data through a special function so that it would be registered with the system properly. We also just hoped that you would also create a job that would work on that data so we wouldn't have data loss.

There was confusion with the naming mostly stemming from the word `Data` and wondering when it meant the plural form or the singular form. That propagating down to `Tasks` and the `Jobs` and it became weird quick.

## Naming

`ProxyDataStream` is a collection of instances that implement `IProxyInstance`. 

When referring to the singular, it's an "instance", "TInstance", "TimerInstance", "TimerCompletedInstance" etc

Each instance is wrapped in an `ProxyInstanceWrapper` which wraps the payload of an `IProxyInstance` with a `ProxyInstanceID` to be able to give context to the data as it flows through the application.

Everything that operates on the `Stream` is in the plural sense. `DataStreamReader` allows you to read from a `ProxyDataStream` etc. 

At the core are `TaskSystem`s and `TaskDrivers`. 
- The `TaskSystem` operates on generic data to update massively in parallel. You as the end developer won't interact directly with a `TaskSystem`.
- The `TaskDriver` gives context to data that will flow into a `TaskSystem` and flow back out to the `TaskDriver`. They can be stitched together and provide the logic for Instances to flow around the application touching different systems as appropriate.

Within the `TaskSystem` (and in the future in `TaskDrivers`) are `TaskProcessors`. 

These are objects that take in a `ProxyDataStream` and run it through a `Job` to process the data and update it or convert it to some form of result. 

## Changes

- `AbstractTaskSystem` will now use reflection to get any `ProxyDataStream` field and properly create it and keep track if a job gets associated. This is done automatically in the constructor so all `ProxyDataStream` fields are valid right away and can be linked together for jobs outside of the class.
- Since we know the number of data after `OnCreate` we can optimize bulk scheduling to use a pre-allocated array and remove a temp allocation per frame.
- `ProxyDataStream` fields can be ignored with the `IgnoreProxyDataStreamAttribute`
- Lots of safety checks.

- None

 - None

 - [ ] Yes
 - [x] No


Scripts/Runtime/Entities/ProxyDataStream/Data/ProxyInstanceWrapper.cs
```diff
@@ -0,0 +1,75 @@
+using System;
```
jkeon, line 1: I don't know why it caught that `ProxyDataID` was renamed to `ProxyInstanceID` but this shows up as a new class. 

Only change is the name

Scripts/Runtime/Entities/Tasks/AbstractTaskSystem.cs
```diff
@@ -9,26 +15,33 @@ namespace Anvil.Unity.DOTS.Entities
     public abstract partial class AbstractTaskSystem : AbstractAnvilSystemBase,
                                                        ITaskSystem
     {
+        private static readonly Type I_PROXY_DATA_STREAM_TYPE = typeof(IProxyDataStream);
```
mbaker3, line 18: I'm not sure it's beneficial to do this. I think this is just a cached retrieval of the Type instance anyway.
It would be cleaner to just do this at the start of `CreateProxyDataStreams()`

Scripts/Runtime/Entities/Tasks/AbstractTaskSystem.cs
```diff
@@ -117,5 +193,55 @@ private JobHandle UpdateTaskDriverSystem(JobHandle dependsOn)
...
+
+        [Conditional("ENABLE_UNITY_COLLECTIONS_CHECKS")]
+        private void Debug_CheckDataStreamUpdateJobsExist()
+        {
+            if (m_HasCheckedDataStreamUpdateJobsIntegrity)
+            {
+                return;
+            }
+
+            m_HasCheckedDataStreamUpdateJobsIntegrity = true;
```
mbaker3, line 223: This will fail to compile when `ENABLE_UNITY_COLLECTIONS_CHECKS` isn't defined. Need to wrap in an `#if`

mbaker3 (changes requested):
Nice work! Naming is half the battle.

One compilation error and then some questions and suggestions.

Scripts/Runtime/Entities/ProxyDataStream/ProxyDataStreamFactory.cs
```diff
@@ -0,0 +1,59 @@
...
+        private static readonly MethodInfo PROTOTYPE_METHOD = typeof(ProxyDataStreamFactory).GetMethod(nameof(CreateProxyDataStream), BindingFlags.Static | BindingFlags.NonPublic);
+        private static readonly Dictionary<Type, MethodInfo> TYPED_GENERIC_METHODS = new Dictionary<Type, MethodInfo>();
+
+        [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.SubsystemRegistration)]
+        private static void Init()
+        {
+            TYPED_GENERIC_METHODS.Clear();
+        }
+
+        public static IProxyDataStream Create(Type instanceType)
```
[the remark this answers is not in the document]
jkeon: I create these data streams via reflection so I don't know what the type is at compile time.

```
//Don't worry the safety checks are above this so the below line is valid :)
ProxyDataStreamFactory.Create(systemField.FieldType.GenericTypeArguments[0]);
```

Scripts/Runtime/Entities/Tasks/AbstractTaskSystem.cs
jkeon: Good catch, this is now changed as the bool is useful for other aspects of the state. It's no longer wrapped in `#if`

Scripts/Runtime/Entities/ProxyDataStream/ProxyDataStreamFactory.cs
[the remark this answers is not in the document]
mbaker3: ah ok, makes sense why you're jumping through all these hoops.

Scripts/Runtime/Entities/Tasks/AbstractTaskSystem.cs
mbaker3: huh neat, When I was looking it up `typeof` is frequently labelled as "compile-time" resolved which isn't really accurate but MSDN does repeatedly say it is an efficient implementation.

That was a good post you linked. I agree with the guy saying it is faster but you shouldn't use it unless you're solving a provable bottleneck or need that type repeatedly across many scopes.

PR #59, opened 2022-09-20
Re-Enable TaskDrivers

jkeon:
Good spot for the next PR, this re-enables TaskDrivers and introduces the concept of the `TaskFlowGraph`.

In addition, lots of streamlining, deleted code and safety checks.

TaskDrivers were temporarily disabled during the big refactor to support cancelling.

- `ProxyDataStream` loses the interface and gains an `AbstractProxyDataStream`
    - This lets us get rid of a layer of indirection in the call stack for bulk scheduling and gives us a nice debug string 
- Did a pass through all the code to make things that should be internal, internal. You don't need to care about a lot of stuff as the consuming dev and your intellisense will thank you.
- Deleted a bunch of old files that were there for reference as well as some that are just no longer needed.
- Reworked the `DataStreamUpdater`, `DataStreamWriter` and `DataStreamReader` to be easier to read/follow and have safety functions nicely grouped.
- Re-Enabled the `AbstractTaskDriver` 
    - `TaskDrivers` represent contextual groupings of data that need to operate through a general set of jobs on a `TaskSystem`. They'll need to keep that context when running through the job and `TaskDrivers` let you do that so you can stitch together many different types of `TaskDrivers` to create more complex patterns.
-  Reworked the `TaskSystem` so that it leverages the `TaskFlowGraph`
    - Removed the concept of `Processors` it wasn't necessary
    - `TaskDrivers` are enabled and working now 
- Added a `ResolveChannelAttribute` which lets you annotate that piece of data is a channel that other data can resolve to. 
    - Ex.  
    - ```csharp
        //The TimerCompletedDataStream can be written to by someone who tries to write to the channel with enum Completed
        //They don't need a direct reference to this variable, just the enum.
        [ResolveChannel(TimerInstance.ResultType.Completed)]
        public readonly ProxyDataStream<TimerCompletedInstance> TimerCompletedDataStream;
        ```
    - Ex.
    - ```csharp
        //Note that we just specify the enum this data can be resolved to, which might be up to N different DataStreams depending on the context of the instance we're updating. 
        ConfigureUpdateJob(UpdateTimerInstancesJob.Schedule)
               .ScheduleOn(TimerDataStream, BatchStrategy.MaximizeChunk)
               .RequireDataStreamForUpdate(TimerDataStream)
               .RequireResolveChannel(TimerInstance.ResultType.Completed);
       ```
     - This leads into supporting the `Cancel` workflow or any other workflow we might need in the future. 
- `JobConfig` and `JobData` got a bit makeover.
    - Much simpler, deleted a ton of code/files.
    - Switched to a public facing interface API so that I can lock down what you are able to do and better guide you to create valid jobs. Intellisense will do all the work.
    - Added support for `EntityQuery` Hooray! 
        - Only for `NativeArray<Entity>` at the moment, but adding support for the `ToComponentArray<IComponentData>` will be trivial in the future.
        - This allows you to schedule your jobs based on the result of an `EntityQuery` and also to get data to read from. 
- Implemented a `TaskFlowGraph`
   - This maintains the relationships between `TaskSystems`, `TaskDrivers` and their Data and Jobs. 
   - It's not 100% finished, the final building of the relationships is still needed but eventually you could spit out a map of how the whole program interacts. 
   - For now, we're using it to build optimized data structures once "hardened" to updating data and scheduling jobs.
   - We'll also be able to use it for safety checks.
       - Ex. You don't have an Update job for this data so it's going to get wiped and you'll have data loss.
       - Ex. You don't have a Populate job for this data so it's never going to run. 
       - Ex. You don't have a job that handles this ResolveChannel so you'll never be able to react to it. 
       - The idea being that through the whole "flow" of the app, we want to be alerted to dead-ends, islands or loss of data scenarios so that we don't have to ever debug that. 
   

- None

 - None

 - [x] Yes - No one is using this yet. It's cool.
 - [ ] No


Scripts/Runtime/Entities/Tasks/AbstractTaskSystem.cs
```diff
@@ -1,246 +1,236 @@
...
-
-            m_PendingDataStreams = new HashSet<IProxyDataStream>();
-            m_SystemTaskProcessors = new List<ISystemTaskProcessor>();
+            Context = m_TaskDriverIDProvider.GetNextID();
 
-            CreateProxyDataStreams();
+            //TODO: Talk to Mike about this. The World property is null for the default world because systems are created via Activator.CreateInstance.
+            //TODO: They don't go through the GetOrCreateSystem path. Is this the case for other worlds? Can we assume a null World is the default one?
+            World currentWorld = World ?? World.DefaultGameObjectInjectionWorld;
+            m_TaskFlowGraph = currentWorld.GetOrCreateSystem<TaskFlowDataSystem>().TaskFlowGraph;
```
jkeon, line 45: I know you ran into how this all works with the multi world stuff. Any thoughts?

Scripts/Runtime/Entities/ProxyDataStream/AbstractProxyDataStream.cs
```diff
@@ -0,0 +1,45 @@
...
+        {
+            get;
+        }
+
+        private readonly string m_DebugString;
+
+        protected AbstractProxyDataStream()
+        {
+            Type = GetType();
+            m_DebugString = $"{Type.Name[..^2]}<{Type.GenericTypeArguments[0].Name}>";
```
mbaker3, line 28: Does this actually work? I thought you needed C#8
image

Scripts/Runtime/Entities/Tasks/Attributes/ResolveChannelAttribute.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+
+namespace Anvil.Unity.DOTS.Entities
+{
+    [AttributeUsage(AttributeTargets.Field, AllowMultiple = true)]
+    public class ResolveChannelAttribute : Attribute
```
mbaker3, line 6: As discussed think about changing the use of the term channel to something else.
 - ResolveStream? ResolveTarget? - I thinkI like target best because it implies writing
 - Make attribute name more fluent ResolveTargetFor()

Scripts/Runtime/Entities/Tasks/Attributes/ResolveChannelAttribute.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+
+namespace Anvil.Unity.DOTS.Entities
+{
+    [AttributeUsage(AttributeTargets.Field, AllowMultiple = true)]
```
mbaker3, line 5: `AttributeTargets.Property` is ok too?

Scripts/Runtime/Entities/Tasks/ITaskSystem.cs
```diff
@@ -1,7 +1,17 @@
...
-        
+        public byte Context
+        {
+            get;
+        }
+
+        public World World
+        {
+            get;
+        }
```
mbaker3, line 15: 🚲 you can inline these if you want. My style preference has always been.
``` csharp
public int AutoGet { get; }

public int OneLineGet
{
   get => GetMyInt();
}

public int MultiLineGet
{
   get
   {
      int i = GetMyInt()+1;
      return i;
   }
}
```

Scripts/Runtime/Entities/Tasks/TaskFlow/Nodes/AbstractNodeLookup.cs
```diff
@@ -0,0 +1,30 @@
+using Anvil.CSharp.Core;
+
+namespace Anvil.Unity.DOTS.Entities
+{
+    internal abstract class AbstractNodeLookup : AbstractAnvilBase
+    {
+        
+        public TaskFlowGraph TaskFlowGraph
```
mbaker3, line 8: Could probably just call TaskGraph?

Scripts/Runtime/Entities/Tasks/TaskFlow/Nodes/DataStreamNode.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        public override string ToString()
+        {
+            return $"{DataStream} located in {TaskDebugUtil.GetLocation(TaskSystem, TaskDriver)}";
+        }
+
+        public void RegisterAsResolveChannel(ResolveChannelAttribute resolveChannelAttribute)
+        {
+            Type type = resolveChannelAttribute.ResolveChannel.GetType();
```
mbaker3, line 44: If this happens a lot should the type get cached in the attribute?

Scripts/Runtime/Entities/Tasks/TaskFlow/TaskFlowDataSystem.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        public TaskFlowDataSystem()
+        {
+            TaskFlowGraph = new TaskFlowGraph();
+        }
+
+        protected override void OnStartRunning()
+        {
+            base.OnStartRunning();
+            TaskFlowGraph.Harden();
```
mbaker3, line 21: 🚲 Consider building in a way that the graph is read by a factory type so there is no Harden/Bake/Render method. You'd instead do `TaskRelationshipFactory.Build(TaskFlowGraph)`

That way we can use the graph as the data and pass it into various different factories for debugging, visual rendering, etc... in the future. It also keeps the responsibilities separated.

Scripts/Runtime/Entities/Tasks/Util/ResolveChannelUtil.cs
```diff
@@ -0,0 +1,44 @@
...
+
+            Debug_EnsureEnumIsSizedProperly(type);
+        }
+        
+        [Conditional("ENABLE_UNITY_COLLECTIONS_CHECKS")]
+        private static void Debug_EnsureEnumIsSizedProperly(Type type)
+        {
+            int sizeOfType = UnsafeUtility.SizeOf(type);
+            int sizeOfByte = UnsafeUtility.SizeOf<byte>();
+
```
mbaker3, line 37: Do you also need to ensure that it's Assignable to `byte`?

Scripts/Runtime/Entities/Tasks/Util/TaskDebugUtil.cs
```diff
@@ -0,0 +1,12 @@
+namespace Anvil.Unity.DOTS.Entities
+{
+    internal static class TaskDebugUtil
+    {
+        public static string GetLocation(ITaskSystem taskSystem, ITaskDriver taskDriver)
```
mbaker3, line 5: 🚲 `GetLocationName`?

Scripts/Runtime/Entities/Tasks/AbstractTaskSystem.cs
mbaker3: World isn't set until `OnCreate()`
`World.AddSystem_OnCreate_Internal` is where the World reference gets passed into a system

jkeon: Ok thanks, I've added a TODO to fix this.

Scripts/Runtime/Entities/ProxyDataStream/AbstractProxyDataStream.cs
jkeon: Apparently we have access to C# 9.0 in Unity 2021.3 with a few caveats. 

https://docs.unity3d.com/Manual/CSharpCompiler.html

Everything is compiling and working.

Scripts/Runtime/Entities/ProxyDataStream/AbstractProxyDataStream.cs
```diff
@@ -0,0 +1,45 @@
+using Anvil.CSharp.Core;
+using Anvil.Unity.DOTS.Jobs;
+using System;
+using System.Reflection;
+using Unity.Jobs;
+
+namespace Anvil.Unity.DOTS.Entities
+{
+    public abstract class AbstractProxyDataStream : AbstractAnvilBase
```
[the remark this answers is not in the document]
jkeon: Hmmn maybe, I am a little confused now on where Proxy should go. This "DataStream" is being used as a proxy for Entity data. 

I've added a TODO to finalize this once and for all.

Scripts/Runtime/Entities/Tasks/Attributes/ResolveChannelAttribute.cs
jkeon: Cool, I like `ResolveTarget` as well. Renamed. 

I think the attribute still makes sense as:

```csharp
[ResolveTarget(TimerInstance.ResolveTarget.Completed)]
public readonly TaskStream<TimerCompletedInstance> TimerCompletedDataStream;
```

Because it's saying the `TimerCompletedDataStream` is a ResolveTarget for the enum.

Scripts/Runtime/Entities/Tasks/Attributes/ResolveChannelAttribute.cs
jkeon: As of today, no. I've added a TODO to expand support for it. 

Reflection is just easier to do on fields so I did that first.

Scripts/Runtime/Entities/Tasks/ITaskSystem.cs
jkeon: Yes you are right. As I see these I'm editing to make those changes. I have done this specific file of course already.

Scripts/Runtime/Entities/Tasks/TaskFlow/Nodes/AbstractNodeLookup.cs
jkeon: Done

Scripts/Runtime/Entities/Tasks/TaskFlow/Nodes/DataStreamNode.cs
jkeon: Good idea, done

Scripts/Runtime/Entities/Tasks/TaskFlow/TaskFlowDataSystem.cs
jkeon: Added a TODO to rework this. I agree it's better to do it that way, just more work right now.

Scripts/Runtime/Entities/Tasks/Util/ResolveChannelUtil.cs
jkeon: It's not assignable, that's only for inheritance. But I can check via:

`if (Enum.GetUnderlyingType(type) != BYTE_TYPE)`

so now I am.

Scripts/Runtime/Entities/Tasks/Util/TaskDebugUtil.cs
jkeon: Sure, fixed.

Scripts/Runtime/Entities/ProxyDataStream/AbstractProxyDataStream.cs
mbaker3: 🤷 ok then. Fun!

Scripts/Runtime/Entities/ProxyDataStream/AbstractProxyDataStream.cs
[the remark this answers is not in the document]
mbaker3: ohh I see. I misunderstood. I thought this was a proxy for a data stream. Should it be an `EntityDataStream` then?
I'm fine to leave it as is too.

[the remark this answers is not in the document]
jkeon: I'm going with `EntityProxyDataStream`, `IEntityProxyInstance`  etc.

We have data that we want to associate with an `Entity` but we can't so we associate it with the Proxy instead. 

At least that's my latest read of the definition of `Proxy` and it seems to make sense. 

The good news is that you don't directly interact with this class anymore. It's all internal.

PR #60, opened 2022-09-21
Resolve Channels (inside job path)

jkeon:
This implements the ability to resolve data to a specific channel and context from inside your job. 

This is useful for the flow where you are processing some data and need to say that it is `Successful` or a `Failure`. 

The same mechanism will be used behind the scenes for you when resolving from outside your job. Ex. Someone `Cancelling` you.

Let's look at `Timers` as a good example for the "inside" case.

```csharp
ConfigureUpdateJob(UpdateTimerInstancesJob.Schedule)
               .ScheduleOn(TimerDataStream, BatchStrategy.MaximizeChunk)
               .RequireDataStreamForUpdate(TimerDataStream)
               .RequireResolveChannel(TimerInstance.ResultType.Completed);
```

Easy peasy, schedule the job, tell it to use the `TimerDataStream` for scheduling and that we're going to update instances in that stream. We also specify that this job can resolve to the `TimerInstance.ResultType.Completed`

You can have N resolve channels as you deem necessary. Multiple success or failure reasons are fine, up to you what makes sense as the author.

In the `UpdateTimerInstancesJob` we either still need to process so we continue:

```csharp
timerInstance.ContinueOn(ref m_TimerDataStreamUpdater);
```

or our Timer is done, so we resolve to the channel

```csharp
timerInstance.Resolve(TimerInstance.ResultType.Completed, 
                                      new TimerCompletedInstance(timerInstance.Entity),
                                      ref m_TimerDataStreamUpdater);
```

Behind the scenes, the system will automatically route the completed instance to the right context (TaskDriver) as appropriate. You as the author don't need to concern yourself with it, enabling your update logic to be generic. 

Also for reference `ResultsDestinations` are now `ResolveChannels`, better name.

- Couldn't resolve without manually jumping through hoops.
- Earlier on, we had to include the `ResultsDestinationPointer` as a property on every single Instance! Which was enforced by an interface. 

Ex. 
- Note the reliance on `IVirtualDataInstance` and how we'd have to specify the type.
- This also meant that we could only ever resolve to one "type". No flexibility.
- Every instance carried along the pointer.
- Note the constructor requiring you to always pass it in too. Bleh!
```csharp
    public struct TimerInstance : IKeyedData,
                                  IVirtualDataInstance<TimerResult>
    {
        public Entity Entity
        {
            get;
        }

        public VDResultsDestination<TimerResult> ResultsDestination
        {
            get;
        }

        public float RemainingTime;

        public TimerInstance(Entity entity, float remainingTime, VDResultsDestination<TimerResult> responseWriter)
        {
            Entity = entity;
            RemainingTime = remainingTime;
            ResultsDestination = responseWriter;
        }
    }
```

- Some more safety checks.
- ResolveChannels are singular for a job. No need to carry alongside the instance anymore.

Ex. 
- Note only one interface.
- Much less code to write
- I have put the enum for the different ResolveChannels in here but it could be external. 
```csharp
public struct TimerInstance : IProxyInstance
    {
        public enum ResolveChannel : byte
        {
            Completed,
            Cancelled
        }

        public Entity Entity
        {
            get;
        }

        public float RemainingTime;

        public TimerInstance(Entity elementSource, float remainingTime)
        {
            Entity = elementSource;
            RemainingTime = remainingTime;
        }
    }
```

- `JobConfig` is more fully featured with ResolveChannels and now hardens into optimized data structures for scheduling and job execution.
- 

- None

 - #59 

 - [x] Yes - No one is using it yet so it's cool.
 - [ ] No

## Discussion

I've highlighted a couple spots in the code to discuss with you. The problem is the way the World worlds with Systems and the Groups.

If I have a system that runs in `InitializationGroup`, it registers jobs etc, but then the `TaskFlowGraph` doesn't harden till later in the `SimulationGroup` or `PresentationGroup`. But before those groups even go, my jobs are starting to run in the `InitializationGroup` and it fails. 

I've gotten around this in a hacky way for now by putting the `TaskFlowGraph` hardening at the end of `PresentationGroup`. This gives all groups a chance to register all possible jobs and data. 

Then in all the `TaskSystems` I check if the `TaskFlowGraph` is hardened yet and if it's not I early return and don't update that frame. 

This means that the first frame, we don't run any jobs or parse any data. 

I had previously thought that ALL systems in every group would go:

- OnCreate
- OnStartRunning
- OnUpdate

But it obviously looks like that's not the case and each group does a full pass before the next group starts. 

You've done a bunch of custom playerloop stuff, so I'm curious to your thoughts here on making this safe and fool proof in the future. 

Can we/should we inject some custom group that always executes last to harden the `TaskFlowGroup` so no one accidentally tries to register jobs after? 

Or should we harden based on Groups? 

I'm not sure what's better. Anyway, to discuss.


Scripts/Runtime/Entities/Tasks/AbstractTaskSystem.cs
```diff
@@ -152,6 +146,12 @@ protected IScheduleUpdateJobConfig ConfigureUpdateJob(IJobConfig.ScheduleJobDele
         //*************************************************************************************************************
         protected override void OnUpdate()
         {
+            //TODO: Discuss with Mike about how we can get around this
```
jkeon, line 149: @mbaker3 See the overall conversation for a discussion point about this.

Scripts/Runtime/Entities/Tasks/TaskFlow/TaskFlowSystem.cs
```diff
@@ -1,24 +1,28 @@
+using Unity.Entities;
+
 namespace Anvil.Unity.DOTS.Entities
 {
     /// <summary>
     /// Data System (no update) for managing the world's <see cref="TaskFlowGraph"/>
     /// </summary>
-    public partial class TaskFlowDataSystem : AbstractDataSystem
+    //TODO: Safer way to handle this. Discussion with Mike.
```
jkeon, line 8: @mbaker3 See the overall conversation for a discussion point about this.

mbaker3:
As discussed with @jkeon I've only read through the front page summary. There is a major refactor PR coming that makes reviewing the code changes in here irrelevant at this point.

## Commenting on the front page summary.
This is beautiful. The API is really starting to fall into place!

I think we've already discussed the discussion section in your post and the specific items you called out in code. Let me know if there's anything left there.

While reviewing this I did run a suggestion by @jkeon that we discussed on a call. Capturing the results of that conversation:
 - Can we potentially eliminate the inner `ResultType` enum and key results on the response type? Answer: Yes!
 - Has some nice boiler plate reduction
    - Elimitates inner result type enum.
    - Can resolve with one less parameter
    ``` csharp
    // old
   timerInstance.Resolve(TimerInstance.ResultType.Completed, 
                                         new TimerCompletedInstance(timerInstance.Entity),
                                         ref m_TimerDataStreamUpdater);
   // new
   timerInstance.Resolve(new TimerCompletedInstance(timerInstance.Entity),
                                      ref m_TimerDataStreamUpdater);
    ```
   - Resolve target configuration is simplified
   ``` csharp
      // old
      [ResolveTargetFor(TimerInstance.ResolveTarget.Cancelled)]
       public readonly TaskStream<TimerCancelledInstance> TimerCancelledDataStream;

      // new
      [ResolveTarget]
       public readonly TaskStream<TimerCancelledInstance> TimerCancelledDataStream;       
    ```

   - Configuring jobs to consume resolve targets now describes type. Self describes types that jobs need to expect and is inline with conventions in the rest of dots (types used for identification)
   ``` csharp
         // old
         ConfigureJobToCancel(Timers, 
                           BatchStrategy.MaximizeChunk,
                           CancelTimerInstancesJob.Schedule)
         .RequireResolveTarget(TimerInstance.ResolveTarget.Cancelled);
         
         // new
         ConfigureJobToCancel(Timers, 
                           BatchStrategy.MaximizeChunk,
                           CancelTimerInstancesJob.Schedule)
         .RequireResolveTarget<TimerCancelledInstance>();
   ```

mbaker3:
Reviewed the Discussion points about Player Loop order. Based on what I'm reading and some limited testing the order of operations per world is:
1. `OnCreate` called for every system in no particular order (this is ok because your systems can call `GetOrCreateSystem<T>()` and get any system you want reference to ahead of time).
   - Note: This will breakdown if there are circular dependencies (SystemA references System B and vice-versa). If this comes up in practice we'll have to take a different approach (see below) custom system at the start of the player loop.
2. `OnStartRuning` called for each system in the first group (usually `InitializeSystemGroup`)
3. `OnUpdate` called for each system in the first group (usually `InitializeSystemGroup`)
4. Repeat Step 2-3 for the second, third, fourth,...n system groups.

I think to solve your issue you want to move your  `TaskFlowSystem` to update at the start of the `InitializationSystemGroup` A good relative position would probably be just before `BeginInitializationEntityCommandBufferSystem` since it's already bringing everything back to the main thread. Attributes something like this:
```csharp
[UpdateBefore(typeof(BeginInitializationEntityCommandBufferSystem))]
[UpdateInGroup(typeof(InitializationSystemGroup))]
```

## Alternative Approaches
If the above doesn't work for whatever reason (circular dependencies, systems needing to get in early, etc..) then we have two more options.
1. Custom Player Loop Phase - We add our own player loop phase and throw the system in there. Nice, but makes usage more complex. Each work you create you'll need to remember to setup that special player loop phase.
2. Implement `ICustomBootstrap` - This gives us full control of the system creation logic and flow. We'd be able to call whatever custom multi-phase init we want on Task Driver Systems. It's a lot to implement for one feature and it would probably be better to save this functionality for application specific changes. Trying to also incorporate Anvil bootstrapping logic complicates things for consumers.

jkeon:
Closing this out.

PR #61, opened 2022-09-26
Cancelling Flow

jkeon:
Building on #60, this is another good milestone for a review to show the evolution of the Task Driver system.

This also contains all fixes/changes and TODO's as a result of #59 's review.

We didn't have a good way to handle dealing with cancellations from the outside. 

- Added some better debug strings for finding where things came from in logs/exceptions.
- Added a `CancelRequestsDataStream`
    - This works to mark the ID's of instances of data that have been requested to be cancelled.
    - The associated job structs like `CancelRequestsReader` feed from it.
-  All `ResolveChannel` names are now `ResolveTarget`. This bloats the files changed a bit. Sorry.
- `DataStreamUpdater` is now very specific to the Cancel flow and runs from a custom job call `ITaskUpdateJobForDefer`. 
    - This is very similar to the more general `IAnvilJobForDefer` job but we can take some really nice liberties in that:
    - We know it's always going to be scheduled on a `DeferredNativeArray`
    - We know it will always be in the `Update` context and therefore needs to respond to cancellation. 
    - As the author, you will get a callback to your `Execute` function ONLY for instances that have no been cancelled. Instances that were cancelled will get written to the cancel target or they will cease to exist if no cancel target exists. 
    - This allows for very fast filtering as part of the update job which is already going to run wide on many threads. 
- `TaskDriver`s and `TaskSystem`s no longer have `ProxyDataStreams` but instead `TaskDataStream` which wrap multiple `ProxyDataStreams` depending if they are cancellable or not. 
    - If a `MoveTo` is cancelled. We probably want to know so we can handle it properly and snap to a grid coordinate.
    - If a `Timer` is cancelled. We probably don't care to unwind it so it can just cease to exist and not be updated anymore. No need to have a separate data stream and job that just does a no-op. 
- JobConfigs reworked a bit for the different contexts of use. Mainly, The `Update` case which is special and then all the other ones. 

- None

 - #60 

 - [x] Yes, but no one is using this anyway yet.
 - [ ] No

## What's next?

With this infrastructure in place, we can now actually cancel and then respond to that cancellation to go from Pending Cancel to Cancel Complete.


mbaker3:
All sounds good. Same as #60 will defer code review until new PR post refactor.

jkeon:
Closing this out.

PR #62, opened 2022-09-28
Cancelling Flow - Part 2

jkeon:
Whereas #61 was about cleanly handling the ability to request a cancel and split pending to be cancelled instances into their own bucket to be processed, this PR is about closing that loop and allowing you to write a "Cancel Job" that will act on those pending instances to gracefully cancel. 

- We weren't able to write specific cancel jobs, anything that was cancelled would sit in a bucket that would then be cleaned up the next consolidate pass.

- Split functionality of `CancelRequestsDataStream` and `ProxyDataStream` so that we can gate the API better and you won't send one or the other where they shouldn't belong.
- Added a `HashCodeUtil` to handle Burstable code for what normally would be `ValueTuple`.
- Added the buckets for cancel requests to sit on both TaskDrivers and TaskSystems.
- Added support for Cancel Jobs to operate on cancelled instances
- Added support for a job to be enabled/disabled and to be run only once. This is useful for setup jobs that we don't want to run every frame.
- Fixed some performance issues with value id's getting boxes and concurrent dictionaries being checked for every job.

- None

 - #61 

 - [x] Yes - But no one is using it yet so it's ok.
 - [ ] No


mbaker3:
Sounds good. All good features! Same as #60 waiting on post-refactor PR.

jkeon:
Closing this out.

Issue #71, opened 2022-10-04
Task System - Move responsibility for creation/disposal into TaskFlowGraph


jkeon:
Most of the data/job related objects are created in the `TaskFlowGraph` and then disposed the same way via `TaskFlowGraph.DisposeFor` 

However some aspects like sub task drivers and cancel requests data stream are still manual and it's a bit weird.

mbaker3:
Does this conflict with the effort in #68 where we want the graph to just be a set of relationships and not anything concrete? Or is this task just some bridge work to wrap up loose ends and #68 will move the logic again when migrating to a factory pattern?

jkeon:
I think this is just some cleanup work to make sure everything is happening in a consistent place. With #68 doing the split I think the ownership/disposal part will be separate from the relationships and it will move.

jkeon:
With this comment: https://github.com/decline-cookies/anvil-unity-dots/pull/77/files#r989150346

It's pretty apparent that the ownership should reside in the TaskSystem/TaskDriver and the disposal happens there too. 

The graph should only handle the relationships.

Issue #68, opened 2022-10-04
Task System - Rework TaskFlowGraph into a Factory


jkeon:
See https://github.com/decline-cookies/anvil-unity-dots/pull/59#discussion_r977823711

Issue #66, opened 2022-10-04
Task System - Build Relationships in TaskFlowGraph


jkeon:
The individual nodes are already there, now we want to do a pass to build up the relationships between each node so we can spit out a "map" of how all the data enters the system, is transformed by various jobs to other data and ultimately either loops back around, exits the system gracefully or blinks out of existence. 

From this we can then raise errors for cases that aren't handled such as:
- When data is not being updated and will be lost
- When data is not being cancelled and will be lost
- When data never gets written to
- etc.

mbaker3:
Echoing a comment I made on an earlier PR about the graph data being its own block of data that can be passed into many types of interpreters (if not now then eventually). This seems to feed into the requirements of #67 so that the following can each be independent pieces of logic that crawl the graph data:
 - Actual graph relationship hardening/baking/creation
 - Relationship validity analysis (probably turned off in release builds)
 - Graph debug rendering/visualization (future feature)

jkeon:
This is a bit different now as the graph has been removed. 

We should still implement the safety analysis and let the user know but this is a future feature as needed.

PR #80, opened 2022-10-05
Remove ResolveTarget Enum and use the type of resolved instance instead

### LINKED ISSUE #64: Task System - Remove ResolveTarget as an Enum and just use the type
See https://github.com/decline-cookies/anvil-unity-dots/pull/60#issuecomment-1265692944

We can remove the enum and instead just use the type.

jkeon:
Mike proposed a good idea here: https://github.com/decline-cookies/anvil-unity-dots/pull/60#issuecomment-1265692944

Where we remove the need for an enum and just use the type of the instance that is being resolved.

- Currently uses an enum to resolve
- It DOES allow for a `TimerInstance` to resolve to a `TimerInstance` for example based on the enum but we decided this wasn't worth the extra steps in creating/maintaining the enum.

- No more enums, simpler attributes.
- You can simply tag your TaskStreams like so:
```csharp
        [ResolveTarget]
        public readonly TaskStream<TimerCompletedInstance> TimerCompletedDataStream;

        [ResolveTarget]
        public readonly TaskStream<TimerCancelledInstance> TimerCancelledDataStream;
```
- When you require the usage in a job you specify the type of instance to resolve to:
```csharp
ConfigureJobToUpdate(Timers,
                                 UpdateTimerInstancesJob.Schedule,
                                 BatchStrategy.MaximizeChunk)
               .RequireResolveTarget<TimerCompletedInstance>();
```
- And when you actually resolve, you just pass in the right instance type.
```csharp
instance.Resolve(new TimerCompletedInstance(instance.Entity),
                                 ref updater);
```
- NOTE: This means that you cannot do something where you resolve a `FailedInstance` to different targets. You would need to specify by using different types. `FailedNoPath` vs `FailedPathTooLong` vs `FailedPathValidButNoAccess` etc. 
    - We've decided this is better practice anyway for debugging and communicating intent. 
-  Fixed a logic error where you wouldn't be able to have different types for resolving in the `JobResolveTargetMapping`.
    - We must have multiple types that a job can resolve to in order to handle Success vs Failure1 vs Failure2 etc.

#### Special Notes

- There's some stinky-ness with aspects of the `TaskFlowGraph` and related nodes but that will be addressed in #66, #68 and/or #71. 
- Burst. Our good friend Burst doesn't like managed types. Which means the `Type` class is a no-no. 
    - We got rid of the enums and we're using the Type now so we need a way to convert Types into some blittable ID.
    - Unity itself handles this via a very convoluted manner in the `TypeManager` but unfortunately we can't just use that system since we're not an `IComponentData` or related types. And frankly, doing so just confuses our intent. They manage it by hashing in a burstable way and using the hash as the id. The good news is that its deterministic although it's possible to run into a collision albeit unlikely. 
    - We've gone a simpler route by just having an `IDProvider` and giving out unique ID's to every type that tried to register as a Resolve Target. 
    - This means it is NOT deterministic so we just have to make sure we're converting to and from Types during Serialization and Deserialization. 
    - The data structure to support this is already setup and the TODO is there for when we get to serialization later on.
    - If we find that this is too much of a pain, we can investigate making our own deterministic burst-compatible hasher but that seems like more work.

- Resolves #64

 - #77

 - [x] Yes, but no one was using it yet so it's ok.
 - [ ] No

### Tag-Alongs
- Fixed a potential issue with the `IDProvider` not being disposed in `ManagedReferenceStore` during a double domain reload.

### DIFF
--- a/Scripts/Runtime/Entities/TaskSystem/Attribute/ResolveTargetForAttribute.cs.meta
+++ b/Scripts/Runtime/Entities/TaskSystem/Attribute/ResolveTargetForAttribute.cs.meta
@@ -1,3 +1,0 @@
-fileFormatVersion: 2
-guid: f3928627c7a347dfa2d9a30b8503499f
-timeCreated: 1663592047
--- a/Scripts/Runtime/Entities/TaskSystem/Job/Delegates.meta
+++ b/Scripts/Runtime/Entities/TaskSystem/Job/Delegates.meta
@@ -1,3 +1,0 @@
-fileFormatVersion: 2
-guid: 84cb735e51d0422c84a22af9dae880cc
-timeCreated: 1664215430


Scripts/Runtime/Entities/ManagedReference/ManagedReferenceStore.cs
```diff
@@ -27,6 +27,7 @@ static ManagedReferenceStore()
         [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.AfterAssembliesLoaded)]
         private static void Init()
         {
+            s_IDProvider?.Dispose();
```
jkeon, line 30: @mbaker3 I noticed this when having a static IDProvider in `ResolveTargetUtil`. I believe this fixes a potential issue but wanted to flag it to you so you can weigh in.

Scripts/Runtime/Entities/Tasks/TaskFlow/Data/ResolveTargetData.cs
```diff
@@ -2,15 +2,8 @@ namespace Anvil.Unity.DOTS.Entities.Tasks
...
-        {
-            get;
-        }
-
-        public byte Context
-        {
-            get;
-        }
+        public AbstractEntityProxyDataStream DataStream { get; }
+        public byte Context { get; }
```
mbaker3, line 6: image

Scripts/Runtime/Entities/Tasks/TaskStream/JobDataInteraction/DataStreamResolver.cs
```diff
@@ -62,7 +62,7 @@ private void Debug_EnsureContainsContext(byte context)
         {
             if (!m_DataStreamByContext.ContainsKey(context))
             {
-                throw new InvalidOperationException($"Trying to get Resolve Channel Data Stream with context of {context} but no data stream exists! Does the data you are expecting to write to have the {nameof(ResolveTargetForAttribute)} attribute?");
+                throw new InvalidOperationException($"Trying to get Resolve Channel Data Stream with context of {context} but no data stream exists! Does the data you are expecting to write to have the {nameof(ResolveTargetAttribute)} attribute?");
```
mbaker3, line 65:
```suggestion
                throw new InvalidOperationException($"Trying to get Resolve Channel Data Stream with context of {context} but no data stream exists! Does the data you are expecting to write to have the {nameof(ResolveTargetAttribute)}?");
```

attribute attribute

mbaker3 (approved):
Looking good. Minor comments, take or leave

> If we find that this is too much of a pain, we can investigate making our own deterministic burst-compatible hasher but that seems like more work.

yea or find an existing community build solution.

Scripts/Runtime/Entities/ManagedReference/ManagedReferenceStore.cs
mbaker3: Today there's no real problem with forgetting to dispose. Stale instances remain listening to an unused instance of IDProvider and probably get garbage collected.

It's good practice though in case some day disposing an ID instance actually does become critical.

jkeon: Ah ok, cool

Scripts/Runtime/Entities/Tasks/TaskFlow/Data/ResolveTargetData.cs
jkeon: Lol

Scripts/Runtime/Entities/Tasks/TaskStream/JobDataInteraction/DataStreamResolver.cs
jkeon: Fixed.

Issue #67, opened 2022-10-04
Task System - Data Integrity Safety Checks


jkeon:
Building on #66, analyze the graph and spit out errors for issues with the data flow.

For developing, this will require some helper jobs for the developer to use for the inbetween states. 

For example, we need to update a piece of data but we're not ready to do so. So we just use a job that keeps it around forever. 
Similar for cancel. 
Similar for data that is never written to yet.

PR #87, opened 2022-10-12
Task Flow Ordering and Responsibility Changes

### LINKED ISSUE #65: Task System - Task Flow Graph/System Startup
See: https://github.com/decline-cookies/anvil-unity-dots/pull/60#issuecomment-1265977475

We want to make sure that the Task Flow Graph gets hardened before anyone else gets to Update. BUT everyone else needs to have the opportunity to populate the graph first. 

Mike's solution here should work and simplify our `Update` in `AbstractTaskSystem`.

### LINKED ISSUE #71: Task System - Move responsibility for creation/disposal into TaskFlowGraph

jkeon:
Fixes an issue with the ordering of TaskSystems, TaskDrivers and the `TaskFlowSystem` with `TaskFlowGraph`.
Also changes the responsibility of who owns various data streams and job configs so that creation and disposal are in the same consistent spot.

- Ordering could prevent some systems from being able to execute jobs because TaskDrivers or other systems haven't been created yet.
- The `TaskFlowGraph` owned the creation and disposal of some but not all data which was weird.
- The reference to `TaskFlowGraphs` would always be the default `World`

- Nicer `ToString` for Drivers and Systems
- TaskDrivers now own their own data and Harden.
- TaskSystems now own their own data too.
    - They get reference to the `TaskFlowGraph` in `OnCreate` so that they get the right graph for the `World` they exist in.
    -  Now have a `ConfigureJobs` function that is overridden to configure jobs. Previously when done in `OnCreate` it was possible for `TaskDrivers` governed by the System to not yet be created and therefore missed in building up the job configuration. Now all jobs are configured in this function and it is guaranteed that all TaskDrivers have been created by then.
- `CancellableTaskStream` is gone. There is just `TaskStream` now. 
    - It was a little weird already along with the `ResolveTarget` being an attribute. Now `Cancellable` is an attribute as well that can be mixed/matched with `ResolveTarget` to give "flags" to how the internals can use the `TaskStream`.
    - This also makes the code paths a lot more consistent.
- `TaskFlowGraph` and all related nodes no longer dispose or own anything.

- Resolves #65 
- Resolves #71 

 - None

 - [x] Yes - No one is using it yet so it's ok.
 - [ ] No

### Notes
 
- There is still some wonkiness in `TaskFlowGraph` and related nodes. This will be addressed in #66, #67 and #68


Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
```diff
@@ -36,6 +38,10 @@ public abstract class AbstractTaskDriver : AbstractAnvilBase
     {
         private readonly List<AbstractTaskDriver> m_SubTaskDrivers;
         private readonly TaskFlowGraph m_TaskFlowGraph;
+        private readonly List<AbstractJobConfig> m_JobConfigs;
+        private readonly string m_TypeString;
```
mbaker3, line 42: 🚲 Rename to TypeName?

How often is this used? Is it actually worth caching?

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
```diff
@@ -48,35 +54,70 @@ public abstract class AbstractTaskDriver : AbstractAnvilBase
         /// </summary>
         public AbstractTaskSystem TaskSystem { get; }
         
+        /// <summary>
+        /// Reference to the associated <see cref="World"/>
+        /// </summary>
+        public World World { get; }
```
mbaker3, line 60: 🚲 Can just pull world off the system and remove need to pass in a world?

```suggestion
        public World World 
        { 
           get => TaskSystem.World;
        }
```

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
```diff
@@ -48,35 +54,70 @@ public abstract class AbstractTaskDriver : AbstractAnvilBase
...
+
         internal CancelRequestsDataStream CancelRequestsDataStream { get; }
+        internal List<AbstractTaskStream> TaskStreams { get; }
+        internal TaskDriverCancellationPropagator CancellationPropagator { get; private set; }
 
         protected AbstractTaskDriver(World world, AbstractTaskSystem abstractTaskSystem)
         {
-            TaskSystem = abstractTaskSystem;
+            World = world;
```
mbaker3, line 68: Here is where you wouldn't need to pass in the world. Would avoid potential errors where `world != abstractTaskSystem.World`

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
```diff
@@ -110,5 +160,18 @@ internal List<CancelRequestsDataStream> GetSubTaskDriverCancelRequests()
 
 
         //TODO: #73 - Implement other job types
+        
+        //*************************************************************************************************************
+        // SAFETY
+        //*************************************************************************************************************
+        
+        [Conditional("ENABLE_UNITY_COLLECTIONS_CHECKS")]
```
mbaker3, line 168: Maybe worth putting under `DEBUG` or `UNITY_ASSERTIONS` since it's not really a collections usage safety check is it?

Scripts/Runtime/Entities/TaskSystem/AbstractTaskSystem.cs
```diff
@@ -15,9 +16,8 @@ namespace Anvil.Unity.DOTS.Entities.Tasks
     public abstract class AbstractTaskSystem : AbstractAnvilSystemBase
     {
         private readonly ByteIDProvider m_TaskDriverContextProvider;
-        private readonly List<AbstractTaskDriver> m_TaskDrivers;
-        private readonly TaskFlowGraph m_TaskFlowGraph;
-        private readonly CancelRequestsDataStream m_CancelRequestsDataStream;
+        private readonly List<AbstractJobConfig> m_JobConfigs;
+        private readonly string m_TypeString;
```
mbaker3, line 20: Same as above.

m_TypeName?
Worth caching?

Scripts/Runtime/Entities/TaskSystem/AbstractTaskSystem.cs
```diff
@@ -27,99 +27,126 @@ public abstract class AbstractTaskSystem : AbstractAnvilSystemBase
...
-            //TODO: #71 - Let the TaskFlowGraph handle creating this for us.
-            m_CancelRequestsDataStream = new CancelRequestsDataStream();
-
-            //TODO: #65 - Need to look at having this happen in OnCreate instead. The World is only set there. 
-            World currentWorld = World ?? World.DefaultGameObjectInjectionWorld;
-            m_TaskFlowGraph = currentWorld.GetOrCreateSystem<TaskFlowSystem>().TaskFlowGraph;
-            m_TaskFlowGraph.CreateTaskStreams(this);
-            m_TaskFlowGraph.RegisterCancelRequestsDataStream(m_CancelRequestsDataStream, this, null);
+            m_TaskFlowGraph = world.GetOrCreateSystem<TaskFlowSystem>().TaskFlowGraph;
+            m_TaskFlowGraph.RegisterTaskSystem(this);
```
mbaker3, line 77: 🚲 May be worth just having a `Register` method on TaskFlow graph with overloads for each type it can handle?

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
```diff
@@ -48,35 +54,70 @@ public abstract class AbstractTaskDriver : AbstractAnvilBase
...
+            TaskStreams = new List<AbstractTaskStream>();
+            m_JobConfigs = new List<AbstractJobConfig>();
+            
+            TaskStreamFactory.CreateTaskStreams(this, TaskStreams);
             CancelRequestsDataStream = new CancelRequestsDataStream();
             
             m_TaskFlowGraph = world.GetOrCreateSystem<TaskFlowSystem>().TaskFlowGraph;
-            m_TaskFlowGraph.CreateTaskStreams(TaskSystem, this);
-            m_TaskFlowGraph.RegisterCancelRequestsDataStream(CancelRequestsDataStream, TaskSystem, this);
+            m_TaskFlowGraph.RegisterTaskDriver(this);
```
mbaker3, line 83: Since this seems to be the only place that you're using the graph should the system just register the task driver with the graph when it gets registered with the system?
Then you don't need reference to the graph.
(Ex: in `TaskSystem.RegisterTaskDriver`)

Scripts/Runtime/Entities/TaskSystem/Job/JobConfig/UpdateJobConfig.cs
```diff
@@ -42,7 +40,7 @@ private void RequireRequestCancelDataStreamForRead(CancelRequestsDataStream canc
                              new CancelRequestsAccessWrapper(cancelRequestsDataStream, AccessType.SharedRead, byte.MaxValue));
         }
         
-        private void RequireCancellableTaskStreamForWrite(CancellableTaskStream<TInstance> cancellableTaskStream)
+        private void RequireCancellableTaskStreamForWrite(TaskStream<TInstance> cancellableTaskStream)
         {
```
mbaker3, line 44: Do you now need to assert that `cancellableTaskStream.IsCancellable == true`?

Scripts/Runtime/Entities/TaskSystem/Job/JobConfigFactory.cs
```diff
@@ -41,7 +41,7 @@ internal static class JobConfigFactory
         public static CancelJobConfig<TInstance> CreateCancelJobConfig<TInstance>(TaskFlowGraph taskFlowGraph,
                                                                                   AbstractTaskSystem taskSystem,
                                                                                   AbstractTaskDriver taskDriver,
-                                                                                  CancellableTaskStream<TInstance> taskStream,
+                                                                                  TaskStream<TInstance> taskStream,
```
mbaker3, line 44: Same, do you need to assert that the provided is cancellable?

Scripts/Runtime/Entities/TaskSystem/TaskFlow/Node/DataStreamNode.cs
```diff
@@ -18,27 +18,17 @@ internal class DataStreamNode : AbstractNode
...
                               AbstractTaskDriver taskDriver,
-                              AbstractTaskStream taskStream,
-                              bool isResolveTarget) : base(taskFlowGraph, taskSystem, taskDriver)
+                              AbstractTaskStream taskStream) : base(taskFlowGraph, taskSystem, taskDriver)
         {
             m_Lookup = lookup;
             DataStream = dataStream;
             TaskStream = taskStream;
-            IsResolveTarget = isResolveTarget;
+            IsResolveTarget = taskStream.IsDataStreamAResolveTarget && taskStream.GetDataStream() == dataStream;
```
mbaker3, line 26: Just so I understand. It's possible to pass a task stream that's unrelated to the data stream? It just won't be a resolve target?

Scripts/Runtime/Entities/TaskSystem/TaskStream/TaskStream.cs
```diff
@@ -7,30 +9,88 @@ namespace Anvil.Unity.DOTS.Entities.Tasks
     /// <see cref="AbstractTaskSystem"/>.
     /// </summary>
     /// <typeparam name="TInstance">The type of <see cref="IEntityProxyInstance"/> in this stream</typeparam>
+    // ReSharper disable once ClassNeverInstantiated.Global
     public class TaskStream<TInstance> : AbstractTaskStream
         where TInstance : unmanaged, IEntityProxyInstance
     {
+        private readonly string m_TypeString;
```
mbaker3, line 16: Same as before

m_TypeName?
Worth caching like this?

Scripts/Runtime/Entities/TaskSystem/TaskStream/TaskStream.cs
```diff
@@ -7,30 +9,88 @@ namespace Anvil.Unity.DOTS.Entities.Tasks
...
+        }
+        
+        internal TaskStream(TaskStreamFlags flags)
         {
-            //TODO: #71 - This is weird with no dispose but #71 will fix it.
+            Flags = flags;
+
+            Type type = GetType();
+
+            //TODO: Extract to Anvil-CSharp Util method -Used in AbstractJobConfig as well
```
mbaker3, line 39: issue number

Scripts/Runtime/Entities/TaskSystem/TaskStream/TaskStream.cs
```diff
@@ -7,30 +9,88 @@ namespace Anvil.Unity.DOTS.Entities.Tasks
...
-        internal TaskStream()
+        internal sealed override bool IsDataStreamAResolveTarget
+        {
+            get => (Flags & TaskStreamFlags.IsResolveTarget) != 0;
+        }
+        
+        internal TaskStream(TaskStreamFlags flags)
         {
-            //TODO: #71 - This is weird with no dispose but #71 will fix it.
+            Flags = flags;
```
mbaker3, line 35: Do we need to assert valid flag combinations? Can this be cancellable without being a resolve target?

Scripts/Runtime/Entities/TaskSystem/TaskStream/TaskStream.cs
```diff
@@ -7,30 +9,88 @@ namespace Anvil.Unity.DOTS.Entities.Tasks
...
         }
     }
+    
+    public abstract class AbstractTaskStream : AbstractAnvilBase
+    {
+        internal abstract bool IsCancellable { get; }
+        internal abstract bool IsDataStreamAResolveTarget { get; }
+        internal abstract AbstractEntityProxyDataStream GetDataStream();
+        internal abstract AbstractEntityProxyDataStream GetPendingCancelDataStream();
+    }
```
mbaker3, line 95: is this worth just making an internal interface that extends `IAnvilDisposable`?

If not, since this is a public type it probably makes more sense in its own file

Scripts/Runtime/Entities/TaskSystem/TaskFlow/TaskFlowSystem.cs
```diff
@@ -19,11 +19,18 @@ public TaskFlowSystem()
             TaskFlowGraph = new TaskFlowGraph();
         }
 
-        protected override void OnUpdate()
+        protected override void OnStartRunning()
         {
+            base.OnStartRunning();
+            
+            TaskFlowGraph.ConfigureTaskSystemJobs();
```
mbaker3, line 26: is it ok that this will be executed more than once per run session? It will be executed anytime the system moves from disabled to enabled.

Obviously this won't happen during normal execution but what if the system is enabled/disabled in the editor?

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
jkeon: Sadly no. 

When a TaskDriver is created, it will do a `World.GetOrCreateSystem<TTaskSystem>` to create the System if it doesn't already exist. 

You would need the World to do that. 

Let's assume we can somehow get around that, there's no guarantee that the System has hit it's `OnCreate` yet so the system's World doesn't exist.

I've added a comment to explain why we don't pull it off though.

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
jkeon: See: https://github.com/decline-cookies/anvil-unity-dots/pull/87/files#r995078637

We wouldn't ever get `world != abstractTaskSystem.World` because the typed generic version creates it for us:

`protected AbstractTaskDriver(World world) : base(world, world.GetOrCreateSystem<TTaskSystem>())`

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
jkeon: Possibly. 

Added TODO to investigate for #66, #67 and/or #68 and linked to this comment.

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
jkeon: We had decided a while back to piggy back on `ENABLE_UNITY_COLLECTIONS_CHECKS` for all safety checks in the `anvil-unity-dots` library. 

The rationale being that we wanted our safety to be disabled and fast whenever we wanted Unity's safety code to be disabled and fast.

Scripts/Runtime/Entities/TaskSystem/AbstractTaskSystem.cs
jkeon: Fair point, removed.

Scripts/Runtime/Entities/TaskSystem/AbstractTaskSystem.cs
jkeon: Added TODO for #66, #67, and/or #68 and referenced this comment.

Scripts/Runtime/Entities/TaskSystem/Job/JobConfig/UpdateJobConfig.cs
jkeon: I had thought no because it's a private function and is gated by an if check above. 

But sure, I added the check just in case code changes happen.

Scripts/Runtime/Entities/TaskSystem/Job/JobConfigFactory.cs
jkeon: Not here but done in `CancelJobConfig`

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
mbaker3: > When a TaskDriver is created, it will do a World.GetOrCreateSystem<TTaskSystem> to create the System if it doesn't already exist.

Not sure I understand. Whatever is creating `AbstractTaskDriver` has reference to a system which it's passing through. That system must have reference to a `World`.

> Let's assume we can somehow get around that, there's no guarantee that the System has hit it's OnCreate yet so the system's World doesn't exist.

Doesn't OnCreate get called before `GetOrCreateSystem<>()` returns?

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
mbaker3: I see, I mean there's an off chance if someone else sub-classes this.
If you change it to
`protected AbstractTaskDriver(World world) : base(world.GetOrCreateSystem<TTaskSystem>())`
then it's guaranteed.

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
mbaker3: OK cool. I don't remember that specifically for all assertions but I'm ok with that. Can we update the README at some point to mention that?

Scripts/Runtime/Entities/TaskSystem/TaskFlow/Node/DataStreamNode.cs
jkeon: A TaskStream can contain one or more DataStreams. 

You as the consuming developer can say that a TaskStream is a ResolveTarget and that will ensure that the primary DataStream is used as the ResolveTarget under the hood. 

But you could also say that the TaskStream is Cancellable so it will have a secondary DataStream in it. That DataStream is not a ResolveTarget. 

This check and organization is bit gross but it does the job today. I've added the TODO to ensure this is looked at for #66, #67 or #68

Scripts/Runtime/Entities/TaskSystem/TaskFlow/TaskFlowSystem.cs
jkeon: Good point, added a gate

Scripts/Runtime/Entities/TaskSystem/TaskStream/TaskStream.cs
jkeon: Removed

Scripts/Runtime/Entities/TaskSystem/TaskStream/TaskStream.cs
jkeon: No need currently. All combinations are valid.

Scripts/Runtime/Entities/TaskSystem/TaskStream/TaskStream.cs
jkeon: Weird that this was missed. Done.

Scripts/Runtime/Entities/TaskSystem/TaskStream/TaskStream.cs
jkeon: I think the problem with the interface is that even though the interface itself is internal, the methods it defines have to be public. So anyone who has reference to a `TaskStream` can see the methods. 

I moved this to its own file.

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
jkeon: Removed

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
jkeon: > Not sure I understand. Whatever is creating AbstractTaskDriver has reference to a system which it's passing through. That system must have reference to a World.

A `TaskDriver` can be created anywhere. In a `MonoBehaviour` in managed land, in a regular `SystemBase`, in another `TaskDriver` as a SubTaskDriver or inside a `TaskSystem`.  We need to pass the right `World` to that TaskDriver so it's associated with the right `TaskSystem` (which may not have been created yet so we'll kick that off via `GetOrCreateSystem`).

- `MonoBehaviour` - Probably through player-interaction in which case we'll know what the focused World is.
- `SystemBase` - We just pass our own World.
- Parent `TaskDriver` - We pass through the same World we got.
- `TaskSystem` - Similar to `SystemBase` we just pass our own World.

When a `TaskDriver` is created, it could be at any time, especially for the lower level ones that will be composed like Timers, or MoveTo's. 

The `TaskSystems` that govern those may not yet have been created. So `TaskDrivers` ensure they get linked to the right governing system by calling `World.GetOrCreateSystem`. 

> Doesn't OnCreate get called before GetOrCreateSystem<>() returns?

No, this is the whole problem. It only creates the instance. `OnCreate` gets called later on. It should be called `GetOrConstructSystem` I guess.

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
jkeon: I can't do that unfortunately because I need both `World` and the `TaskSystem` and there's no guarantee that the `TaskSystem` has it's world yet from this chain https://github.com/decline-cookies/anvil-unity-dots/pull/87#discussion_r995078637

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
jkeon: Yes, added issue #90

Scripts/Runtime/Entities/TaskSystem/TaskStream/TaskStream.cs
mbaker3: Ah yea, good point. All good

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
mbaker3: Discussed on a call and we're pushing the `GetOrCreateSystem()` call down to AbstractTaskDriver so we can be certain the World reference matches the world the system exists in.

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
mbaker3: Addressed: See https://github.com/decline-cookies/anvil-unity-dots/pull/87#discussion_r995893521

PR #88, opened 2022-10-12
Deferred Native Array - Clear Optimization

jkeon:
In a mistaken test, I tried to make 1 million entities walk across the screen instead of the usual 100,000.

When doing so, I noticed that the ConsolidateJob was taking 81ms! 

Upon profiling, I discovered that it was hitting a Large Allocation Fallback for free and allocate every frame. 

With some quick optimizations, I got the 81ms down to 18ms which is a pretty big improvement even though we likely won't ever get that high with number of elements. 

- `EntityProxyDataStream` was freeing and allocating `TempJob` memory every frame for Consolidation. Which normally would be fine but becomes an issue at higher element counts.

- `EntityProxyDataStream` now uses `Persistent` memory and only frees and re-allocates if we grow larger. We can reuse the same bucket for Consolidation. 

 - None

 - None

 - [x] Yes - Just be aware that `Clear` doesn't free memory anymore. This is inline with what `Clear` does on pretty much every other data structure we created and Unity has.
 - [ ] No

### DIFF
--- a/Scripts/Runtime/Data/Collections/DeferredNativeArray.cs
+++ b/Scripts/Runtime/Data/Collections/DeferredNativeArray.cs
@@ -1,12 +1,11 @@
 using System;
-using System.Diagnostics;
 using System.Runtime.InteropServices;
 using Unity.Assertions;
 using Unity.Burst;
 using Unity.Collections;
 using Unity.Collections.LowLevel.Unsafe;
 using Unity.Jobs;
-using Debug = UnityEngine.Debug;
+using UnityEngine;
 
 namespace Anvil.Unity.DOTS.Data
 {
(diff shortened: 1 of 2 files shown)


mbaker3 (commented):
The implementation looks good but I'm not sure that it's appropriate to have a variable length array. Seems like your use case would be better suited to use a NativeList (or equivalent).

This PR basically makes the `DeferredNativeArray` a List without the Add/Remove methods.

For the specific problem you were solving I think there are two approaches.
1. Increase the DOTS reserved memory so that your large temp job allocations don't trigger a `malloc`
2. Use a variable length collection for `m_IterationTarget` with a strictly controlled capacity. Maybe only passing through a read only variant to prevent someone from changing the capacity beyond the deferred creation.

PR #89, opened 2022-10-13
SubTaskDriver Registration

### LINKED ISSUE #74: Task System - Implement Sub Task Driver registration
When building an example with nested TaskDrivers, add in the support to actually register the sub-taskdrivers so we can propagate the cancellation and properly dispose.

jkeon:
Small one to finish up support for SubTaskDrivers. 

- You could create a sub task driver but it wasn't hooked into the system so cancelling wouldn't propagate and disposal was manual.

- Using similar reflection to building `TaskStreams` we now will build all sub task drivers for you and ensure they are registered to their parent and properly disposed.
- They will also properly get the cancellation propagation now.

```csharp
public class WanderTaskDriver : AbstractTaskDriver<WanderTaskSystem>
    {
        //NOTE: We don't have to actually create these. Or dispose them. Happens automatically.
        private readonly TimerTaskDriver m_TimerTaskDriver;
        private readonly MoveToTaskDriver m_MoveToTaskDriver;
        
        [SuppressMessage("ReSharper", "PossibleNullReferenceException")]
        public WanderTaskDriver(World world) : base(world)
        {
            ConfigureJobTriggeredBy(TaskSystem.WanderInstances,
                                    StartWanderJob.Schedule,
                                    BatchStrategy.MaximizeChunk)
               .RequireTaskStreamForWrite(m_TimerTaskDriver.TaskSystem.Timers);

            ConfigureJobTriggeredBy(m_TimerTaskDriver.TimerCompletedDataStream,
                                    MoveJob.Schedule,
                                    BatchStrategy.MaximizeChunk)
               .RequireTaskStreamForWrite(m_MoveToTaskDriver.TaskSystem.MoveToInstances);

            ConfigureJobTriggeredBy(m_MoveToTaskDriver.MoveToCompleteDataStream,
                                    RestartWanderJob.Schedule,
                                    BatchStrategy.MaximizeChunk)
               .RequireTaskStreamForWrite(m_TimerTaskDriver.TaskSystem.Timers);
        }
```

- Resolves #74

 - #87 

 - [x] Yes - You don't need to create them manually anymore. 
 - [ ] No

### DIFF
--- a/Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
+++ b/Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
@@ -78,6 +78,8 @@ namespace Anvil.Unity.DOTS.Entities.Tasks
             
             TaskStreamFactory.CreateTaskStreams(this, TaskStreams);
             CancelRequestsDataStream = new CancelRequestsDataStream();
+
+            TaskDriverFactory.CreateSubTaskDrivers(this, m_SubTaskDrivers, world);
             
             m_TaskFlowGraph = world.GetOrCreateSystem<TaskFlowSystem>().TaskFlowGraph;
             m_TaskFlowGraph.RegisterTaskDriver(this);
--- a/Scripts/Runtime/Entities/TaskSystem/TaskDriverFactory.cs
+++ b/Scripts/Runtime/Entities/TaskSystem/TaskDriverFactory.cs
@@ -1,0 +1,84 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Reflection;
+using Unity.Entities;
+using UnityEngine;
+
+namespace Anvil.Unity.DOTS.Entities.Tasks
+{
+    internal static class TaskDriverFactory
+    {
+        private static readonly Type ABSTRACT_TASK_DRIVER_TYPE = typeof(AbstractTaskDriver);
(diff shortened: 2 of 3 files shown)


Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
```diff
@@ -78,6 +78,8 @@ protected AbstractTaskDriver(World world, AbstractTaskSystem abstractTaskSystem)
             
             TaskStreamFactory.CreateTaskStreams(this, TaskStreams);
             CancelRequestsDataStream = new CancelRequestsDataStream();
+
+            TaskDriverFactory.CreateSubTaskDrivers(this, m_SubTaskDrivers, world);
```
mbaker3, line 82: In a previous PR I talked about potentially just grabbing world from the system passed in.
If that was a good suggestion the same goes here. The World may be read off of the `AbstractTaskDriver`

mbaker3 (approved):

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
jkeon: It would be a good suggestion if `World.GetOrCreateSystem` actually gave the System it's World reference. But unfortunately it only constructs the instance and we still have to wait till later for the `OnCreate` to be called and the system to get a World reference.

jkeon: Changed the API here since you are right, we can get the World off the passed in TaskDriver

PR #91, opened 2022-10-14
Generic Data Access

jkeon:
Added the ability to add arbitrary data to a job and have it be access controlled.

- We used to only allow `NativeArray`
- We told you to be careful and make sure you had access to use it.
- Naming was a little bit inconsistent.

- We now allow any `struct` which includes all `NativeCollections` or nested native collections or custom structs that have nested structs and collections etc inside them.
- We enforce that these must be access controlled via `AccessControlledValue`. This ensures that our job safety is met.
- Naming has been redone a bit so that:
    - "Read", "Reader" etc - represents shared-reading. Many threads can read at the same time.
    - "Write", "Writer" etc - represents shared-writing. Many threads can write to different areas/indices/fields on the data at the same time. Developer assumes some responsibility here to prevent clobbering.
    - "ExclusiveWrite", "Update", "Updater" etc - represents exclusive-writing. Only one thread at a time can access the data for writing and all areas/indicies/fields are fair game. 

- None

 - None

 - [x] Yes - Some API changes but no one was using it.
 - [ ] No

### Some Context

In Pathfinding, we need to have temporary structures that hold the cost so far for a given node and where that node came from. It would look a little something like this:

```csharp
NativeArray<float> costSoFar = new NativeArray<float>(m_PathfindingGrid.CellLength,
                                                                      Allocator.Temp,
                                                                      NativeArrayOptions.ClearMemory);
                NativeArray<int2> cameFrom = new NativeArray<int2>(m_PathfindingGrid.CellLength,
                                                                   Allocator.Temp,
                                                                   NativeArrayOptions.ClearMemory);

```

Note that we're using `Temp` allocator which is fast but there isn't a ton of it. In practice, for long paths, we often hit the Fallback allocator. 

These arrays are created/destroyed everytime we process a Pathfinding request. They are just a scratch pad to perform calculations but aren't used in the final result.

So what if instead, we create these structures outside on the system and allocate them to chunksize as persistent scratch pads.

```csharp
private readonly int m_NumThreads;
        
        private readonly AccessControlledValue<NativeParallelHashMap<int, UnsafeParallelHashMap<int2, float>>> m_ScratchPad_CostSoFar;
        private readonly AccessControlledValue<NativeParallelHashMap<int, UnsafeParallelHashMap<int2, int2>>> m_ScratchPad_CameFrom;
        
        public PathfindTaskSystem()
        {
            m_NumThreads = ParallelAccessUtil.CollectionSizeForMaxThreads;
            m_ScratchPad_CostSoFar = new AccessControlledValue<NativeParallelHashMap<int, UnsafeParallelHashMap<int2, float>>>(new NativeParallelHashMap<int, UnsafeParallelHashMap<int2, float>>(m_NumThreads, Allocator.Persistent));
            m_ScratchPad_CameFrom = new AccessControlledValue<NativeParallelHashMap<int, UnsafeParallelHashMap<int2, int2>>>(new NativeParallelHashMap<int, UnsafeParallelHashMap<int2, int2>>(m_NumThreads, Allocator.Persistent));

            NativeParallelHashMap<int, UnsafeParallelHashMap<int2, float>> costSoFar = m_ScratchPad_CostSoFar.Acquire(AccessType.ExclusiveWrite);
            NativeParallelHashMap<int, UnsafeParallelHashMap<int2, int2>> cameFrom = m_ScratchPad_CameFrom.Acquire(AccessType.ExclusiveWrite);
            for (int i = 0; i < m_NumThreads; ++i)
            {
                costSoFar.Add(i, new UnsafeParallelHashMap<int2, float>(ChunkUtil.MaxElementsPerChunk<int2>(), Allocator.Persistent));
                cameFrom.Add(i, new UnsafeParallelHashMap<int2, int2>(ChunkUtil.MaxElementsPerChunk<int2>(), Allocator.Persistent));
            }
            m_ScratchPad_CostSoFar.Release();
            m_ScratchPad_CameFrom.Release();
        }
```

We now have a hashmap for "costSoFar" and "cameFrom" that is unique per thread. In the job, we don't need to allocate/deallocate and we'll never hit the fallback allocator. We can also now split a long running pathfind request over multiple frames since the scratch pads remain for us.

We can schedule our job like so:

```csharp
ConfigureJobToUpdate(Requests,
                                 PathfindJob.Schedule,
                                 BatchStrategy.MaximizeThreads)
               .RequireDataForWrite(m_ScratchPad_CameFrom)
               .RequireDataForWrite(m_ScratchPad_CostSoFar);
```

And then populate the job via:

```csharp
private PathfindJob(UpdateJobData<PathfindRequest> jobData) : this()
            {
                m_CostSoFarThreadLookup = jobData.GetDataForWriting<NativeParallelHashMap<int, UnsafeParallelHashMap<int2, float>>>();
                m_CameFromThreadLookup = jobData.GetDataForWriting<NativeParallelHashMap<int, UnsafeParallelHashMap<int2, int2>>>();
            }
            
            public void InitForThread(int nativeThreadIndex)
            {
                m_CostSoFar = m_CostSoFarThreadLookup[nativeThreadIndex];
                m_CameFrom = m_CameFromThreadLookup[nativeThreadIndex];
            }
```

It's a nice efficient way to bring in system level data or any other struct to allow for reading/writing.


mbaker3:
Neat idea.
A few thoughts before I get into reviewing this.

> Naming has been redone a bit so that:
> 
>     * "Read", "Reader" etc - represents shared-reading. Many threads can read at the same time.
> 
>     * "Write", "Writer" etc - represents shared-writing. Many threads can write to different areas/indices/fields on the data at the same time. Developer assumes some responsibility here to prevent clobbering.
> 
>     * "ExclusiveWrite", "Update", "Updater" etc - represents exclusive-writing. Only one thread at a time can access the data for writing and all areas/indicies/fields are fair game.

Not sure about this one. Throughout the rest of DOTS "Write" means exclusive write. Shared writing is the exceptional case. It feels unintuitive to change the expectation of Write for AccessController. What drove the change? Just that shared writing is more common than exclusive writing when working with TaskDrivers?

> They are just a scratch pad to perform calculations but aren't used in the final result.

Should we have a higher level scratch pad type that configures all of this for you? Not sure if that's practical.

> We now have a hashmap for "costSoFar" and "cameFrom" that is unique per thread.

Why does each thread get a Hashmap instead of a native array?
With a Hashmap aren't you paying random lookup costs per pathfinding request iteration?
Woudn't it be better to get a NativeArray and the index of the pathfinding request in the chunk? That way you're moving through the chunk and scratch on the same index front to back?

jkeon:
> Not sure about this one. Throughout the rest of DOTS "Write" means exclusive write. Shared writing is the exceptional case. It feels unintuitive to change the expectation of Write for AccessController. What drove the change? 

That's fair. We have just be using `Write` to mean Shared Writing for all of this repo. I just noticed we weren't respecting that in some cases. I haven't changed anything with AccessController, just made sure the function names matched what they were doing.

I'm a-ok to be more explicit and change the API's to SharedWrite/ExclusiveWrite or SharedWrite/Write if you prefer. 

> Just that shared writing is more common than exclusive writing when working with TaskDrivers?

Yep, we want to always shared-write where we can, exclusive write only if we have to.

> Should we have a higher level scratch pad type that configures all of this for you? Not sure if that's practical.

Maybe, I don't know how possible that is because the underlying data is going to be wildly different depending on the context. If I see some patterns emerge then yes we can make some sort of ScratchPad struct.


> Why does each thread get a Hashmap instead of a native array?

Each thread needs to get the inner hashmap because the length of those is often far less than the entire size of grid. In the quick and dirty pathfinding I made before I made those NativeArrays the size of the grid for ease of use. Basically it's better for A* to use the HashMap in this case. (I think. It's trivial to change it to a NativeArray in the future if we have the memory to spare and it proves to be faster)

> With a Hashmap aren't you paying random lookup costs per pathfinding request iteration?

Yes but with the low counts and the hashmap fitting into the cacheline/chunks it's ok.

> Woudn't it be better to get a NativeArray and the index of the pathfinding request in the chunk? That way you're moving through the chunk and scratch on the same index front to back?

It doesn't quite work like that. The pathfinding job can't work on more than one request at a time per thread. So if we have  1000 requests and 16 threads, we can only process them 16 at a time.

So thread 1 uses scratch pad 1, thread 2 uses scratch pad 2 etc. Whatever index the Request is doesn't matter. 

All that said, you raised a good point that the outer bit doesn't need to be a HashMap at all, it can be a native array. Good call!

```csharp
m_ScratchPad_CostSoFar = new AccessControlledValue<NativeArray<UnsafeParallelHashMap<int2, float>>>(new NativeArray<UnsafeParallelHashMap<int2, float>>(m_NumThreads, Allocator.Persistent));
            m_ScratchPad_CameFrom = new AccessControlledValue<NativeArray<UnsafeParallelHashMap<int2, int2>>>(new NativeArray<UnsafeParallelHashMap<int2, int2>>(m_NumThreads, Allocator.Persistent));

            NativeArray<UnsafeParallelHashMap<int2, float>> costSoFar = m_ScratchPad_CostSoFar.Acquire(AccessType.ExclusiveWrite);
            NativeArray<UnsafeParallelHashMap<int2, int2>> cameFrom = m_ScratchPad_CameFrom.Acquire(AccessType.ExclusiveWrite);
            for (int i = 0; i < m_NumThreads; ++i)
            {
                costSoFar[i] = new UnsafeParallelHashMap<int2, float>(ChunkUtil.MaxElementsPerChunk<int2>(), Allocator.Persistent));
                cameFrom[i] = new UnsafeParallelHashMap<int2, int2>(ChunkUtil.MaxElementsPerChunk<int2>(), Allocator.Persistent));
            }
            m_ScratchPad_CostSoFar.Release();
            m_ScratchPad_CameFrom.Release();
```

Scripts/Runtime/Entities/TaskSystem/Job/Wrapper/GenericDataAccessWrapper.cs
```diff
@@ -0,0 +1,32 @@
+using Anvil.Unity.DOTS.Jobs;
+using Unity.Jobs;
+
+namespace Anvil.Unity.DOTS.Entities.Tasks
+{
+    internal class GenericDataAccessWrapper<TData> : AbstractAccessWrapper
```
mbaker3, line 6: Should this just be `DataAccessWrapper`?
I think the only other subclass on `AbstractAccessWrapper` now is the CDFE one right?

mbaker3 (approved):
looks good. Just one naming suggestion

Scripts/Runtime/Entities/TaskSystem/Job/Wrapper/GenericDataAccessWrapper.cs
jkeon: No we have a few:

Screen Shot 2022-10-17 at 3 00 43 PM

I'm ok renaming to `DataAccessWrapper` since it's internal and the user won't ever see it, I just felt like it signified that it was this general case for when it doesn't fit into the more specialized flows.

mbaker3: Nope, good rationale. Thanks!

PR #81, opened 2022-10-06
Added support for CDFE Access

jkeon:
With the Task System, we supported 
- `TaskStreams`
- `NativeArray`
- `EntityQuery`

As ways to require and access data in jobs. 

Now we also support `ComponentDataFromEntity` or CDFE's as a first class citizen.

I ran into this while taking the new TaskSystem for a spin to quickly build up a MoveTo TaskSystem/Driver.

### What is the current behaviour?

Previously if you wanted to use a CDFE you could do so by using a reference to your system and calling:

```csharp
            [ReadOnly] private readonly float m_DeltaTime;
            [ReadOnly] private readonly ComponentDataFromEntity<PathFollowSpeed> m_SpeedsByEntity;
            [NativeDisableParallelForRestriction] private ComponentDataFromEntity<Translation> m_TranslationByEntity;
            
            private MoveToJob(UpdateJobData<MoveToInstance> jobData)
            {
                m_DeltaTime = jobData.Time.DeltaTime;
                m_SpeedsByEntity = jobData.System.GetComponentDataFromEntity<PathFollowSpeed>(true);
                m_TranslationByEntity = jobData.System.GetComponentDataFromEntity<Translation>();
            }
```

Note the need to remember to add the `NativeDisableParallelForRestriction` on the CDFE we want to update and the need to remember to pass in the `readonly = true` optional param for the CDFE we need to read. 

There are also no safety checks that you required access to these when configuring your job. While that doesn't really matter for actual execution, it would be nice for the Task Flow and communicating what data your job is going to touch without having the read the actual job.

### What is the new behaviour?

I added two helper structs similar to `DataStreamWriter` called `CDFEReader` and `CDFEUpdater` along with their various JobConfig and JobData APIs. 

Scheduling a job now communicates that you intend to use them:

```csharp
            ConfigureJobToUpdate(MoveToInstances,
                                 MoveToJob.Schedule,
                                 BatchStrategy.MaximizeChunk)
               .RequireResolveTarget<MoveToComplete>()
               .RequireCDFEForRead<PathFollowSpeed>()
               .RequireCDFEForUpdate<Translation>();
```

And the actual job is more simple and easy to write:

```csharp
            [ReadOnly] private readonly float m_DeltaTime;
            [ReadOnly] private readonly CDFEReader<PathFollowSpeed> m_SpeedsByEntity;
            private CDFEUpdater<Translation> m_TranslationsByEntity;

            private MoveToJob(UpdateJobData<MoveToInstance> jobData)
            {
[...]

PR #94, opened 2022-10-18
Dynamic buffer access

jkeon:
Similar to #81, this adds support for `DynamicBuffer` access.

- Had to manually include `DynamicBuffers` and remember the attributes for the different use cases of read/write.

- Wrapped structs that configure the `DynamicBuffer` for you. 

NOTE: 
- I have not yet included support for the case where you want multiple threads to work on the same `DynamicBuffer` but different indicies. That will come later as its a more specialized case and I feel like it's going to share some work/functionality with the more general case of sharing the same NativeArray or NativeList etc. 

- None

 - None

 - [ ] Yes
 - [x] No

### Tag Alongs
- Some naming fixes and better error messaging.


mbaker3:
> I have not yet included support for the case where you want multiple threads to work on the same DynamicBuffer but different indicies. That will come later as its a more specialized case and I feel like it's going to share some work/functionality with the more general case of sharing the same NativeArray or NativeList etc.

Fair, it's probably just the same case because of `DynamicBuffer.AsNativeArray()`. Aside from any resize functionality.

Scripts/Runtime/Entities/TaskSystem/Job/JobConfig/AbstractJobConfig.cs
```diff
@@ -275,6 +275,32 @@ public IJobConfigRequirements RequireCDFEForWrite<T>()
...
+        //*************************************************************************************************************
+        // CONFIGURATION - REQUIRED DATA - DynamicBuffer
+        //*************************************************************************************************************
+        
+        /// <inheritdoc cref="IJobConfigRequirements.RequireDBFEForRead{T}"/>
+        public IJobConfigRequirements RequireDBFEForRead<T>()
+            where T : struct, IBufferElementData
+        {
+            DynamicBufferAccessWrapper<T> wrapper = new DynamicBufferAccessWrapper<T>(AccessType.SharedRead, TaskSystem);
+            AddAccessWrapper(new JobConfigDataID(typeof(DynamicBufferAccessWrapper<T>.DynamicBufferType), Usage.Read),
```
mbaker3, line 288: Discussed on call. Maybe we should have a token type in abstract wrapper so that all of the wrappers consistently report their type vs some using the raw data type and others some wrapped internal token.

With a comment on this that it's used for a token for job config.

Scripts/Runtime/Entities/TaskSystem/Job/Wrapper/AbstractAccessWrapper.cs
```diff
@@ -1,19 +1,44 @@
...
+#endif
+
+        [Conditional("ENABLE_UNITY_COLLECTIONS_CHECKS")]
+        private void Debug_SetWrapperType()
+        {
+            Type type = GetType();
+            Debug_WrapperType = type.IsGenericType
+                ? type.GetGenericTypeDefinition()
+                : type;
+        }
```
mbaker3, line 42: This will produce a compiler error when `!ENABLE_UNITY_COLLECTIONS_CHECKS`.
You're trying to access `Debug_WrapperType` which will not exist.

mbaker3 (changes requested):
Just need to fix the one compiler error

Scripts/Runtime/Entities/TaskSystem/Job/Wrapper/AbstractAccessWrapper.cs
jkeon: Ooops, I forgot that the function still exists, just isn't called.

mbaker3 (approved):

PR #105, opened 2022-11-08
Cancellation Updates

jkeon:
Cancellation now works properly!

This is _probably_ _**hopefully**_ the last big refactor needed. Everything else should be nicer smaller changes already captured in Issues and behind the scenes.

With this PR, it is ready for be used for gameplay code.

Cancellation sort of works but requires a lot of hoops to jump through, a lot of boiler plate code and you aren't guaranteed a complete flow to the end unless you manually wait for sub completions and aggregate them yourself.

## Major Change - DataStreams

`TaskStream` is gone. We no longer need to wrap `DataStreams`, instead we can use them directly and expose public interfaces to the end developer to communicate intent. 

Screen Shot 2022-11-08 at 2 49 43 PM

#### Internal DataStreams
- **AbstractDataStream** 
    - Untyped and access controlled.
- **AbstractTypedDataStream**
    - Typed to `unmanaged` and contains a `Pending` wide writer to allow for shared writing.
- **AbstractLookupDataStream** 
    - Contains a lookup keyed by the type
    - Wide `Pending` consolidates into this lookup
- **AbstractArrayDataStream** 
    - Contains an array for iteration
    - Wide `Pending` consolidates into this array
- **CancelRequestDataStream**
    - Acts as a lookup to signify which `Entity+TaskDriverContext` has been requested for cancellation.
    - A TaskDriver and a TaskSystem has one of these.
- **DataStream** 
    - Most common use case, it's the data that would normally be run through jobs.
    - If any element's ID (`Entity+TaskDriverContext`) is found in the `CancelRequestDataStream` that corresponds to the TaskDriver or TaskSystem that own's this DataStream, the element will NOT get consolidated into the array. It will just "poof" out of existence.
    - TaskDrivers and TaskSystems can have as many of these as they want.
    - It is expected (but not yet enforced) for there to be an Update job for each of these.
- **CancellableDataStream** 
    - Similar to `DataStream` but has an internal `PendingCancelDataStream`.
    - If any element's ID is found in the `CancelRequestDataStream` instead of poofing out of existence, it will be written to the `PendingCancelDataStream`.
    - TaskDrivers and TaskSystems can have as many of these as they want. 
    - It is expected (but not yet enforced) for there to be an Update job and a Cancel job for each of these.
 - **PendingCancelDataStream** 
    - Same as `DataStream` but won't check `CancelRequestDataStream`. You can't cancel a cancel. 
    - Cancel jobs will read from this stream. 
    - Every `CancellableDataStream` has one of these.
- **CancelCompleteDataStream** 
    - Similar to `DataStream` but instead of holding data, it just holds the ID.
    - Ex. You request an ID to be cancelled, this stream will be populated with ID's that have completed their cancellation work. 
    - A TaskDriver and a TaskSystem has one of these.
- **CancelResultDataStream** 
    - Same as `DataStream` but won't check `CancelRequestDataStream`. You can't cancel a cancel.
    - TaskDrivers and TaskSystems can have as many of these as they want.
    - The intended usage is to schedule a job to trigger on the `CancelCompleteDataStream` to know that the cancellation logic is complete for a given ID. BUT if you need more information than just the ID, you would write to a `CancelResultDataStream` and you could schedule a job to trigger off of that instead or in addition to the `CancelCompleteDataStream`. 
     
#### Public Facing DataStream Interfaces

- **IDataStream** 
    - Most common, use when you don't care if you get cancelled. There is no unwinding.
- **ICancellableDataStream** 
    - Use when you do care if you get cancelled and you need to unwind gracefully.
- **ICancelResultDataStream** 
    - Use when you want to communicate specific results about the unwinding. 

## Major Change - Cancel Flow

I don't really like the name but I can't think of anything better. CancelFlow's are helper objects that preserve the flow of cancellation through a TaskDriver to all systems and sub taskdrivers and handle collapsing back up the chain to report the completion. 

Screen Shot 2022-11-08 at 3 44 23 PM

At any given frame, a job or UI will request to cancel for an `Entity` or group of `Entities` for a certain TaskDriver.
Immediately, all `CancelRequestDataStreams` all the way down the chain will be written to. We want to cancel whatever is happening for our Entity at any of these steps. 

Screen Shot 2022-11-08 at 3 46 25 PM

The next frame, we first consolidate these `CancelRequestDataStreams` which writes a value to the corresponding `CancelProgressLookup`. A value of `true` will mean that we intend to do some processing of the cancellation so we initially set it to true to give this frame a chance to process.

We then consolidate all the data streams. Anything that was an `IDataStream` will see the requested cancel for an `Entity` and will delete it. Poof! It's gone, we didn't need to unwind gracefully.

Anything that is a `ICancellableDataStream` will write to it's `PendingCancelDataStream`. 

The `PendingCancelDataStreams` will consolidate and Cancel Jobs will run off of that to allow for cancellation logic of graceful unwinding to happen is T+1 frames from when the Request was made. 

Screen Shot 2022-11-08 at 3 51 09 PM

Next frame, we consolidate the `CancelProgressLookups`. This works by starting at the deepest level of the chain and working back up the chain in a series of sequential jobs with early outs.

In this example, our `MoveToTaskSystem` is still processing a CancelJob so the lookup marks `True` again to say it's still processing and holds itself open. It also tells it's parent in `MoveTaskDriver` to hold open. 

`MoveTaskDriver` gets to go at the same time as `PathfindTaskSystem` and `TimerTaskSystem` as they are at the same level of depth. 

`MoveTaskDriver` was told to stay open, so it tells its parent to do the same. 
`PathfindTaskSystem` and `TimerTaskSystem` did not have a CancelJob OR a child tell them to stay open so we assume that any cancellation that was going to happen is now done and we write to the complete.

Any completes that happened could be operated on this frame but that's really only for inner TaskDriver logic to respond to inner cancels if that ever arises.

The CancelJob for `MoveToTaskSystem` continues to run this frame.

Screen Shot 2022-11-08 at 3 55 53 PM

Finally, we consolidate again on the next frame and this time the `MoveToTaskSystem` is complete so nothing holds it open and nothing holds open its parent. This cycle of not holding open our parent continues all the way to the top which allows the top level `WanderTaskDriver` to write that the Cancellation is now complete. 

The other progress and completes were taken care of on the previous frame and no longer exist this frame.

## Major Change - Execution Order

As mentioned above, Consolidation of Data Streams now happens at the beginning of the frame. 

This means that it could take longer for a piece of data to make it's way through a set of systems and taskdrivers where as before you could get lucky with the ordering and move through many in one frame. 

We decided this was not a great idea because it meant luck introduced some variability to the performance. Fast forward to end of the project and we feel something is going too slow due to order, so we manually re-order which then causes two other things that were working fine to now to be slow. It's a never ending nightmare and makes it a stack of cards. 

By always having the data consolidated at the beginning, we get worst case performance from day 1 so we can design our data flow and architecture better as it scales. If something is SLOW then we can circumvent by aggregating data or changing the overall flow. 

Most importantly, it ensures that we always have data integrity with a cancellation and don't have orphaned data that wasn't cancelled but should have been.

## Minor Changes
- You don't need to register `DataStreams` as resolve targets anymore. You can resolve to any DataStream on the Driver or System that match your type.
- Misc ordering/moving things around
- Better logging!
- Added `ANVIL_DEBUG_LOGGING_EXPENSIVE` and instrumented the internals with it. Much easier to turn on and off for those times when you want to debug the flow.
- Added `ProfileMarkers` for common jobs so we can see what Type and the corresponding System/Driver responsible for big time chunks. Previously they would all display as `ConsolidateJob`1` which isn't helpful.

- None

 - None

 - [x] Yes - But no one is using it yet
 - [ ] No


Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
```diff
@@ -16,14 +15,20 @@ public abstract class AbstractTaskDriver<TTaskSystem> : AbstractTaskDriver
...
+        /// </remarks>
+        /// TODO: #67 - Should add a Safety check on Job scheduling that we're allowed to write
+        /// TODO: #100 - OR we don't need this to be hidden because we're allowed to write to TaskSystem data.
+        protected new TTaskSystem TaskSystem
         {
             get => (TTaskSystem)base.TaskSystem;
         }
 
-        protected AbstractTaskDriver(World world) : base(world, typeof(TTaskSystem))
+        // MIKE: I can't think of a better pattern here.
```
jkeon, line 30: I need to pass a parent in here because I need to know if I'm a top level TaskDriver or if I was created as a SubTaskDriver. 

But this kind of sucks because then you have to do:

`new WanderTaskDriver(World, null);`

But I can't call a different constructor when building through the sub task driver path because that's a user land constructor.

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
```diff
@@ -53,34 +56,40 @@ public abstract class AbstractTaskDriver : AbstractAnvilBase
...
 
-            TaskStreamFactory.CreateTaskStreams(this, TaskStreams);
-            CancelRequestsDataStream = new CancelRequestsDataStream();
-
-            TaskDriverFactory.CreateSubTaskDrivers(this, m_SubTaskDrivers);
+            TaskData = new TaskData(this, TaskSystem);
+            CancelFlow = new TaskDriverCancelFlow(this, Parent?.CancelFlow);
+            TaskDriverFactory.CreateSubTaskDrivers(this, SubTaskDrivers);
+            
+            //MIKE - Ordering Question, not sure if there is a better flow
```
jkeon, line 90: I'm not sure if there's a better way to do this. 

A `TaskDriver` can have a parent or if null is a top level TaskDriver. We're guaranteed this now by the constructor arg. 

But a `CancelFlow` needs a parent as well and it's not quite the same mapping because while a TaskSystem can have multiple TaskDrivers, a `TaskDriverCancelFlow` must have a unique `SystemCancelFlow`. There are N `SystemCancelFlows` for every `TaskDriver` a `TaskSystem` has.

Which means, I need to get reference to the parent CancelFlow when creating the sub task drivers and I need the whole chain to be complete in order to build up the request data. 

Writing it out, I'm not sure there is a solution, but it just feels gross.

jkeon:
@mbaker3 This is still good to look at for the overall description. I wouldn't worry too too much about the code. 

I've solved a bunch of the other in progress problems with some re-architecting for sharing Pending Data.

Scripts/Runtime/Entities/TaskSystem/Job/Wrapper/CancelCompleteDataStreamAccessWrapper.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        public override JobHandle Acquire()
+        {
+            return CancelCompleteDataStream.AccessController.AcquireAsync(AccessType);
+        }
+
+        public override void Release(JobHandle releaseAccessDependency)
+        {
+            CancelCompleteDataStream.AccessController.ReleaseAsync(releaseAccessDependency);
+        }
```
mbaker3, line 25: How complex do these get in other variants?
Could the `AbstractAccessWrapper` just taken in the `AccessController` instance and do the acquire/release rather than creating a specialty type for each type of stream.

If it's not always access controllers should we have an interface that describes something as access controlled?
Ex: `IAccessControlled` with `AcquireAsync` and `ReleaseAsync` then you can pass anything that satisfies that contract without needing specialty wrappers.

Scripts/Runtime/Entities/TaskSystem/Job/Wrapper/CancelCompleteDataStreamAccessWrapper.cs
```diff
@@ -0,0 +1,27 @@
...
+        public CancelCompleteDataStream CancelCompleteDataStream { get; }
+
+        public CancelCompleteDataStreamAccessWrapper(CancelCompleteDataStream cancelCompleteDataStream,
+                                                     AccessType accessType,
+                                                     AbstractJobConfig.Usage usage) : base(accessType, usage)
+        {
+            CancelCompleteDataStream = cancelCompleteDataStream;
+        }
+
+        public override JobHandle Acquire()
```
mbaker3, line 17: Should we add the async suffix to this and `Release` so it's consistent with the rest of the lib?

Scripts/Runtime/Entities/TaskSystem/Job/Wrapper/PendingCancelDataStreamAccessWrapper.cs
```diff
@@ -0,0 +1,26 @@
...
+        where T : unmanaged, IEntityProxyInstance
+    {
+        public PendingCancelDataStream<T> PendingCancelDataStream { get; }
+
+        public PendingCancelDataStreamAccessWrapper(PendingCancelDataStream<T> pendingCancelDataStream, AccessType accessType, AbstractJobConfig.Usage usage) : base(accessType, usage)
+        {
+            PendingCancelDataStream = pendingCancelDataStream;
+        }
+
+        public sealed override JobHandle Acquire()
```
mbaker3, line 16: should the class be `sealed` too?

Scripts/Runtime/Entities/TaskSystem/TaskData/DataStream/Abstract/AbstractArrayDataStream.cs
```diff
@@ -0,0 +1,142 @@
...
+#endif
+
+namespace Anvil.Unity.DOTS.Entities.Tasks
+{
+    internal abstract class AbstractArrayDataStream<T> : AbstractTypedDataStream<T>,
+                                                         IArrayDataStream
+        where T : unmanaged
+    {
+        //Deliberately not using pointers because that messes up what the safety handle pointer points to.
+        protected DeferredNativeArray<T> Live;
```
mbaker3, line 21: Not a fan of this name. `Current`, `Active`, `Data`?

Should also be `m_` prefixed right?

Scripts/Runtime/Entities/TaskSystem/TaskData/DataStream/Abstract/AbstractArrayDataStream.cs
```diff
@@ -0,0 +1,142 @@
...
+            {
+#if DEBUG
+                m_ProfilerMarker.Begin();
+#endif
+                m_Live.Clear();
+
+                int pendingCount = m_Pending.Count();
+                NativeArray<T> liveArray = m_Live.DeferredCreate(pendingCount);
+                m_Pending.CopyTo(ref liveArray);
+                m_Pending.Clear();
```
mbaker3, line 127: Does this not mean you need write access to `m_Pending`?

Scripts/Runtime/Entities/TaskSystem/TaskData/DataStream/Abstract/AbstractArrayDataStream.cs
```diff
@@ -0,0 +1,142 @@
...
+            AccessController.ReleaseAsync(dependsOn);
+            return dependsOn;
+        }
+
+        //*************************************************************************************************************
+        // JOBS
+        //*************************************************************************************************************
+
+        [BurstCompile]
+        private struct ConsolidateArrayDataStreamJob : IJob
```
mbaker3, line 79: Not necessary now but I feel like there should just be CopyAsync methods on `UnsafeTypedStream and/or `DeferredNativeArray` so we don't need these generic jobs that do nothing unique beyond copy.

Scripts/Runtime/Entities/TaskSystem/TaskData/DataStream/Abstract/AbstractDataStream.cs
```diff
@@ -0,0 +1,82 @@
...
+
+#if DEBUG
+            Debug_ProfilerMarker = new ProfilerMarker(ProfilerCategory.Scripts, ToString(), MarkerFlags.Script);
+#endif
+#if ANVIL_DEBUG_LOGGING_EXPENSIVE
+            Debug_DebugString = new FixedString128Bytes(ToString());
+#endif
+        }
+
+        protected sealed override void DisposeSelf()
```
mbaker3, line 52: I think I brought this up in an issue but we want to consider if an async dispose method is useful (`Dispose(JobHandle)`)

mbaker3, line 52: I'm not sure about sealing `DisposeSelf`

Why not require each subclass to acquire before they're going to call dispose on their data type(s)? This allows others to implement dispose logic beyond just data streams.

mbaker3 (approved):
Generally looks good. I think my biggest wish is to clean up the #if checks all over. I've provided some suggestions for cleanup.

> This is probably hopefully the last big refactor needed. Everything else should be nicer smaller changes already captured in Issues and behind the scenes.
> 
> With this PR, it is ready for be used for gameplay code.

🤣  what a change 1 month makes.

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
jkeon: I have managed to fix this now in my current work. Publicly you can only create the constructor that doesn't take in a parent. 

With the `TaskDriverFactory`, when generating the sub TaskDrivers, it will call a generated version that sets the parent.

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
jkeon: Ok thanks

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
```diff
@@ -16,14 +15,20 @@ public abstract class AbstractTaskDriver<TTaskSystem> : AbstractTaskDriver
...
-        /// Hides the base reference to the abstract version.
         /// </summary>
-        public new TTaskSystem TaskSystem
+        /// <remarks>
+        /// Hides the base reference to the abstract version. This is so that from the outside, a developer
+        /// doesn't try and select a DataStream to configure a job on.
+        /// </remarks>
+        /// TODO: #67 - Should add a Safety check on Job scheduling that we're allowed to write
+        /// TODO: #100 - OR we don't need this to be hidden because we're allowed to write to TaskSystem data.
+        protected new TTaskSystem TaskSystem
```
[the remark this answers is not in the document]
jkeon: No longer relevant with the recent changes. 

Before It was when you were scheduling jobs. In your derived TaskDriver you would want access to your System's data, but outside TaskDrivers should not be allowed to access the System data directly.

Scripts/Runtime/Entities/TaskSystem/AbstractTaskDriver.cs
```diff
@@ -53,34 +56,40 @@ public abstract class AbstractTaskDriver : AbstractAnvilBase
...
-            TaskStreamFactory.CreateTaskStreams(this, TaskStreams);
-            CancelRequestsDataStream = new CancelRequestsDataStream();
-
-            TaskDriverFactory.CreateSubTaskDrivers(this, m_SubTaskDrivers);
+            TaskData = new TaskData(this, TaskSystem);
+            CancelFlow = new TaskDriverCancelFlow(this, Parent?.CancelFlow);
+            TaskDriverFactory.CreateSubTaskDrivers(this, SubTaskDrivers);
+            
+            //MIKE - Ordering Question, not sure if there is a better flow
+            CancelFlow.BuildRequestData();
```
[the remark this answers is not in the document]
jkeon: This is now fixed in my recent work. 

Before this was needed because there was a pass to build down the tree to create all the SubTaskDrivers and it's data but then we had to crawl back up to setup the links of request data and whether things were cancellable. 

Today it's a build phase in the constructor, and then a Harden phase for being ready for job execution and is much cleaner and easier to follow.

Scripts/Runtime/Entities/TaskSystem/AbstractTaskSystem.cs
```diff
@@ -37,23 +31,21 @@ public abstract class AbstractTaskSystem : AbstractAnvilSystemBase
...
             m_TaskDriverContextProvider = new ByteIDProvider();
-            TaskStreams = new List<AbstractTaskStream>();
             TaskDrivers = new List<AbstractTaskDriver>();
             m_JobConfigs = new List<AbstractJobConfig>();
 
             Context = m_TaskDriverContextProvider.GetNextID();
 
-            TaskStreamFactory.CreateTaskStreams(this, TaskStreams);
-            CancelRequestsDataStream = new CancelRequestsDataStream();
+            TaskData = new TaskData(null, this);
```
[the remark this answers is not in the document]
jkeon: Agreed and no longer relevant in the new work

Scripts/Runtime/Entities/TaskSystem/Cancelling/AbstractCancelFlow.cs
```diff
@@ -0,0 +1,247 @@
...
+
+        protected TaskData TaskData { get; }
+        protected TaskDriverCancelFlow Parent { get; }
+
+#if DEBUG
+        private ProfilerMarker Debug_ProfilerMarker { get; }
+#endif
+#if ANVIL_DEBUG_LOGGING_EXPENSIVE
+        private FixedString128Bytes Debug_DebugString { get; }
+#endif
```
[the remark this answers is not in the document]
jkeon: Agreed, I think I had them protected before and forgot to change back.

Scripts/Runtime/Entities/TaskSystem/Cancelling/AbstractCancelFlow.cs
```diff
@@ -0,0 +1,247 @@
...
+                                                                                       Parent?.TaskDriverContext ?? 0
+#if DEBUG
+                                                                                      ,
+                                                                                       Debug_ProfilerMarker
+#endif
+#if ANVIL_DEBUG_LOGGING_EXPENSIVE
+                                                                                      ,
+                                                                                       Debug_DebugString
+#endif
+                                                                                      );
```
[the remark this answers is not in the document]
jkeon: Oh ok, that's nice, I like the separate method. Thanks!

Scripts/Runtime/Entities/TaskSystem/Cancelling/AbstractCancelFlow.cs
```diff
@@ -0,0 +1,247 @@
...
+            }
+
+            private void HandleProgress(bool isStillProcessing, ref EntityProxyInstanceID id)
+            {
+                //If we're still processing...
+                if (isStillProcessing)
+                {
+#if ANVIL_DEBUG_LOGGING_EXPENSIVE
+                    Debug.Log($"Still processing for {id.ToFixedString()} on {m_DebugString} - Holding open");
+#endif
```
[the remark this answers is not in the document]
jkeon: Fair point, I'll look into that.

Scripts/Runtime/Entities/TaskSystem/Job/Wrapper/CancelCompleteDataStreamAccessWrapper.cs
jkeon: I will double check on this. Good point.

Scripts/Runtime/Entities/TaskSystem/Job/Wrapper/PendingCancelDataStreamAccessWrapper.cs
jkeon: Good point, will check

Scripts/Runtime/Entities/TaskSystem/TaskData/DataStream/Abstract/AbstractArrayDataStream.cs
jkeon: I'll go with `Active`. 

With the `m_` prefix, yes it should technically. I wanted it to be a getter but that messes up the safety pointer. So this was a bit of a special case. I'm not sure it's still relevant but I'll save this to see what I can do.

Scripts/Runtime/Entities/TaskSystem/TaskData/DataStream/Abstract/AbstractArrayDataStream.cs
jkeon: That's a good point. Will add an issue.

Scripts/Runtime/Entities/TaskSystem/TaskData/DataStream/Abstract/AbstractDataStream.cs
jkeon: In my mind it was so that anyone making derived DataStreams didn't have to remember to do that acquisition. It felt much safer to acquire for disposal, allow the derived classes to clean up and then dispose the access controller.

Scripts/Runtime/Entities/TaskSystem/TaskData/DataStream/Abstract/AbstractArrayDataStream.cs
mbaker3: May be able to do a `ref` getter to help with the safety pointer.

Scripts/Runtime/Entities/TaskSystem/TaskData/DataStream/Abstract/AbstractDataStream.cs
mbaker3: hmmm, yea I'm just worried about constraining other disposal logic. Maybe we just override `Dispose`?

I don't think it's a big deal to ask subclasses to acquire before they perform a write operation but see where you're coming from.

Scripts/Runtime/Entities/TaskSystem/TaskData/DataStream/Abstract/AbstractArrayDataStream.cs
mbaker3: but `m_Pending` has the `[ReadOnly]` attribute?

jkeon:
Closing because #142 handles this.

PR #109, opened 2022-11-09
AccessControlledValue Disposal - Acquire for Dispose first

### LINKED ISSUE #104: AccessControlledValue Dispose Method
An `AccessControlledValue` today just disposes the `AccessController` and tries to dispose the value inside.

```csharp
protected override void DisposeSelf()
        {
            m_AccessController.Dispose();
            (m_Value as IDisposable)?.Dispose();
            base.DisposeSelf();
        }
```

When you call `AccessControlledValue.Dispose()` you might run into a problem where the underlying data is still in flight. 

So the pattern is to acquire synchronously for disposal first:

```csharp
m_MyAccessControlledValue.Acquire(AccessType.Disposal);
m_MyAccessControlledValue.Dispose();
```

Seems error-prone to forget about this and annoying to write, should we just do it automatically in the Dispose?

```csharp
protected override void DisposeSelf()
        {
            m_AccessController.Acquire(AccessType.Disposal);
            m_AccessController.Dispose();
            (m_Value as IDisposable)?.Dispose();
            base.DisposeSelf();
        }
```

jkeon:
Added automatic acquiring for disposal when disposing.

You must (or really really should) manually acquire for Disposal first. 
Otherwise when disposing, the data may be in flight leading to a crash/exception being thrown.

```csharp
m_MyAccessControlledValue.Acquire(AccessType.Disposal);
m_MyAccessControlledValue.Dispose();
```

The `Dispose` method automatically acquires for you so you don't forget.

- Resolves #104 

 - None

 - [x] Yes - If you're already manually acquiring you'll need to change that so you're not acquiring twice. However at runtime it will squawk at you for acquiring twice making it quick and easy to fix.
 - [ ] No

### DIFF
--- a/Scripts/Runtime/Job/AccessControl/AccessControlledValue.cs
+++ b/Scripts/Runtime/Job/AccessControl/AccessControlledValue.cs
@@ -28,9 +28,12 @@ namespace Anvil.Unity.DOTS.Jobs
 
         protected override void DisposeSelf()
         {
-            m_AccessController.Acquire(AccessType.Disposal);
+            if (m_Value is IDisposable disposable)
+            {
+                m_AccessController.Acquire(AccessType.Disposal);
+                disposable.Dispose();
+            }
             m_AccessController.Dispose();
-            (m_Value as IDisposable)?.Dispose();
             base.DisposeSelf();
         }
 
--- a/Scripts/Runtime/Job/AccessControl/AccessControlledValue.cs
+++ b/Scripts/Runtime/Job/AccessControl/AccessControlledValue.cs
@@ -28,6 +28,7 @@ namespace Anvil.Unity.DOTS.Jobs
 
         protected override void DisposeSelf()
         {
+            m_AccessController.Acquire(AccessType.Disposal);
             m_AccessController.Dispose();
             (m_Value as IDisposable)?.Dispose();
             base.DisposeSelf();


jkeon:
@mbaker3 This is still good to look at. Quick and easy and I think it a good change.

Scripts/Runtime/Job/AccessControl/AccessControlledValue.cs
```diff
@@ -28,6 +28,7 @@ public AccessControlledValue(T value)
 
         protected override void DisposeSelf()
         {
+            m_AccessController.Acquire(AccessType.Disposal);
             m_AccessController.Dispose();
             (m_Value as IDisposable)?.Dispose();
             base.DisposeSelf();
```
mbaker3, line 34: If `m_Value` isn't disposable should we really require a main thread sync on the write handle?

```suggestion
            if (m_Value is IDisposable disposable)
            {
                m_AccessController.Acquire(AccessType.Disposal);
                disposable.Dispose();
            }
            m_AccessController.Dispose();
            
            base.DisposeSelf();
```

mbaker3: Another thought. There is a comment in `AccessController` to call complete on the relevant job handles when we call `AccessController.Dispose` should we just do that instead? That seems like a more robust solution.

mbaker3 (dismissed):
Yea, this is a good addition. As mentioned in other places we should probably support disposal scheduling at some point.

One take it or leave it point in the review.

Scripts/Runtime/Job/AccessControl/AccessControlledValue.cs
mbaker3: Right, ok that makes sense. Sounds like as a future feature we may want to make a burstable version of AccessControlledValue and/or some pattern to set off a main thread task after a job handle is complete.

> When we dispose an AccessController, we're just saying that we can no longer use that AccessController, but the already scheduled jobs that are in flight I think are fine to continue onward.

This feels uncomfortable but makes sense. I guess we'll find out if this comes back to bite us one day.

mbaker3 (approved):

PR #110, opened 2022-11-09
Script Define Toggle

### LINKED ISSUE #106: Editor menu tool to toggle compiler flags
Just a quick simple menu to make it quick and easy to toggle:

ENABLE_UNITY_COLLECTIONS_CHECKS - Might as well have it in this menu for ease of finding rather than going through Unity's built in.
ANVIL_DEBUG_SAFETY_EXPENSIVE - For expensive checks to ensure integrity (performance will degrade)
ANVIL_DEBUG_LOGGING_EXPENSIVE - For expensive log spam (use ideally with only one entity or a handful so you can debug flow issues)

If we have this, does it make sense to not bother using ENABLE_UNITY_COLLECTIONS_CHECKS anymore and instead have our own ANVIL_DEBUG_SAFETY instead? 

Should these be DOTS only? Or at the csharp-core level so that all submodules benefit from the same thing?

jkeon:
Added a simple menu for toggling on/off Script Defines

You would have to manually toggle them on/off in Player Settings and remember what they were called.

You can use a menu `Anvil > Script Defines > ANVIL_DEBUG_ETC`

- Resolves #106 

 - None

 - [ ] Yes
 - [x] No

## Discussion

In the Issue #106 I raised a couple of questions. In building this I think I've answered them but let's make sure we're on the same page.

1. I think the defines should exist throughout Anvil. CSharp, Unity Core and Unity Dots. To that end I think this PR should be deleted and this Toggle be part of Unity Core. If there are specific Unity Dots flags it can add via the same API.

2. I looked into ENABLE_UNITY_COLLECTIONS_CHECKS and it's something that is ALWAYS on in the Editor and ALWAYS off in the Player. You can only toggle it on/off for Burst code in the Editor. 

I think we should do a find replace on all our code (new issue) to remove any references to ENABLE_UNITY_COLLECTIONS_CHECKS and replace with ANVIL_DEBUG_SAFETY

### DIFF
--- a/Scripts/Editor/Compiler/ScriptDefinesToggle.cs
+++ b/Scripts/Editor/Compiler/ScriptDefinesToggle.cs
@@ -1,0 +1,86 @@
+using System.Collections.Generic;
+using System.Linq;
+using UnityEditor;
+using UnityEditor.Build;
+
+namespace Anvil.Unity.DOTS.Editor.Compiler
+{
+    /// <summary>
+    /// Provides a quick and easy way to enable/disable commonly used Scripting Defines for
+    /// development and debugging. 
+    /// </summary>
+    [InitializeOnLoad]
+    public static class ScriptDefinesToggle
+    {
+        private const string MENU_PATH_BASE = "Anvil/Script Defines";
+        private const string DEFINE_ANVIL_DEBUG_SAFETY_EXPENSIVE = "ANVIL_DEBUG_SAFETY_EXPENSIVE";
+        private const string MENU_PATH_ANVIL_DEBUG_SAFETY_EXPENSIVE = MENU_PATH_BASE + "/" + DEFINE_ANVIL_DEBUG_SAFETY_EXPENSIVE;
+        private const string DEFINE_ANVIL_DEBUG_LOGGING_EXPENSIVE = "ANVIL_DEBUG_LOGGING_EXPENSIVE";
+        private const string MENU_PATH_ANVIL_DEBUG_LOGGING_EXPENSIVE= MENU_PATH_BASE + "/" + DEFINE_ANVIL_DEBUG_LOGGING_EXPENSIVE;
+
+        private static readonly HashSet<string> ACTIVE_DEFINES = new HashSet<string>();
+        private static readonly NamedBuildTarget NAMED_BUILD_TARGET;
+
+        static ScriptDefinesToggle()
+        {
+            BuildTarget buildTarget = EditorUserBuildSettings.activeBuildTarget;
+            BuildTargetGroup buildTargetGroup = BuildPipeline.GetBuildTargetGroup(buildTarget);
+            NAMED_BUILD_TARGET = NamedBuildTarget.FromBuildTargetGroup(buildTargetGroup);
+            
+            ACTIVE_DEFINES.Clear();
+            PlayerSettings.GetScriptingDefineSymbols(NAMED_BUILD_TARGET, out string[] defines);
+            foreach (string define in defines)
+            {
+                ACTIVE_DEFINES.Add(define);
+            }
+        }
+        
+        [MenuItem(MENU_PATH_ANVIL_DEBUG_SAFETY_EXPENSIVE, true)]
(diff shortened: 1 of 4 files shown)


jkeon:
@mbaker3 Still good to look at. Worthwhile change, albeit some discussions or choices needed.

Scripts/Editor/Compiler/ScriptDefinesToggle.cs
```diff
@@ -0,0 +1,86 @@
...
+            BuildTarget buildTarget = EditorUserBuildSettings.activeBuildTarget;
+            BuildTargetGroup buildTargetGroup = BuildPipeline.GetBuildTargetGroup(buildTarget);
+            NAMED_BUILD_TARGET = NamedBuildTarget.FromBuildTargetGroup(buildTargetGroup);
+            
+            ACTIVE_DEFINES.Clear();
+            PlayerSettings.GetScriptingDefineSymbols(NAMED_BUILD_TARGET, out string[] defines);
+            foreach (string define in defines)
+            {
+                ACTIVE_DEFINES.Add(define);
+            }
```
mbaker3, line 35:
```suggestion
            ACTIVE_DEFINES.UnionWith(defines);
```

Scripts/Editor/Compiler/ScriptDefinesToggle.cs
```diff
@@ -0,0 +1,86 @@
...
+using UnityEditor.Build;
+
+namespace Anvil.Unity.DOTS.Editor.Compiler
+{
+    /// <summary>
+    /// Provides a quick and easy way to enable/disable commonly used Scripting Defines for
+    /// development and debugging. 
+    /// </summary>
+    [InitializeOnLoad]
+    public static class ScriptDefinesToggle
```
mbaker3, line 13: Thinking this should be DEBUG specific rather than a central place for all defines.
Same goes for the `Compiler` namespace probably belongs under `Debug` or core instead.

Scripts/Editor/Compiler/ScriptDefinesToggle.cs
```diff
@@ -0,0 +1,86 @@
...
+    {
+        private const string MENU_PATH_BASE = "Anvil/Script Defines";
+        private const string DEFINE_ANVIL_DEBUG_SAFETY_EXPENSIVE = "ANVIL_DEBUG_SAFETY_EXPENSIVE";
+        private const string MENU_PATH_ANVIL_DEBUG_SAFETY_EXPENSIVE = MENU_PATH_BASE + "/" + DEFINE_ANVIL_DEBUG_SAFETY_EXPENSIVE;
+        private const string DEFINE_ANVIL_DEBUG_LOGGING_EXPENSIVE = "ANVIL_DEBUG_LOGGING_EXPENSIVE";
+        private const string MENU_PATH_ANVIL_DEBUG_LOGGING_EXPENSIVE= MENU_PATH_BASE + "/" + DEFINE_ANVIL_DEBUG_LOGGING_EXPENSIVE;
+
+        private static readonly HashSet<string> ACTIVE_DEFINES = new HashSet<string>();
```
mbaker3, line 21: Can probably remove this. Just fetch the define symbols on each validator or set call.
Caching active defines here adds unnecessary complexity to this script.

Working directly with the values returned by `PlayerSettings` means there's a single point of truth.

mbaker3 (changes requested):
Not approving because this should be in `anvil-unity-core`

> 1. I think the defines should exist throughout Anvil. CSharp, Unity Core and Unity Dots. To that end I think this PR should be deleted and this Toggle be part of Unity Core. If there are specific Unity Dots flags it can add via the same API.

Agreed

> I think we should do a find replace on all our code (new issue) to remove any references to ENABLE_UNITY_COLLECTIONS_CHECKS and replace with ANVIL_DEBUG_SAFETY

Agreed

Would also like to suggest that we change to ANVIL_LOGGING_VERBOSE vs Expensive.

jkeon:
Closing this as it's been dealt with in the Unity core repo: https://github.com/decline-cookies/anvil-unity-core/pull/108

Issue #108, opened 2022-11-09
Custom profiler module - DataStreams


jkeon:
Unity allows for creating custom profiler modules.

We can leverage this to collect info on the data streams for memory usage and lifecycle information.

jkeon:
I've removed the profiling for now since we had re-architected things and it was difficult to keep it in sync. The appropriate TODO's and links to the older PRs are preserved for when this gets re-implemented.

mbaker3:
Does this need to get prioritized? The profiler seemed really valuable during your development.

jkeon:
Not the full wazoo and visual  aspect but yes some little pieces will go in sooner rather than later.

mbaker3:
OK, will need to break out that task and schedule it into the project

jkeon:
Agreed.

PR #111, opened 2022-11-15
DataStream Custom Profiler - Initial Feature

jkeon:
A simplified version of a custom profiler for DataStreams

There isn't a profiler for DataStreams

There now is a profiler for DataStreams! And we discovered we were dramatically over allocating. 

Screen Shot 2022-11-15 at 9 54 34 AM

- Linked to #108 but doesn't fully complete it yet.

 - #105 

 - [ ] Yes
 - [x] No

### Questions

- This profiler introduces a syncpoint at the end of a frame for getting the data into the profiler. Should we wrap the profiling aspects into a `ANVIL_DEBUG_PROFILING` define and update #106?


jkeon:
@mbaker3 Doesn't need to be looked at too hard. My architecture changes are going to move things around a bunch so I might end up re-adding the profiler as a separate step.

Scripts/Runtime/Entities/TaskSystem/TaskData/DataStream/Profiling/DataStreamProfilingSystem.cs
```diff
@@ -0,0 +1,60 @@
...
+            }
+
+            m_HasInitialized = true;
+
+            m_TaskFlowGraph.AddAllDataStreamsTo(m_AllDataStreams);
+            
+            base.OnStartRunning();
+        }
+
+        protected override void OnUpdate()
```
mbaker3, line 47: Should this only be enabled/update when the profiler is set to record?

Scripts/Runtime/Entities/TaskSystem/TaskData/DataStream/Profiling/DataStreamProfilingSystem.cs
```diff
@@ -0,0 +1,60 @@
...
+namespace Anvil.Unity.DOTS.Entities.Tasks
+{
+#if DEBUG
+    /// <summary>
+    /// System that will handle populating profiling counters once per frame when guaranteed that all writing jobs
+    /// have completed.
+    /// </summary>
+    //TODO: #86 - Revisit with Entities 1.0 for "Create Before/After"
+    //TODO: #108 - Revisit to see if there is a better way to handle acquire/release. Or can we do it all in a job? 
+    [UpdateInGroup(typeof(PresentationSystemGroup), OrderLast = true)]
```
mbaker3, line 14: We don't have this group in the simulation worlds right now. We should probably have this happen at the very start of the initialization group or end of the simulation group.

I do have thoughts around introducing some new phases into the PlayerLoop for Anvil but I'm not sure how soon those will be introduced yet and if they're a good idea.

Scripts/Runtime/Entities/TaskSystem/TaskData/DataStream/Profiling/DataStreamProfilingSystem.cs
```diff
@@ -0,0 +1,60 @@
...
+        protected override void OnUpdate()
+        {
+            foreach (AbstractDataStream dataStream in m_AllDataStreams)
+            {
+                dataStream.AccessController.Acquire(AccessType.SharedRead);
+                
+                dataStream.PopulateProfiler();
+                
+                dataStream.AccessController.Release();
+            }
```
mbaker3, line 56: Seems like you could build a native array of structs that contain `ProfilerData` structs. Go through each datastream and AcquireAsync. Then create a job that is given:
 - data stream
 - native array of profiler structs and index to write to (or a pointer to the right element) 

The job would pull the profiler data off of the datastream and write it to the correct element in the native array. Then the native array gets read when the profiler updates.

mbaker3 (commented):
Looks good for a first pass. Would just like to address some folder and namespace concerns.

> - This profiler introduces a syncpoint at the end of a frame for getting the data into the profiler. Should we wrap the profiling aspects into a `ANVIL_DEBUG_PROFILING` define and update #106?

Yes but each toggle should be handled by its own script.

jkeon:
Closing, #142 supercedes this and we're deferring profiling until later. Captured in #108

PR #112, opened 2022-11-16
Cancellation Optimization

jkeon:
As mentioned in #107, we can optimize the cancel flow. 

### What is the current behaviour?

Currently, when a Top Level TaskDriver is cancelled, we allow each SubTaskDriver and System to have the opportunity to run one frame and let us know if we should keep processing the cancel request or not get anything back and we assume we're done and complete the cancellation. 

See the following log:
```
[RandoCancelTaskDriver] - Rando Cancel - Timer Complete, Cancelling WanderTaskDriver
~~ Frame N+1 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
CancelRequestDataStream, Driver: FollowPathTaskDriver|2, System: FollowPathTaskSystem - Count 1
CancelRequestDataStream, System: FollowPathTaskSystem - Count 1
CancelRequestDataStream, System: PathfindTaskSystem - Count 1
CancelRequestDataStream, Driver: WanderTaskDriver|2, System: WanderTaskSystem - Count 1
CancelRequestDataStream, System: WanderTaskSystem - Count 1
CancelRequestDataStream, Driver: PathfindTaskDriver|2, System: PathfindTaskSystem - Count 1
CancelRequestDataStream, System: MoveToTaskSystem - Count 1
CancelRequestDataStream, System: TimerTaskSystem - Count 1
CancelRequestDataStream, Driver: MoveToTaskDriver|2, System: MoveToTaskSystem - Count 1
CancelRequestDataStream, Driver: TimerTaskDriver|2, System: TimerTaskSystem - Count 1
Still processing for Entity(1:1) - Context: 2 on SystemCancelFlow, System: MoveToTaskSystem - Holding open
Still processing for Entity(1:1) - Context: 2 on SystemCancelFlow, System: PathfindTaskSystem - Holding open
Still processing for Entity(1:1) - Context: 2 on SystemCancelFlow, System: TimerTaskSystem - Holding open
Still processing for Entity(1:1) - Context: 2 on SystemCancelFlow, System: FollowPathTaskSystem - Holding open
Still processing for Entity(1:1) - Context: 2 on TaskDriverCancelFlow, Driver: MoveToTaskDriver|2, System: MoveToTaskSystem - Holding open
Still processing for Entity(1:1) - Context: 2 on SystemCancelFlow, System: WanderTaskSystem - Holding open
[...]

PR #113, opened 2022-11-18
Allow for writing directly to system data

### LINKED ISSUE #100: Allow TaskDriver jobs to write directly to TaskSystem data
A current boiler plate pattern is for a job to write something like a `WanderStart` to a TaskDriver which then runs a job on `WanderStart`s to turn them into `WanderInstance`s on the TaskSystem. 

The only reason for this is so that the TaskDriver's context is written into that `WanderInstance`. 

Not only is it tedious for the developer to write the data and the job, it takes up additional memory, unnecessary processing and a one frame delay to get into the system.

We should be able to remove this and have jobs write from taskdrivers directly to the TaskSystem and since the job knows it was from a TaskDriver, the context can be passed along to the writer.

This might already be happening now that I think about it, but needs to be investigated. 

Do this will remove a bunch of unnecessary data/jobs and simplify the overall code. 

The `TaskDriver` references to `TaskSystem` can also be updated since we don't need the `new` operator to hide aspects from the outside.

jkeon:
Instead of writing to Task Driver data and then having a TaskDriver job to convert into System data we can now write directly to the system data.

If you were to write directly to system data, there would be no context associated. When the System Data's update job executes and tries to write to the completed bucket it wouldn't know where to put it. 

To work around that you would write to task driver data (which would give it context) and the task driver would have it's own job to then write to the system data. The System Data's update job executes, there is context and it writes to the correct completed bucket. 

The problem with this is that it's duplicated memory for no good reason and a ton of boiler plate for the developer to write which is annoying. It's also additional pressure on the scheduler for a superficial job. 

`IDataStream` -> Now `IDriverDataStream` and `ISystemDataStream`
Same deal for CancellableDataStream and CancelResultDataStream.

I don't really like this but it's the best way to solve it from the most compiler catching way and not relying on the developer to implement a getter in their TaskDrivers a proper way. 

`Driver` data is owned by Drivers. 
`System` data is owned by Systems.

At runtime on startup, a runtime error will be thrown if you mix these up in the reflection generation pass so we're safe there.

The API for configuring jobs has been changed for the `Write` case so that when you write to `System` data, you need to supply the corresponding `TaskDriver` for context. 

`.RequireDataStreamForWrite(m_TimerTaskDriver.Timers, m_TimerTaskDriver);`

This should be pretty simple because that's where you likely got access to the System data from anyway. 

- Resolves #100 

 - #112 

 - [x] Yes - No one is using this yet so it's ok. 
 - [ ] No


jkeon:
@mbaker3 Quick glance only, the re-architecting solves this as everything is defined and configured on the TaskDriver only.

mbaker3 (approved):

jkeon:
Closing this as #142 handles this now.

PR #121, opened 2022-12-07
Add MathExtension

mbaker3:
Add a collection of extension methods to aid in working with `Unity.Mathematics` from the perspective of DOTS.
This is the DOTS compliment to https://github.com/decline-cookies/anvil-unity-core/pull/96

There are some operations that are cumbersome or error prone when working with `Unity.Mathematics`

Add the following extension methods:
 - `GetScale(float4x4)` - Get the signed scale component from a 4x4 matrix.
 - `GetScaleMagnitude(float4x4)` - Get the unsigned scale component from a 4x4 matrix.
 - `GetRotation(float4x4)` - Get the rotation component from a 4x4 matrix.
 - `GetTranslation(float4x4)` - Get the translation component from a 4x4 matrix.
 - `IsValidTransform(float4x4)` - Evaluate whether a 4x4 matrix represents a valid transform

- None

 - #120 

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Mathematics/MathExtension.cs
+++ b/Scripts/Runtime/Mathematics/MathExtension.cs
@@ -1,0 +1,147 @@
+using System.Runtime.CompilerServices;
+using Anvil.CSharp.Mathematics;
+using Unity.Mathematics;
+
+namespace Anvil.Unity.DOTS.Mathematics
+{
+    /// <summary>
+    /// A collection of extension methods for working with types under the <see cref="Unity.Mathematics"/> namespace.
+    /// (float3, int2, etc..)
+    /// </summary>
+    public static class MathExtension
+    {
+        /// <summary>
+        /// Get the inverse of a <see cref="float3"/>.
+        /// </summary>
+        /// <remarks>
+        /// Any components that are 0 will invert to <see cref="float.NaN"/>.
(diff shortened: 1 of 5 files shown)


Scripts/Runtime/Mathematics/MathExtension.cs
```diff
@@ -0,0 +1,99 @@
+using System.Runtime.CompilerServices;
+using Unity.Mathematics;
+
+namespace Anvil.Unity.DOTS.Mathematics
+{
+    /// <summary>
+    /// A collection of extension methods for working with types under the <see cref="Unity.Mathematics"/> namespace.
+    /// (float3, int2, etc..)
+    /// </summary>
+    public static class MathExtension
```
jkeon, line 10: Same questions as PR in Unity Core about `MathematicsExtension` instead of `MathExtension`

jkeon (changes requested):
I don't think this needs to be DOTS specific. These are all general to transformation matrices which can be used in Unity Core.

jkeon (approved):
Nah it's fine. From a purity perspective I think it could be moved down, but for ease of getting this merged, I think it's fine to keep here until we run into a use-case to pull it down.

Issue #116, opened 2022-11-30
TransformUtil - Investigate support for non uniform scale transforms


mbaker3:
The `TransformUtil` rotation and scale conversion methods don't produce correct results for transform matrices that contain non-uniform scaling operations.

Ideally, the values represent a perfect conversion between spaces (local and world). However perfect conversion may not be possible because even in the Unity editor world orientation/scale isn't preserved.

If perfect conversion isn't possible are we able to match Unity or at least get correct results for the special cases that Unity produces correct results for (single axis rotations, etc...)

## Example
Ex: Moving a `GameObject` from world space into the local space of another `GameObject` that is rotated and has a non uniform scale.

![Nov-30-2022 09-38-19](https://user-images.githubusercontent.com/1426795/204824756-6a4dd3b1-9aea-4a43-9553-731d0e28ff2f.gif)

PR #122, opened 2022-12-07
Add `TransformUtil`

mbaker3:
Add a collection of utility methods for working with spacial transformations in DOTS.

There are limited built in operations to convert between local and world contexts in Unity's transform system. Having to repeatedly re-implement the calculations is cumbersome and error prone...particularly when handling edge cases.

Add methods that provide the ability to convert the following components between local to world space (and back)
 - Scale
 - Rotation
 - Position
 - Rectangles

These conversions can be performed through either a `LocalToWorld` instance or valid 4x4 matrix(`float4x4) represented.

Note: Transformations that contain non-uniform scale operations are not currently supported. #116 

**Unit Tests**
This PR also includes a robust set of unit tests to help clarify expected behaviour and ensure future optimizations and additional features maintain accurate output.

To help generate expected values I hacked together a quick script that applies transformations to Unity's `GameObject` based hierarchy and prints out the values. This script doesn't have a logical home so it lives as a gist at the link below.
https://gist.github.com/mbaker3/c1f1e83aacb80879c88af8c4638bdfdd

- None

 - #121 

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Editor/Tests/Entities/Transform/TransformUtilTests.cs
+++ b/Scripts/Editor/Tests/Entities/Transform/TransformUtilTests.cs
@@ -1,13 +1,8 @@
-using System.Collections.Generic;
-using System.Collections.ObjectModel;
 using System.Text.RegularExpressions;
 using Anvil.CSharp.Logging;
-using Anvil.CSharp.Mathematics;
 using Anvil.Unity.DOTS.Entities.Transform;
 using Anvil.Unity.DOTS.Mathematics;
-using Anvil.Unity.Logging;
 using NUnit.Framework;
-using Unity.Collections;
 using Unity.Mathematics;
 using Unity.Transforms;
 using UnityEngine;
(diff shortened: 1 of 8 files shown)


Scripts/Editor/Tests/Entities/Transform/TransformUtilTests.cs
```diff
@@ -0,0 +1,4391 @@
...
+            LogAssert.Expect(LogType.Error, s_NonUniformScaleError);
+            TransformUtil.ConvertWorldToLocalScale(worldToLocal_nonUniform, -point_two);
+            LogAssert.Expect(LogType.Error, s_NonUniformScaleError);
+            TransformUtil.ConvertWorldToLocalScale(worldToLocal_nonUniform, point_zero);
+            LogAssert.Expect(LogType.Error, s_NonUniformScaleError);
+            TransformUtil.ConvertWorldToLocalScale(worldToLocal_nonUniform, new float3(2, -2, 0));
+            Logger.Debug("END: Expected error messages.");
+#endif
+
+            // float3x3 worldToLocal_nonUniformRotation = new float3x3(TransformUtil.ConvertWorldToLocalRotation(worldToLocal_nonUniform, quaternion.identity));
```
jkeon, line 1782: There are a few of these commented out sections. Should we just delete all the commented out ones?

jkeon (changes requested):
One minor change to remove commented out code, otherwise, looks good.

Scripts/Editor/Tests/Entities/Transform/TransformUtilTests.cs
mbaker3: They were all supposed to have this comment block above but I see I missed some.
The idea is that the day you go to implement #116 you have ready made tests to start working with.

jkeon: Ah gotcha

jkeon (approved):

PR #125, opened 2022-12-15
Move `ComponentDataFromSingleEntity` and `BufferDataFromSingleEntity` to extension methods

mbaker3:
Move `ComponentDataFromSingleEntity` and `BufferDataFromSingleEntity` out of `AbstractAnilSystemBase` to extension methods so they may be used when working with any stock entities system/CDFE/BDFE.

Developers must subclass `AbstractAnvilSystemBase` to get access to convenience methods to create single entity CDFE/BDFE instances.

Developers get access to the convenience methods when using any stock CDFE, BDFE, or `ComponentSystemBase`

 - None

 - None

 - [ ] Yes
 - [x] No - The extension methods have the same signature, they're just defined in a new location and more widely available. It should just work ™️

### DIFF
--- a/Scripts/Runtime/Entities/Data/ComponentDataFromSingleEntity.cs
+++ b/Scripts/Runtime/Entities/Data/ComponentDataFromSingleEntity.cs
@@ -10,6 +10,9 @@ namespace Anvil.Unity.DOTS.Entities
     /// <remarks>Allows developers to define jobs with fewer parameters that clearly communicate intent.</remarks>
     public readonly struct ComponentDataFromSingleEntity<T> where T : struct, IComponentData
     {
+        //TODO: #115 - Implement a safety check to make sure there isn't another lookup/entity combination in flight.
+        // Until the above is implemented leave it to consuming jobs to add the attributes.
+        // [NativeDisableContainerSafetyRestriction]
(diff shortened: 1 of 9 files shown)


Scripts/Runtime/Entities/Data/ComponentDataFromSingleEntity.cs
```diff
@@ -25,7 +28,7 @@ public ComponentDataFromSingleEntity(ComponentDataFromEntity<T> lookup, Entity e
         }
 
         /// <summary>
-        /// Gets the <see cref="DynamicBuffer{T}" />. 
+        /// Gets the <see cref="DynamicBuffer{T}" />.
```
jkeon, line 31: I know this said DynamicBuffer before but that isn't correct for this struct.

jkeon (changes requested):
Looks good, one small docs issue

Scripts/Runtime/Entities/Data/ComponentDataFromSingleEntity.cs
mbaker3: Good catch, fixed.

jkeon (approved):

PR #126, opened 2022-12-16
Native extensions

mbaker3:
Add a series of utilities to make working with native collections easier and more performant.

#### Tag Alongs
 - `DeferredNativeArray` - Improve wording of documentation

### What is the current behaviour?

The default public API on many native collections is sometimes restricted to calls that incur unnecessary allocations verbose instructions to work with.

### What is the new behaviour?

`NativeArrayInternalExtension` - Methods that leverage `internal` members to
  - Read the allocator of a `NativeArray`

`NativeParallelHashMapInternalExtension` - Methods that leverage `internal` members to
  - Read the allocator of a `NativeArray`
  - Populate an existing `NativeArray` with the the Keys, Values, or KeyValue arrays without allocating a new array

`NativeArrayExtension` - Migrated business logic into `UnsafeCollectionUtil` so any unsafe collection can leverage the operations.

`NativeParallelHashMapExtension` - Add an override to `Remove()` that writes the removed value to an out parameter. It's common to want the value at a key while also removing the key.

`NativeParallelHashSetExtension` - Add a CopyFrom method to bulk copy values from an IEnumerable with optimized paths for copying from another `NativeParallelHashSet`

`FloodSetComponentJob` - Efficiently set all components in a query with one write per chunk.

### What issues does this resolve?
 - None

### What PRs does this depend on?
 - None

### Does this introduce a breaking change?
 - [ ] Yes 
 - [x] No

PR #127, opened 2022-12-16
AccessController - Add GetDependencyFor

mbaker3:
Provides a way to peek at the job handle that would be required for different acquisitions.

This has been useful for
 - Asserting that an access controlled value has already synced to the main thread before making an `Acquire` call.
    - Ensuring assumed execution context
 - Detecting whether an access controlled value has potentially changed.

There is no way to see what the `JobHandle` would be if an acquire request was made to an `AccessController`

Developers can call `GetDependencyFor(AccessType)` to get the job handle for a given type of access.

 - None

 - #126 - Not really but it's not worth rebasing back onto main

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Job/AccessControl/AccessController.cs
+++ b/Scripts/Runtime/Job/AccessControl/AccessController.cs
@@ -186,6 +186,8 @@ namespace Anvil.Unity.DOTS.Jobs
                 default:
                     throw new ArgumentOutOfRangeException(nameof(accessType), accessType, $"Tried to acquire with {nameof(AccessType)} of {accessType} but no code path satisfies!");
             }
+
+            //TODO: #129 - Remove once we have unit tests.
             Debug.Assert(acquiredHandle.Equals(GetDependencyFor(accessType)));
 
             m_LastHandleAcquired = acquiredHandle;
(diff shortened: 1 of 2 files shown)


Scripts/Runtime/Job/AccessControl/AccessController.cs
```diff
@@ -157,6 +186,7 @@ public JobHandle AcquireAsync(AccessType accessType)
                 default:
                     throw new ArgumentOutOfRangeException(nameof(accessType), accessType, $"Tried to acquire with {nameof(AccessType)} of {accessType} but no code path satisfies!");
             }
+            Debug.Assert(acquiredHandle.Equals(GetDependencyFor(accessType)));
```
jkeon, line 189: Minor but I feel like this is better used as a Conditional method check under `ENABLE_UNITY_COLLECTIONS_CHECKS`. Seems too heavy for a Debug Assert.

jkeon (approved):
Looks good, one comment to take or leave.

Scripts/Runtime/Job/AccessControl/AccessController.cs
mbaker3: Going to leave it for now. Discussed a bit and
 - Really this should get handled by a Unit test and removed. #129 
 - This is a cheap check
 - Collections checks tend to be correct usage type checks/assertions while this check is more of a "is our logic consistent" check.

I did add a TODO reference comment for the assert.

jkeon (approved):

PR #130, opened 2022-12-19
AccessController - Add last access stack to errors

mbaker3:
Add the stack of the last access operation to lifecycle error messages.

When debugging a forgotten `Release/Async()` call or multiple `Acquire/Async()` calls in a stack it's difficult to track down the other method(s) that caused the contention or unexpected state.

When `ENABLE_UNITY_COLLECTIONS_CHECKS` is present `AccessController` will record the stack at the end of each `Acquire/Async()` and `Release/Async()` call. When an `InvalidOperationException` is thrown it will include the stack of the previous call. This is usually the hint needed to figure out why the `AccessController` was in an unexpected state.

 - None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Job/AccessControl/AccessController.cs
+++ b/Scripts/Runtime/Job/AccessControl/AccessController.cs
@@ -56,6 +56,11 @@ namespace Anvil.Unity.DOTS.Jobs
 
         private AcquisitionState m_State;
 
+#if ENABLE_UNITY_COLLECTIONS_CHECKS
+        private StackTrace m_LastAccessOperationStack;
+#endif
+
+
         public AccessController()
         {
         }
@@ -210,6 +215,7 @@ namespace Anvil.Unity.DOTS.Jobs
             Debug.Assert(acquiredHandle.Equals(GetDependencyFor(accessType)));
 
             m_LastHandleAcquired = acquiredHandle;
+            CaptureAccessOperationStack();
 
             return acquiredHandle;
         }
@@ -269,6 +275,7 @@ namespace Anvil.Unity.DOTS.Jobs
                     throw new ArgumentOutOfRangeException(nameof(m_State), m_State, $"Tried to release but {nameof(m_State)} was {m_State} and no code path satisfies!");
             }
 
+            CaptureAccessOperationStack();
             m_State = AcquisitionState.Unacquired;
         }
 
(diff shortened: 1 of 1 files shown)


Scripts/Runtime/Job/AccessControl/AccessController.cs
```diff
@@ -56,6 +56,11 @@ private enum AcquisitionState
 
         private AcquisitionState m_State;
 
+#if ENABLE_UNITY_COLLECTIONS_CHECKS
```
jkeon, line 59:
```suggestion
#if ANVIL_DEBUG_SAFETY_EXPENSIVE
```

Scripts/Runtime/Job/AccessControl/AccessController.cs
```diff
@@ -307,6 +320,24 @@ private void ValidateReleaseState(JobHandle releaseAccessDependency)
             }
         }
 
+        [Conditional("ENABLE_UNITY_COLLECTIONS_CHECKS")]
+        private void CaptureAccessOperationStack()
+        {
+#if ENABLE_UNITY_COLLECTIONS_CHECKS
```
jkeon, line 326:
```suggestion
#if ANVIL_DEBUG_SAFETY_EXPENSIVE
```

Scripts/Runtime/Job/AccessControl/AccessController.cs
```diff
@@ -307,6 +320,24 @@ private void ValidateReleaseState(JobHandle releaseAccessDependency)
...
+            m_LastAccessOperationStack = new StackTrace(1, true);
+#endif
+        }
+
+        private string GetLastAccessOperationStack()
+        {
+#if ENABLE_UNITY_COLLECTIONS_CHECKS
```
jkeon, line 333:
```suggestion
#if ANVIL_DEBUG_SAFETY_EXPENSIVE
```

jkeon (changes requested):
Looks good but I would change the flag from COLLECTIONS_CHECKS to our more expensive debug check ANVIL_DEBUG_SAFETY_EXPENSIVE

mbaker3:
Converted all of `AccessController` over to use the anvil defines and updated messages.
Now you'll get your safety checks only if `ANVIL_DEBUG_SAFETY` is defined. If an error is encountered the log message will direct the developer to enable `ANVIL_DEBUG_SAFETY_EXPENSIVE` to get the stack trace info.

jkeon (approved):

PR #132, opened 2022-12-20
TransformUtil - Add AddMissingStandardComponents

mbaker3:
Add a utility method to add a standard set of transform components to an `Entity` if those components are missing.

I'm open to changing the name of the method if you have a better idea.

Unity's game object conversion system doesn't add transform components during conversion if those components are default values (Ex: Scale 1,1,1).

This can cause issues if there are systems that expect to modify or query these values at runtime.

`TransformUtil.AddMissingStandardComponents` analyzes the components on an entity and adds any missing components with default values to ensure there are at least basic components for rotation, translation and uniform scale.

Although it is slightly more efficient to omit the values when they are default values having the components present makes for a much smoother development experience. Cutting out unused components is better suited to an optimization pass for high count archetypes.

This method is also suitable for use at runtime.

 - None

 - None

 - [ ] Yes 
 - [x] No


jkeon (approved):
Looks good!

PR #134, opened 2023-01-11
Make CopyBuffer methods more flexible and available

mbaker3:
Now that we have a better understanding (and toolset) for ECS the `AbstractAnvilSystemBase.CopyFromSingletonBufferAsync` methods and friends no longer make sense.

 - The CopyTo/From SingletonBufferAsync methods are made available to all `ComponentSystemBase` derived types
 - The CopyTo/From SingletonBufferAsync jobs and methods are now able to be used with any individual buffer. Not just Singleton buffers.

#### Tag Along
 - `AbstractAnvilSystemBase` - Replaced  `ENABLE_UNITY_COLLECTIONS_CHECKS` pre-processor flag with `ANVIL_DEBUG_SAFETY` for the `Dependency` continuity safety check.

 - Systems must derive their system from `AbstractAnvilSystemBase` to access convenience methods to asyncronously copy data between Buffers and Native Arrays.
 - Convenience methods are only available if the buffer exists as a singleton.

 - The copy methods have been moved to `ComponentSystemBaseExtension` for broader availability
 - The copy methods now accept a `BufferFromSingleEntity` container and leverage a CDFE rather than iterating chunks and getting the buffer through a type handle. This allows any buffer on an entity to be copied to/from.
 - Jobs have been renamed to reflect new capability
 - Added `CopyBufferToDeferredNativeArray` for scenarios where the length of the buffer is not known ahead of time.

 - None

 - None

 - [x] Yes
 - [ ] No

Any use of the copy to/from methods need to use the new methods. All conversions follow the same pattern. Following the example below should provide guidance for the rest of the methods.
``` csharp
// Old
Dependency = CopyToSingletonBufferAsync(Dependency, myInputBuffer);

// New
Dependency = CopyNativeArrayToBuffer(Dependency, myInputBuffer, GetSingletonEntity<MyElementType>());
```


jkeon (approved):
Looks good

PR #135, opened 2023-01-13
LazyReactiveValue - Imlement

mbaker3:
Add a wrapper type to reduce the boilerplate to setup a lazily updating/re-caching/recomputing a value.

Currently if a developer wants to setup a value that lazily updates when another condition changes they must setup:
 - The value itself
 - A field to track the previous state of the condition
 - The logic to compare the condition and update the value if needed on every get. 

This is a lot of boiler plate and prone to errors.

`LazyReactiveValue` encapsulates this functionality in a single type while providing maximum flexibility.

Contrived Simple Example:
``` csharp
AccessControlledValue<int> myACV = new AccessControlledValue<int>(5);
LazyReactiveValue<uint, JobHandle> myCachedValue = new LazyReactiveValue(
   () => myACV.GetDependencyFor(AccessType.SharedRead),
   (ref uint value) =>
   {
       using var myACVHandle = myACV.AcquireWithHandle(AccessType.SharedRead);
       value = myACVHandle.Value;
   }
);
```

Real World Example:
``` csharp
private readonly LazyReactiveValue<NativeList<uint>, JobHandle> m_Environments;

// In the constructor
m_Environments = new LazyReactiveValue<NativeList<uint>, JobHandle>(
    () => EnvironmentIDToEntity.GetDependencyFor(AccessType.SharedRead),
    (ref NativeList<uint> value) =>
    {
        using var environmentIDToEntityHandle = EnvironmentIDToEntity.AcquireWithHandle(AccessType.SharedRead);
        value.ResizeUninitialized(environmentIDToEntityHandle.Value.Count());
        environmentIDToEntityHandle.Value.GetKeyArray(value);
    },
    new NativeList<uint>(5, Allocator.Persistent)
);
```

 - None

 - None

 - [ ] Yes 
 - [x] No


Scripts/Runtime/Data/LazyReactiveValue.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// A delegate that is expected to modify the <see cref="value"/> parameter in place with up to date value(s).
+        /// </summary>
+        /// <param name="value">A reference to the value to update.</param>
+        public delegate void UpdateValueDelegate(ref TValue value);
+
+        private readonly Func<TTriggerType> m_GetCurrentTriggerValue;
+        private readonly UpdateValueDelegate m_UpdateValue;
+
+        private TValue m_Value;
+        private TTriggerType m_LastDependencyHandle;
```
jkeon, line 39: Minor but since this isn't a handle anymore, rename? 
```suggestion
        private TTriggerType m_LastTrigger;
```

Scripts/Runtime/Data/LazyReactiveValue.cs
```diff
@@ -0,0 +1,96 @@
...
+            base.DisposeSelf();
+        }
+
+        /// <summary>
+        /// Pro-actively update the value. Typically used if it's known that the source trigger has changed and the
+        /// application is in an optimal state to update the value.
+        /// </summary>
+        public void UpdateIfDirty()
+        {
+            TTriggerType currentReadHandle = m_GetCurrentTriggerValue();
```
jkeon, line 85:
```suggestion
            TTriggerType currentTrigger = m_GetCurrentTriggerValue();
```

jkeon (approved):
Looks good, minor naming for better readability for someone who doesn't know the history of this class' development

Scripts/Runtime/Data/LazyReactiveValue.cs
mbaker3: Fixed

Scripts/Runtime/Data/LazyReactiveValue.cs
mbaker3: Fixed

jkeon (approved):

PR #144, opened 2023-01-19
DRAFT - i143 - Persistent Data

### LINKED ISSUE #143: Persistent Data
Add the ability to have persistent data similar to how Entities data exists in an EntityManager. 

This is useful for data that we need to have as lookups, thread specific scratch pads or entity data that doesn't make sense to put on another entity and have a redirect to the proxy entity.

jkeon:
@mbaker3 Looking for opinions and feedback on this before I do a docs pass. I've made github conversations on specific questions I have below but everything is fair game.

I'll use a real world example from the pathfinding prototype. 

Pathfinding requires that we keep track of costs for a node and a list of where a node came from. These are arrays or lookups but they can be small or grow very large depending on how long the path is. 

Pathfinding also requires some state data such as the Grid and whether a cell is walkable or not and what a cells neighbours are. Can we allow diagonal movement or not for example.

Finally, for every Actor that wants to Pathfind, it needs a corresponding buffer to hold what its current path is and where it is on that path. 

Today I can accomplish this by having these Data Systems that purely exist to hold an `AccessControlledValue` that wraps the data I need. 

**ScratchPads Example**
```csharp
//Declare
public readonly AccessControlledValue<NativeArray<UnsafeParallelHashMap<int, float>>> ScratchPad_CostSoFar;

//Long winded instantiation
m_NumThreads = ParallelAccessUtil.CollectionSizeForMaxThreads;
ScratchPad_CostSoFar = new AccessControlledValue<NativeArray<UnsafeParallelHashMap<int, float>>>(new NativeArray<UnsafeParallelHashMap<int, float>>(m_NumThreads, Allocator.Persistent));

NativeArray<UnsafeParallelHashMap<int, float>> costSoFar = ScratchPad_CostSoFar.Acquire(AccessType.ExclusiveWrite);
for (int i = 0; i < m_NumThreads; ++i)
{
                costSoFar[i] = new UnsafeParallelHashMap<int, float>(ChunkUtil.MaxElementsPerChunk<int>(), Allocator.Persistent);
}

ScratchPad_CostSoFar.Release();

//Long winded disposal
NativeArray<UnsafeParallelHashMap<int, float>> costSoFar = ScratchPad_CostSoFar.Acquire(AccessType.Disposal);
for (int i = 0; i < m_NumThreads; ++i)
{
                costSoFar[i].Dispose();
}

ScratchPad_CostSoFar.Dispose();
```

And then in the other `Systems` where i need access to it, I have to do something like this:

```csharp
//Get the system
m_PathfindTaskSystem = world.GetOrCreateSystem<TEMP_PathfindTaskSystem>();
//Then get the data
m_PathfindTaskSystem.ScratchPad_CostSoFar.Acquire(....)
```

Data in Unity's ECS system is globally accessible via `EntityManager`, `CDFE`, `DynamicBuffers` and `EntityQueries`. You can write Systems that interact with data without needing to know other systems that also interact or declare it. 

But of course it requires you to opt in fully to the Entities framework and have everything keyed off of an Entity which isn't necessary.

So the concept of `PersistentData` is introduced. This is data that exists globally, can be accessed from anywhere but is still AccessControlled. 

It is all managed/created through a central `PersistentDataManager` allowing you to easily create via:

```csharp
PersistentDataManager.CreatePersistentData(id, ...);
```
and get:
```csharp
PersistentDataManager.GetPersistentData(id);
```

Today there are three types. 

**ThreadPersistentData**

This is data that gets replicated for every thread that could be run at the same time. It is very useful for things like scratch pads for temp calculations like in the Pathfinding example. We don't have to allocate new collections each time the job runs, and the size of the collections grows as needed and stays that way to handle the loads it is receiving.

Creation is now easier and you get a callback to create the piece of data you want stored for each thread and a callback for disposing as well if needed.

```csharp
PersistentDataManager.CreateThreadPersistentData(SCRATCH_PAD_COST_SO_FAR,
                                                             threadIndex => new UnsafeParallelHashMap<int, float>(ChunkUtil.MaxElementsPerChunk<int>(), Allocator.Persistent),
                                                             ((threadIndex, data) =>
                                                              {
                                                                  data.Dispose();
                                                              }));
```

and usage in a job:

```csharp
private ThreadPersistentDataAccessor<UnsafeParallelHashMap<int, float>> m_CostSoFar;
m_CostSoFar.ThreadData[initialIndex] = 0.0f;
```

**EntityPersistentData**

This is data that is tied to an Entity but it is not on the Entity itself nor is it on a proxy Entity. 

Take an Actor that might have a structure like so:

- Actor (Entity)
    - Health
    - Speed
    - Etc

The more components we add, the larger the Entity and the less of them we can store in a Chunk. If we have some state specific data we have a few ways to handle it. 

We could just add it to the Actor itself

- Actor (Entity)
    - Health
    - Speed
    - PathBuffer
    - PathProgress
    - Etc

But that is wasteful for something we won't be using all the time.

We could add it to proxy entity

- Actor (Entity)
    - Health
    - Speed
    - PathEntity
    - Etc

- PathEntity (Entity)
    - PathBuffer
    - PathProgress

But this requires us to do an additional redirect to get the data we need.
`Actor -> PathEntity -> PathBuffer`

Or we can use `EntityPersistentData` 

- Actor (Entity)
    - Health
    - Speed
    - Etc

- EntityPersistentData 
    - Actor (Entity)
        - PathBuffer
        - PathProgress

This lets us have our small Actor Entity, and use that single Entity as the key to get our state data for Pathfinding.

Similar creation looks like this:

```csharp
PersistentDataManager.CreateEntityPersistentData<UnsafeList<int2>>(PATH_BUFFERS_BY_ENTITY,
                                                                               ((entity, data) =>
                                                                                {
                                                                                    data.Dispose();
                                                                                }));
``` 

And usage in a job:

```csharp
private EntityPersistentDataWriter<UnsafeList<int2>> m_PathBuffers;
UnsafeList<int2> pathBuffer = m_PathBuffers[instance.Entity];
pathBuffer.Add(current);
m_PathBuffers[instance.Entity] = pathBuffer;
```

**PersistentData**

Finally, this is just generic data that you want to have access to in your job. A lookup struct, loot table, etc etc. You need it in your job, you don't want to have a lone Entity that has it as the singular component resulting in using a second query to get it and convert to a NativeArray of 1 element.

Creation:

```csharp
PersistentDataManager.CreatePersistentData(GRID, new PathfindingGrid(80, 40));
```

Usage in job:

```csharp
int initialIndex = m_PathfindingGrid.Data.GetGridIndex(start);
```

- Resolves #143 

 - #142 

 - [ ] Yes
 - [x] No

### Tag Alongs

Adding in an `UnsafeArray` and corresponding helpers plus easy conversion to/from `NativeArray`.

Why? 

Because there is no `UnsafeArray` today, only `NativeArray` which has a safety handle and that allows it to be `struct` but not `unmanaged` and unfortunately we need things to be `unmanaged` because of how pointers work. 

This is similar to why `NativeArray` is not allowed on an `IComponentData` but `UnsafeList` is perfectly fine, although they recommend you use a `DynamicBuffer` instead because they made it safety but `Entity` compatible. The safety handle breaks the unmanaged aspect.

I just took all the `NativeArray` code, duplicated and removed the Safety Handle stuff. Everything else is the same, weird code and all. 

We COULD use an `UnsafeList` instead, but it just requires more thought in setup and usage.


Scripts/Runtime/Data/Collections/UnsafeArray.cs
```diff
@@ -0,0 +1,606 @@
...
+using Unity.Jobs;
+using UnityEngine.Internal;
+
+namespace Anvil.Unity.DOTS.Data
+{
+    [NativeContainerSupportsDeferredConvertListToArray]
+    [DebuggerDisplay("Length = {Length}")]
+    [NativeContainerSupportsDeallocateOnJobCompletion]
+    [NativeContainerSupportsMinMaxWriteRestriction]
+    public struct UnsafeArray<T> : IDisposable,
```
jkeon, line 18: @mbaker3 

1. Should we use this vs `UnsafeList` and just remember to be careful? 
2. If we keep this, should we clean it up or keep it the same as Unity's `NativeArray`?

Scripts/Runtime/Data/Collections/Util/UnsafeArrayUtility.cs
```diff
@@ -0,0 +1,63 @@
...
+                                                                                unsafeArray.m_AllocatorLabel);
+        }
+
+        public static unsafe UnsafeArray<T> AsUnsafeArray<T>(this NativeArray<T> nativeArray)
+            where T : unmanaged
+        {
+            return ConvertExistingDataToUnsafeArray<T>(NativeArrayUnsafeUtility.GetUnsafeBufferPointerWithoutChecks(nativeArray),
+                                                       nativeArray.Length,
+                                                       nativeArray.GetAllocator());
+        }
```
jkeon, line 51: If we do end up scrapping the `UnsafeArray` in favour of `UnsafeList`, these are probably still good to have to convert between `UnsafeList` and `NativeArray` quickly.

Scripts/Runtime/Entities/PersistentData/PersistentDataManager.cs
```diff
@@ -0,0 +1,116 @@
+using Anvil.CSharp.Collections;
+using Anvil.Unity.DOTS.Jobs;
+using System;
+using System.Collections.Generic;
+using UnityEngine;
+
+namespace Anvil.Unity.DOTS.Entities
+{
+    public static class PersistentDataManager
```
jkeon, line 9: Should this be a static class? 

Its nice and easy to access, but then you also have to dispose it when stopping the game and going to a menu for example. 

If this was an object, I'm curious where to place it.

In my mind there are the following use cases:

- ThreadPersistentData 
    - This is global to the App. It shouldn't matter what world we're in because you can only access one thread at a time.
- EntityPersistentData 
    - This, I think, should be global to the World no? As it's written today, if you can have `Entity(3 : 2)` in WorldA and `Entity(3: 2)` in WorldB and they are different Entities then I think this could conflict.
- PersistentData 
    - This I also think should global to the World? But then we might also want one that it global to the app.

So maybe there needs to be a "Scope" value as well. (App or World)

In terms of placement, it could go on it's own System and we get the benefit of it being destroyed when the System gets destroyed. App scope is kind of messed up then but we could have the World Scope managers lazy get or create one singular App scope manager and redirect to that. When all World managers are destroyed, then the App one destroys itself too.

I don't like having to get the system though, but maybe we could do that by default in our base system. So the API then becomes:

```csharp
//PersistentData is World scoped to this System we are writing this line of code in. (Or TaskDriver)
PersistentData.CreatePersistentData(id, scope, etc)
```

Scripts/Runtime/Entities/PersistentData/PersistentDataManager.cs
```diff
@@ -0,0 +1,116 @@
...
+            return threadPersistentData;
+        }
+        
+        public static IPersistentData<T> AcquirePersistentData<T>(string id, AccessType accessType)
+            where T : unmanaged
+        {
+            PersistentData<T> persistentData = GetPersistentData<T>(id);
+            persistentData.Acquire(accessType);
+            return persistentData;
+        }
```
jkeon, line 80: These allow you to do main thread work for setup

Scripts/Runtime/Entities/PersistentData/PersistentDataManager.cs
```diff
@@ -0,0 +1,116 @@
...
+        internal static PersistentData<T> GetPersistentData<T>(string id)
+            where T : unmanaged
+        {
+            if (!PERSISTENT_DATA_LOOKUP.TryGetValue(id, out AbstractPersistentData data))
+            {
+                throw new InvalidOperationException($"Trying to get {nameof(PersistentData<T>)} with {nameof(id)} of {id} but it doesn't exist in the lookup! Did you call {nameof(CreatePersistentData)}?");
+            }
+
+            return (PersistentData<T>)data;
+        }
```
jkeon, line 114: You don't call these directly, instead the job configs get them, you just call:

```csharp
ConfigureSystemJobToUpdate(RequestDataStream,
                                       PathfindJob.Schedule,
                                       BatchStrategy.MaximizeThreads)
               .RequireResolveTarget<PathfindSuccess>()
               .RequireResolveTarget<PathfindFailureNoValidPath>()
               .RequireThreadPersistentDataForWrite<UnsafeParallelHashMap<int, float>>(SCRATCH_PAD_COST_SO_FAR)
               .RequireThreadPersistentDataForWrite<UnsafeParallelHashMap<int, int2>>(SCRATCH_PAD_CAME_FROM)
               .RequireEntityPersistentDataForWrite<UnsafeList<int2>>(PATH_BUFFERS_BY_ENTITY)
               .RequirePersistentDataForRead<PathfindingGrid>(GRID)
               .RequirePersistentDataForWrite<UnsafeArray<int2>>(DIRECTIONS);
```

Scripts/Runtime/Entities/PersistentData/PersistentDataManager.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            Dispose();
+        }
+
+        public static void Dispose()
+        {
+            PERSISTENT_DATA_LOOKUP.DisposeAllValuesAndClear();
+        }
+
+        public static void CreatePersistentData<T>(string id,
```
jkeon, line 24: ID's as strings... hmmmn. 

With Entities, it's based on `Types` and the amalgamation of them into `Archetypes` plus/or `Entity` as a key for CDFE/DynamicBuffer access.

We could use Types as well, but then it prevents us from having more than one persistent data of the same type. 

For example we have a PathBuffer one for a `UnsafeList<int2>`, I'm pretty sure we'll need `UnsafeList<int2>` for another one that isn't related to PathBuffers. You could get around this by wrapping the `UnsafeList<int2>` into a struct though but that's annoying to create that type just to ensure duplicate inner type usage.

So strings as ID's made the most sense for now. Thoughts?

mbaker3:
I haven't read your code or looked at your inline comments yet. We should probably talk about the problems this is intending to solve. I'm not sure I totally understand the issue yet any it's possible these are solved with some approaches I've used with the built in ECS functionality.

Some comments below but this is probably a discussion on a call.

---

> ThreadPersistentData
 I think I see the value in this.

Do instances of TPD get shared between Worlds?
  - I imagine yes since these are generally, per thread, scratch pads so the are irrelevant to different consumers of the data. Each would just thread it as uninitialized data when starting.

---

> EntityPersistentData
I don't currently see the benefit of this over a proxy entity. At best it looks equal in performance and functionality with additional maintenance and usage overhead for the developer.

Your core actor entity could have a single component`StateDataReference` that holds one entity reference that you look up on. Every state writes their proxy entity to this.

> But this requires us to do an additional redirect to get the data we need.
Actor -> PathEntity -> PathBuffer

There's an indirect lookup with persistent data anyway. You still need to get the actor entity and then go fetch your memory on that entity.

With both solutions, if you don't need values from the core actor then you can efficiently iterate through just the proxy or permanent data. 

If a state has different variants (Ex: vehicle pathfinding vs crew pathfinding) that have common but some different pieces of state information. The proxy entity system handles that without any additional work from the developer. For persistent data you'd have to maintain separate collections and decide how you want to handle the layout of the data. You don't get a unified query/iteration mechanism for routines that only operate on common data unless we build that.

With persistent data it's a new concept/system and another special case we have to support for world transitions and serialization.

> PersistentData
I'm not sure I see the advantage here that `ManagedReference/Singleton` or AccessControledValue on systems doesn't solve.

Earlier you mentioned that the data generally has to be put on systems or keyed of an entity. There are two different reasons why you'd use one vs the other but it more or less boils down contextualizing the reference to a World. You have to key the data on a World or some other world unique ID so why not use what's already in front of us?

Systems -> Good for storing world state type information. Content comes and goes from the world and the system adapts.
Entity Bound(Ex: The SingletonEntity) -> Good for storing content specific information. This survives moves between worlds and doesn't need to be continually regenerated.

> you don't want to have a lone Entity that has it as the singular component 

With our singleton management that isn't the case. There are very few entities that hold all managed references. You gain access to them through singleton queries that dictate read/write requirements and you're off to the races. The managed references are restricted to a couple of "wasted" chunks.

What's the value of defining our own ID, access and data storage systems vs accomplishing the same but with the benefit of compatibility with the existing API?

jkeon:
Yes lets have a call first thing on Monday then.

---

### ThreadPersistentData

> I think I see the value in this.
> 
> Do instances of TPD get shared between Worlds?

Yeah the value is that the scope is App wide and per thread. So all Worlds with a Pathfinding system for example can use the same already allocated memory for a scratch pad to do the temp calculations needed and we don't need to re-allocate and potentially grow the collection size each time a job of that type is run.

### EntityPersistentData

> I don't currently see the benefit of this over a proxy entity.

Let's chat about it, I disagree but maybe we're not thinking of each others use cases.

### PersistentData

> I'm not sure I see the advantage here that ManagedReference/Singleton or AccessControledValue on systems doesn't solve.

Yep, lets chat here too.

mbaker3:
Chatted on a call. We're going to move forward with just using **ThreadPersistentData** and seeing if there is a use for the other two at all. For the next week or so this branch will remain as is until @jkeon has explored a couple of implementations in an application to see if he runs into a case where these solutions are clearly better than what we can do with Entities in ECS.

mbaker3:
Unsubscribing until you're ready to have this reviewed again. Ping me directly when ready.

jkeon:
Closing as we won't merge this PR. Will be referenced in a future PR.

PR #149, opened 2023-01-24
Job Config Changes

jkeon:
Very simple change to make it so that you can call `RunOnce` after configuration. 
This is useful for situations where you want to run a job after being notified of structural changes.
Adjusted docs accordingly.

 - None

 - None

 - [ ] Yes
 - [x] No


Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/JobData/AbstractJobData.cs
```diff
@@ -1,3 +1,4 @@
+using Anvil.CSharp.Logging;
```
mbaker3, line 1: Extra?

mbaker3 (approved):
Approved but it looks like there's an unintended import.

Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/JobData/AbstractJobData.cs
jkeon: Fixed

PR #150, opened 2023-01-24
i147 Porting Native Heap

### LINKED ISSUE #147: Pathfinding - Port NativeHeap to UnsafeHeap
We have a `NativeHeap` from https://github.com/Amarcolina/NativeHeap that we were using in our prototypes.

Longer term we should reimplement `PriorityQueue` from https://github.com/BlueRaja/High-Speed-Priority-Queue-for-C-Sharp as Burst compatible but that's a higher effort and not necessary now.

For now, we want an UnsafeHeap so that we can store this in `ThreadPersistentData` and not have to create a new one for every Pathfinding job that occurs.

Copy/Paste, remove any Safety handles so that it's `unmanaged`

jkeon:
Porting Native Heap to Unsafe Heap

NativeHeap has a safety info in it which makes it not possible to nest in another native collection.

UnsafeHeap has no safety info in it so it is `unmanaged` and can be put inside another native collection.

- Resolves #147 

 - #149 

 - [ ] Yes
 - [x] No

### DIFF
--- a/Scripts/Runtime/Data/Collections/UnsafeHeap.cs
+++ b/Scripts/Runtime/Data/Collections/UnsafeHeap.cs
@@ -11,14 +11,14 @@ namespace Anvil.Unity.DOTS.Data
     //NOTE - Taken from: https://github.com/Amarcolina/NativeHeap (MIT LICENCE)
     //This is a lightly cleaned up version of NativeHeap from Amarcolina that removes
     //any safety aspects so that it can be used inside a different Native Collection.
-    
+
     /// <summary>
     /// This is a basic implementation of the MinHeap/MaxHeap data structure.  It allows you
     /// to insert objects into the container with a O(log(n)) cost per item, and it allows you
     /// to extract the min/max from the container with a O(log(n)) cost per item.
     /// 
     /// This implementation provides the ability to remove items from the middle of the container
-    /// as well.  This is a critical operation when implementing algorithms like astar.  When an
+    /// as well.  This is a critical operation when implementing algorithms like A*.  When an
     /// item is added to the container, an index is returned which can be used to later remove
     /// the item no matter where it is in the heap, for the same cost of removing it if it was
     /// popped normally.
@@ -32,6 +32,11 @@ namespace Anvil.Unity.DOTS.Data
     /// For convenience, this library contains the Min and Max comparer, which provide
     /// comparisons for all built in primitives.
     /// </summary>
+    /// <remarks>
+    /// This collection doesn't have the usual safety checks that other collections in Anvil have
+    /// because we copy/pasted and "it just works(TM)".
+    /// TODO: #151 - Implementing PriorityQueue will supersede this class, so not worth adding safety checks
+    /// </remarks>
     [DebuggerDisplay("Count = {Count}")]
     [StructLayout(LayoutKind.Sequential)]
     public struct UnsafeHeap<T, TComparer> : IDisposable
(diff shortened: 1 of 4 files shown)


Scripts/Runtime/Data/Collections/UnsafeHeap.cs
```diff
@@ -0,0 +1,610 @@
...
+    //This is a lightly cleaned up version of NativeHeap from Amarcolina that removes
+    //any safety aspects so that it can be used inside a different Native Collection.
+    
+    /// <summary>
+    /// This is a basic implementation of the MinHeap/MaxHeap data structure.  It allows you
+    /// to insert objects into the container with a O(log(n)) cost per item, and it allows you
+    /// to extract the min/max from the container with a O(log(n)) cost per item.
+    /// 
+    /// This implementation provides the ability to remove items from the middle of the container
+    /// as well.  This is a critical operation when implementing algorithms like astar.  When an
```
mbaker3, line 21:
```suggestion
    /// as well.  This is a critical operation when implementing algorithms like A-Star.  When an
```

or A*
Up to you
https://en.wikipedia.org/wiki/A*_search_algorithm

Scripts/Runtime/Data/Collections/UnsafeHeap.cs
```diff
@@ -0,0 +1,610 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using Unity.Collections;
+using Unity.Collections.LowLevel.Unsafe;
+
+namespace Anvil.Unity.DOTS.Data
+{
+    //NOTE - Taken from: https://github.com/Amarcolina/NativeHeap (MIT LICENCE)
```
mbaker3, line 11: If they already have unit tests we should steal those too.

mbaker3 (commented):
Some additional work but overall looks good.

Scripts/Runtime/Data/Collections/UnsafeHeap.cs
jkeon: I didn't write this, Amarcolina did, but it was a quick change so done.

Scripts/Runtime/Data/Collections/UnsafeHeap.cs
```diff
@@ -0,0 +1,610 @@
...
+    /// 
+    /// This container is parameterized with a comparer type that defines the ordering of the
+    /// container.  The default form of the comparer can be used, or you can specify your own.
+    /// The item that comes first in the ordering is the one that will be returned by the Pop
+    /// operation.  This allows you to use the comparer to parameterize this collection into a 
+    /// MinHeap, MaxHeap, or other type of ordered heap using your own custom type.
+    /// 
+    /// For convenience, this library contains the Min and Max comparer, which provide
+    /// comparisons for all built in primitives.
+    /// </summary>
```
mbaker3, line 34: Based on discussion can we add a `<remarks>` section that mentions this type does not have the safety that other unsafe types have in anvil?

Also a TODO mentioning that this will be replaced by the work in (some issue number) so fixing the above remark isn't worthwhile.

Scripts/Runtime/Data/Collections/UnsafeHeap.cs
jkeon: Done

Scripts/Runtime/Data/Collections/UnsafeHeap.cs
jkeon: Done

PR #153, opened 2023-01-30
Outside Interaction with Task Driver System

jkeon:
Adding the ability to interact with the TaskDriver system easily from the outside. 
- Main Thread
- Other jobs

You need to create jobs on the TaskDriver to populate the DataStreams by triggering off an `EntityQuery` or `NativeArray`. 

This can lead to indirection where you have a normal Entities job that exists solely to populate a trigger array to then get the data inside the TaskDriver.

You can now directly get access to:
- `DataStreamActiveReader<TInstance>`
- `DataStreamPendingWriter<TInstance>`
- `CancelRequestsWriter`
- `CancelCompleteReader`

And use those in your jobs that are outside the Task Driver system. Access is handled for you but you must explicitly call `Release` where appropriate depending if you are accessing synchronously or on the main thread.

- None - Just a bunch of annoyances @mbaker3 and @jkeon ran into during real world setup.

 - None

 - [ ] Yes
 - [x] No

### DIFF
--- a/Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/DataSource/Data/AbstractData.cs
+++ b/Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/DataSource/Data/AbstractData.cs
@@ -55,5 +55,17 @@ namespace Anvil.Unity.DOTS.Entities.TaskDriver
         {
             m_AccessController.ReleaseAsync(releaseAccessDependency);
         }
+        
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void Acquire(AccessType accessType)
+        { 
+            m_AccessController.Acquire(accessType);
+        }
+        
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void Release()
+        {
+            m_AccessController.Release();
+        }
     }
 }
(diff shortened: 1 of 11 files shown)


Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/DataSource/Data/AbstractData.cs
```diff
@@ -55,5 +55,17 @@ public void ReleaseAsync(JobHandle releaseAccessDependency)
         {
             m_AccessController.ReleaseAsync(releaseAccessDependency);
         }
+        
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
mbaker3, line 59: Not really on topic for this PR but does this actually work since the method captures a private class member?

mbaker3 (approved):

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/DataSource/Data/AbstractData.cs
jkeon: Not sure, made an issue to investigate: #154

PR #155, opened 2023-01-30
Singular Scheduling

jkeon:
Allow for Defer'd job types to be scheduled on one thread instead of only parallel

- Can only schedule in parallel right now.

- Can choose to schedule on one thread or in parallel
- Also renamed `UpdateDataStreamScheduleInfo` and `CancelDataStreamScheduleInfo` to `UpdateScheduleInfo` and `CancelScheduleInfo` respectively.

- None

 - None

 - [ ] Yes
 - [x] No


mbaker3 (approved):
Looks good. At some point we should document these methods but it's not necessary today.

PR #156, opened 2023-01-31
Adding AccessControllers for CDFE and DBFE Access Wrappers

jkeon:

Multiple jobs in the same system that access the same CDFE or DBFE will clobber each other.

A World based lookup for AccessControllers exists for CDFE's and DBFE's keyed on the type. These AccessControllers will merge in with the System Dependency enabling multiple jobs in the same system to schedule properly and not have conflicts.

- None

 - None

 - [ ] Yes
 - [x] No

### DIFF
--- a/Scripts/Runtime/Entities/TaskDriver/TaskDriverManagementSystem.cs
+++ b/Scripts/Runtime/Entities/TaskDriver/TaskDriverManagementSystem.cs
@@ -1,5 +1,6 @@
 using Anvil.CSharp.Collections;
 using Anvil.CSharp.Data;
+using Anvil.Unity.DOTS.Jobs;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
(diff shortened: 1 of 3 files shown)


Scripts/Runtime/Entities/TaskDriver/TaskDriverManagementSystem.cs
```diff
@@ -166,6 +170,29 @@ public void RegisterTaskDriver(AbstractTaskDriver taskDriver)
...
+
+        public AccessController GetOrCreateDBFEAccessController<T>()
+            where T : struct, IBufferElementData
+        {
+            return GetOrCreateUnityEntityDataAccessController(typeof(T));
+        }
+        
+        private AccessController GetOrCreateUnityEntityDataAccessController(Type type)
+        {
+            if (!m_UnityEntityDataAccessControllers.TryGetValue(type, out AccessController accessController))
```
mbaker3, line 187: We do this so often I'm starting to think we should just create a `LazyDictionary` that auto-instantiates values for a requested key.

mbaker3 (approved):

Scripts/Runtime/Entities/TaskDriver/TaskDriverManagementSystem.cs
jkeon: It's not a bad idea, it just gets hard when the element doesn't have a parameterless constructor because then you can't generically create it.

PR #160, opened 2023-02-06
Access Wrapper Fix

jkeon:
Fixing an Acquire/Release error when the same job requires writing to two or more DataStreams of the same type.

AccessWrappers are created for the type and given the context of the instance that owns them.
When a job is scheduled, we loop through all the wrappers and `AcquireAsync` to ensure we have access to them.
We then schedule the job and then loop through all the wrappers again and `ReleaseAsync` to complete the cycle.

Because DataStreams share a DataSource all Write operations operate on the same underlying collection regardless of the instance that will use it. 

Ex. TaskDriverA with DataStream<TypeA> and TaskDriverB with DataStream<TypeA> will both write to a DataSource<TypeA>. Upon that DataSource's consolidation, the active instances will be written to the correct TaskDriver's instance of DataStream.

This results in two acquires in a row to the same underlying AccessController.

When hardening the JobConfig, we only add the unique wrappers.

- None

 - None

 - [ ] Yes
 - [x] No

### DIFF
--- a/Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/JobConfig/AbstractJobConfig.cs
+++ b/Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/JobConfig/AbstractJobConfig.cs
@@ -21,12 +21,16 @@ namespace Anvil.Unity.DOTS.Entities.TaskDriver
         {
             //Using in the general context (read or write)
             Default,
+
             //Using in an Updating context for data that can we resolved
             Update,
+
             //Using in the context where it can be written to via a resolve
             Resolve,
+
             //Using in the context for requesting a cancellation
             RequestCancel,
+
             //Using in the context for doing the work to cancel 
             Cancelling
         }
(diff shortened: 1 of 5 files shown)


Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/JobConfig/AbstractJobConfig.cs
```diff
@@ -284,6 +291,13 @@ public void Harden()
 
             foreach (AbstractAccessWrapper wrapper in m_AccessWrappers.Values)
             {
+                //Only allow one wrapper per type for DataStream Pending Access since they will all try to acquire/release
+                //the same DataSource instance.
+                if (wrapper is IDataStreamPendingAccessWrapper pendingAccessWrapper
```
mbaker3, line 296: You don't seem to use `pendingAccessWrapper` did you mean to?

mbaker3 (commented):
In addition to my one comment. Do these config objects live beyond the hardening stage?

If so, is the HashSet required after hardening? Seems a bit wasteful to keep it around if it's not used afterwards. Not saying we fix now but just plan to some day as part of an optimization or hardening cleanup pass.

Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/JobConfig/AbstractJobConfig.cs
jkeon: Good catch, turns out I didn't need it, removed.

jkeon:
> In addition to my one comment. Do these config objects live beyond the hardening stage?
> 
> If so, is the HashSet required after hardening? Seems a bit wasteful to keep it around if it's not used afterwards. Not saying we fix now but just plan to some day as part of an optimization or hardening cleanup pass.

The config objects live on, but your right, that HashSet isn't needed. Removed.

mbaker3 (approved):

PR #165, opened 2023-02-16
Hotfix - Out of order System Creation

jkeon:
Fixed a bug in the auto system group creation in case the group wasn't created yet.

With Unity having two paths for system creation, the default generation path creates the systems (and system groups) but doesn't call `OnCreate` for them yet. 

If your system gets "created" first, the group it wants to be a part of may not yet be created and therefore we can't add it to the update list for that group. 

We also can't just force an `OnCreate` for the group because with the available API's that will just add a duplicate system group to it's system group. 

We can detect if we're in this situation and then defer adding to the Update list until our required group is created. 

- None

 - None

 - [ ] Yes 
 - [x] No


mbaker3 (approved):
Nice fix but what a shitty situation Unity put us in.

jkeon:
Agreed!

PR #166, opened 2023-02-17
Exposing internal groups for multiworld optimizations

jkeon:

The internal groups that get created for multi-world optimizations are not accessible for apps to take advantage of.

Now they are and renamed so that they can be used for anything.

- None

 - None

 - [ ] Yes 
 - [x] No


mbaker3 (changes requested):
Looks good just need to update the doc comments on the groups since their purpose is no longer strictly for ECBs. We can mention that ECBs are moved there but maybe just point users to the `OptimizeWorldForMultiWorldInCurrentPlayerLoop` method for more info

Also add mention that a world needs to be multi-world optimized with `OptimizeWorldForMultiWorldInCurrentPlayerLoop` for these to exist.

mbaker3:
Hold on any work for this. Discussion in another PR may invalidate this work.

jkeon:
Holding and in your court.

Issue #28, opened 2022-05-12
Use ExpressionTrees to build a faster delegate in `PlayerLoopUtil`


jkeon:
We can speed the reflected call up by using Expression Trees to create a strongly typed delegate
We should be able to call with the params we have access to and pass in "object" for the param we don't.
In the expression lambda, we do a convert to the internal type defined on the method info.

PR #169, opened 2023-03-01
Anvil PlayerLoop Phases For Everyone

mbaker3:
This is a replacement/alternate approach to what was proposed in #166. If this is accepted/merged then the other PR should be deleted/closed.

This PR exposes the anvil player loop phases/types `PostUpdate_Anvil`/`PostInitialization_Anvil` and adds the ability to safely add `ComponentSystemGroup`s to them.

Included in this work are a number of related tasks:
 - Utility methods for checking for the existence of systems in the player loop.
 - Leveraging assembly injection a system instance may be fetched from a `PlayerLoopSystem` working around the wrapper object Unity has in place for their own fix.
    - A side effect of this change is that reflection is no longer required for any of the `PlayerUtils` functionality making #28 redundant.

#### Tag Alongs
 - Expose the internal events `DefaultWorldInitialization.DefaultWorldInitialized` and `DefaultWorldInitialization.DefaultWorldDestroyed`

 - There is no way for applications to easily add their own `ComponentSystemGroup` instances to the Anvil player loop phases (or any other player loop phase for that matter)
   - Developers also have to be very careful that any groups they do add are only added once and have not already been instantiated.
 - There is no way to check if an instance of a `ComponentSystemBase` has already been added to a player loop

The proposed solution in #166 addressed some of these challenges by hijacking the `*CommandBufferSystemGroup_Anvil` groups but this was not very flexible and undermined the intent of those groups to focus on main thread processing after all other worlds had scheduled their jobs.

Add `ScriptBehaviourUpdateOrderInternal.TryFindPlayerLoopSystemByType` - A utility method to get system from a `PlayerLoopSystem`
    - Uses assembly injection to allow user code to find an instance of a system in a player loop.
    - This is required because Unity wraps all `ComponentSystemBase` instances added via `ScriptBehaviourUpdateOrder.AppendSystemToPlayerLoop` in a dummy type to overcome a limitation with Mono. See comments on the method for details.

Add `ComponentSystemBaseExtension` - Create extension method to try and find the parent group of a system
   - This is a brute force approach. It's not performant but the current use cases aren't performance critical situations.

`PlayerLoopUtil`
 - Remove need for reflection now that we have `ScriptBehaviourUpdateOrderInternal` to do the work
 - Remove group specific methods since they're basically just casting on the system versions of the methods
 - Make naming consistent
    - "PlayerLoop" for recursive evaluations
    - "PlayerLoopSystem" for single level evaluations
 - Add `IsInPlayerLoop` - Checks whether a `ComponentSystemBase` exists anywhere in the `PlayerLoopSystem`'s tree
 - Add `IsInPlayerLoopSystem` - Checks whether a `ComponentSystemBase` exists in a given `PlayerLoopSystem`
 - Add `IsSubsystemOfPlayerLoopSystem` - Checks whether a `ComponentSystemBase` is a subsystem of a `PlayerLoopSystem`
 - Add `TryFindPlayerLoopSystemByType` - A proxy for ScriptBehaviourUpdateOrderInternal.TryFindPlayerLoopSystemByType. This is the method that should be used since the internal version will be removed when Unity fixes their work around.
 - Add `NO_PLAYER_LOOP` to represent no player loop found for the `Try*` methods

`WorldUtil`
 - Make `PostUpdate_Anvil` and `PostInitialization_Anvil` public types with internal constructors.
    - This allows application code to add top level groups to the player loop phases
 - Define a constant for default World groups that unity defines (`DEFAULT_TOP_LEVEL_GROUPS`)
 - Rename `AddTopLevelGroupsToCurrentPlayerLoop` to `SetupTopLevelGroupsInCurrentPlayerLoop`
    -  This reflects the new ability for this method to handle migrating an existing `ComponentSystemGroup` instances out of their existing parent group and into the `PlayerLoop` as a top level group.
    - This method is now more robust and emits a warning if the group already exists as a top layer group
    - This allows more flexible usage by application code.

Expose `DefaultWorldInitialization.DefaultWorldInitialized` and `DefaultWorldInitialization.DefaultWorldDestroyed` via `DefaultWorldInitializationInternal` using Assembly injection.

 - Close: #28 - now redundant
 - Close: #166 - A replacement for the suggested approach

 - None

 - [x] Yes
    - Rename any use of `WorldUtil.AddTopLevelGroupsToCurrentPlayerLoop` to `WorldUtil.SetupTopLevelGroupsInCurrentPlayerLoop`
    - Rename any use of `PlayerLoopUtil.TryGetSystemGroupFromPlayerLoopSystemNoChecks` to `PlayerLoopUtil.TryGetSystemFromPlayerLoopSystem` and perform your own type check to ensure you're getting a group.
 - [ ] No

### DIFF
--- a/Scripts/Runtime/Entities/Util/WorldUtil.cs
+++ b/Scripts/Runtime/Entities/Util/WorldUtil.cs
@@ -29,12 +29,12 @@ namespace Anvil.Unity.DOTS.Entities
     /// A <see cref="PlayerLoop"/> phase inserted immediately after <see cref="Update"/>.
     /// Add via <see cref="WorldUtil.AddCustomPhasesToCurrentPlayerLoop" />.
     /// </summary>
-    internal class PostUpdate_Anvil
+    public class PostUpdate_Anvil
     {
         /// <summary>
         /// Provides a <see cref="PlayerLoopSystem" /> instance to to identify this phase.
         /// </summary>
-        public static PlayerLoopSystem PlayerLoopSystem
+        internal static PlayerLoopSystem PlayerLoopSystem
         {
             get => new PlayerLoopSystem()
             {
@@ -47,12 +47,12 @@ namespace Anvil.Unity.DOTS.Entities
     /// A <see cref="PlayerLoop"/> phase inserted immediately after <see cref="Initialization"/>.
     /// Add via <see cref="WorldUtil.AddCustomPhasesToCurrentPlayerLoop" />.
     /// </summary>
-    internal class PostInitialization_Anvil
+    public class PostInitialization_Anvil
     {
         /// <summary>
         /// Provides a <see cref="PlayerLoopSystem" /> instance to to identify this phase.
         /// </summary>
-        public static PlayerLoopSystem PlayerLoopSystem
+        internal static PlayerLoopSystem PlayerLoopSystem
         {
             get => new PlayerLoopSystem()
             {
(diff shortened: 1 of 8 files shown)


Scripts/Runtime/Entities/Util/WorldUtil.cs
```diff
@@ -122,11 +145,12 @@ public static void AddCustomPhasesToCurrentPlayerLoop()
         /// Groups must not already exist in the world. This is a limitation of our ability to detect whether a system
         /// has already been added to the player loop.
         /// </remarks>
-        public static ComponentSystemGroup[] AddTopLevelGroupsToCurrentPlayerLoop(
+        public static ComponentSystemGroup[] SetupTopLevelGroupsInCurrentPlayerLoop(
             World world,
             (Type PlayerLoopSystemType, Type SystemGroupType)[] topLevelGroupTypes)
         {
             PlayerLoopSystem playerLoop = PlayerLoop.GetCurrentPlayerLoop();
+            bool systemSortRequired = false;
```
jkeon, line 153: Is this being used?

jkeon (approved):
Looks good!

Scripts/Runtime/Entities/Util/WorldUtil.cs
mbaker3: Nope, good catch!

mbaker3:
Feedback addressed. Ready for re-review

jkeon (approved):

PR #175, opened 2023-03-07
FixedListInternalExtension - Contains, IndexOf

mbaker3:
Add implementations of Contains and IndexOf that support Enum elements for all of the `FixedList*Bytes` types

Instances of `FixedList*Bytes<T>` where `T : System.Enum` do not support calling `Contains()` and `IndexOf()`. This is because the provided methods require the elements to implement `IEquatable<T>` and enums do not implement that interface for their own (or underlying) type.

`FixedListInternalExtension` implements enum compatible `Contains()` and `IndexOf()` methods for all `FixedList*Bytes` types that support elements of `System.Emum`.

The developer is required to define the underlying type of the enum on the value parameter. This will most commonly take the form of casting the search value to the underlying type.

**Example:**
``` csharp
FixedList32Bytes<MyEnum> myList = new FixedList32Bytes<MyEnum>();
myList.Contains((int)MyEnum.SomeValue);
```

For safety, there are checks to try to catch situations where the underlying type for an enum has been changed but the method calls haven't been.

 - None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/AssemblyInjections/Unity.Collections/FixedListInternalExtension.cs
+++ b/Scripts/AssemblyInjections/Unity.Collections/FixedListInternalExtension.cs
@@ -1,0 +1,259 @@
+using Anvil.CSharp.Logging;
+using System;
+using System.Diagnostics;
+using Unity.Burst;
+using Unity.Collections;
+using Unity.Collections.LowLevel.Unsafe;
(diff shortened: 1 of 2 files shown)


jkeon (approved):
Nice!

Issue #151, opened 2023-01-24
Implement DOTS version of PriorityQueue


jkeon:
We use a `NativeHeap` from https://github.com/Amarcolina/NativeHeap that we converted to an `UnsafeHeap`.

Longer term we should reimplement PriorityQueue from https://github.com/BlueRaja/High-Speed-Priority-Queue-for-C-Sharp as Burst compatible.

Once done, we should remove the `UnsafeHeap` and associated Tests as it's not as robust at the PriorityQueue.

PR #176, opened 2023-03-09
Better Persistent Data

jkeon:
Reworking of `PersistentData` 

`PersistentData` was built theoretically but hadn't been battle tested in real world projects. With some experience under our belt, we could make some changes.

- PersistentData is now in two flavours. `EntityPersistentData` and `ThreadPersistentData`
- **EntityPersistentData**
   - Three flavours based on scope.
   - You can have one scoped to the TaskDriver itself `IDriverEntityPersistentData` (one per instance of TaskDriver)
   - You can have one scoped to the TaskDriverSystem `ISystemEntityPersistentData` (one per system per world)
   - You can have one scoped to the World `IWorldEntityPersistentData` (one per world)
   - This allows you to select the data based on it's usage and coupling requirements across different systems.
       - Ex. Read from a `IWorldEntityPersistentData` in one system/driver while writing in a completely separate system/driver.
- **ThreadPersistentData**
    - Same as it was before, just better management of it in the now internal `PersistentDataSystem` which acts as a nice way to manage lifecycle for `World` scoped data.
- All data no longer needs IDs! 
- API is similar to creating DataStreams, should feel really familiar.
- Support for reading/writing to the data from outside in regular jobs or main thread access.
- Expects all data to be a `IEntityPersistentDataInstance` or `IThreadPersistentDataInstance` which does mean you end up creating wrapper structs for collections. I feel like this is better for two reasons:
       - 1. The compiler helps out with ensuring the right data is going the right way.
       - 2. We can embrace the pattern of `GetOrCreate`. 
           - Previously, we had to explicitly `Create` and do so before anyone else called `Get`.  
           - Because we passed in the construction/disposal callback, that meant everywhere we called `GetOrCreate` we have to pass in that same construction/disposal logic in case we happen to go down the create path. That's really annoying and not very safe or maintainable.
           - By having the interface we can keep the construction/disposal logic in that struct (where it belongs anyway) and call them during the right phase for you.

### Tag-Alongs

- Added `IsCreated` check to `UnsafeHeap`
- Added the ability to get an `EntityQuery` from a system from the outside when using `AbstractAnvilSystemBase`
    - This is useful for getting a query and having it cached for TaskDriver Systems
    - It's also just useful in general. Unity locks it down by default but DOES expose it via `GetEntityQuery(in EntityQueryDescBuilder builder)`. This just exposes the `ComponentType[]` version.

- #143 

 - None

 - [x] Yes - But no one was using this.
 - [ ] No


Scripts/Runtime/Entities/PersistentData/PersistentDataSystem.cs
```diff
@@ -0,0 +1,68 @@
...
+            {
+                s_ThreadPersistentData.DisposeAllValuesAndClear();
+            }
+            base.OnDestroy();
+        }
+
+        protected override void OnUpdate()
+        {
+            //Just in case
+            Enabled = false;
```
mbaker3, line 40: Why does this system need to participate in updates? Are there queries created against it? Seems like this should just be a data system.

Scripts/Runtime/Entities/PersistentData/Data/Interface/IEntityPersistentDataInstance.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// </summary>
+    public interface IEntityPersistentDataInstance
+    {
+        /// <summary>
+        /// Called whenever an element is being disposed during cleanup.
+        /// This allows for custom cleanup code.
+        /// </summary>
+        /// <param name="entity">The entity that was associated with this data</param>
+        public void DisposeForEntity(Entity entity);
+    }
```
mbaker3, line 17: Chatted about this already but since neither of us can come up with a use case for needing the Entity in disposal let's delete the current disposal method and have the interface extend `IDisposable`

Scripts/Runtime/Entities/PersistentData/Data/Interface/IThreadPersistentDataInstance.cs
```diff
@@ -0,0 +1,22 @@
...
+        /// <param name="threadIndex">The thread index to be associated with</param>
+        public void ConstructForThread(int threadIndex);
+        
+        /// <summary>
+        /// Called whenever an element is being disposed during cleanup.
+        /// This allows for custom cleanup code.
+        /// </summary>
+        /// <param name="threadIndex">The thread index that was associated with this data</param>
+        public void DisposeForThread(int threadIndex);
+    }
```
mbaker3, line 21: Same comment as `IEntityPersistentDataInstance`

I say we leave the Construct for thread since we need some sort of construct method anyway and if the index is ever needed for disposal then it can be stored and used. I don't have a strong opinion on this though. If you don't want params that's cool too.

mbaker3 (commented):
Looking good. Exciting changes!
Just a few minor comments and questions

Scripts/Runtime/Data/Collections/UnsafeHeap.cs
```diff
@@ -43,6 +43,20 @@ public struct UnsafeHeap<T, TComparer> : IDisposable
         where T : unmanaged
         where TComparer : unmanaged, IComparer<T>
     {
+        /// <summary>
+        /// Whether this collection has been created or not (had memory been allocated)
+        /// </summary>
+        public bool IsCreated
```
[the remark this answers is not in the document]
jkeon: Yeah because this class is a copy/paste job of the `NativeHeap` that was written by someone else, we didn't bother to convert over to our style. 

We could, but we aren't going to use this in the future due to #151

Scripts/Runtime/Entities/PersistentData/Data/EntityPersistentData.cs
```diff
@@ -1,39 +1,33 @@
...
-            if (m_DisposalCallbackPerEntity != null)
+            ref UnsafeParallelHashMap<Entity, T> data = ref Data;
+            foreach (KeyValue<Entity, T> entry in data)
             {
-                ref UnsafeParallelHashMap<Entity, T> data = ref Data;
-                foreach (KeyValue<Entity, T> entry in data)
-                {
-                    m_DisposalCallbackPerEntity(entry.Key, entry.Value);
-                }
+                entry.Value.DisposeForEntity(entry.Key);
```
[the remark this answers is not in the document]
jkeon: No boxing because we're working with the explicit `UnsafeParallelHashMap<Entity, T>` which by virtue of using this class, we know what T is. 

The boxing only happens if you do a `foreach` on an interface like `IEnumerable` since it then uses the interface function for getting the enumerator. 

Because we have the explicit collection, duck typing gets us the right Enumerator call which returns the strong typed version.

Scripts/Runtime/Entities/PersistentData/Data/Interface/IThreadPersistentData.cs
```diff
@@ -1,10 +1,44 @@
+using Anvil.Unity.DOTS.Entities.TaskDriver;
+using Unity.Jobs;
+
 namespace Anvil.Unity.DOTS.Entities
 {
+    /// <summary>
+    /// An <see cref="IAbstractPersistentData"/> that is owned by the overall application and
+    /// used to provide a unique instance of the data that persists for a thread index.
+    /// There is only ever one of these.
```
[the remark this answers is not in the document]
jkeon: Sort of, only one `IThreadPersistentData` but one `T` per thread. I've updated to reflect that.

Scripts/Runtime/Entities/PersistentData/PersistentDataSystem.cs
jkeon: Yep, good point. I forgot we had those. Changed to data system.

Scripts/Runtime/Entities/PersistentData/Data/Interface/IEntityPersistentDataInstance.cs
jkeon: Done!

Scripts/Runtime/Entities/PersistentData/Data/Interface/IThreadPersistentDataInstance.cs
jkeon: Done. I kept the param.

Scripts/Runtime/Entities/PersistentData/Data/EntityPersistentData.cs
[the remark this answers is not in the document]
mbaker3: Ok, notice that the call is different though in IL. It's a class call vs a value call

mbaker3 (approved):

Scripts/Runtime/Entities/PersistentData/Data/EntityPersistentData.cs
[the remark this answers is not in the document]
jkeon: Weird, I'm not sure to be honest. I did just check in the profiler and we're not doing any GC Allocs

PR #184, opened 2023-03-27
Adding EntityLifecycleSystem and associated functionality

jkeon:
Providing an easy way to know when Entities have been created/imported or destroyed/evicted.

We have to manually construct some queries and do some housekeeping to find new entities that we haven't already seen before or figure out that other entities have been destroyed. 

Turns out Unity has a special function called `EntityManager.GetCreatedAndDestroyedEntitiesAsync`

End users can extend the `AbstractEntityLifecycleSystem` to slot it into their app's structural changes flow. From there, they can register different `Masks` of components to create `IReadOnlyEntityLifecycleGroup`s to be able to get Created/Imported and Destroyed/Evicted entities of that type per frame. 

This makes having downstream work like updating lookups or validating state MUCH easier and more efficient. 

### Wait... Reall...

YES! I wasn't sure how it was going to work either, but I tested and `EntityManager.GetCreatedAndDestroyedEntitiesAsync` returns both Created/Imported as well as Destroyed/Evicted. 

If a bunch of entities move from WorldA to WorldB, the we'll get those entities as Destroyed in WorldA and then we'll get them as Created in WorldB. 

This actually is nicer and lets the code have a singular path since pretty much everything is from the World's perspective anyway!

- None

 - None

 - [ ] Yes
 - [x] No

### DIFF
--- a/Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleSystem.cs
+++ b/Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleSystem.cs
@@ -163,6 +163,7 @@ namespace Anvil.Unity.DOTS.Entities
 
             //Then we can run the filter passes on created and destroyed in parallel
             dependsOn = JobHandle.CombineDependencies(
+                dependsOn,
                 UpdateCreationAsync(dependsOn),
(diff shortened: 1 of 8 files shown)


jkeon:
No longer applicable.

PR #187, opened 2023-03-29
Allow systems to be ordered against TaskDriverSystems

mbaker3:
Systems and task drivers could not be ordered on other task drivers. This is because `TaskDriverSystem<T>` was marked `internal`. This PR makes `TaskDriverSystem<T>` public.

**Note:** The changes are the bare minimum required to compile again after making the types public. @jkeon feel free to suggest visibility changes of other members that should be changed. If it's a larger effort I can create an issue for it.

There is no way to order a system or `TaskDriverSystem` on another `TaskDriverSystem`.

`TaskDriverSystem`s may be used to order against. Ex: `UpdateBefore(typeof(TaskDriverSystem<MyTaskDriver>))`

 - None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Entities/TaskDriver/System/AbstractTaskDriverSystem.cs
+++ b/Scripts/Runtime/Entities/TaskDriver/System/AbstractTaskDriverSystem.cs
@@ -7,7 +7,8 @@ using Unity.Jobs;
 
 namespace Anvil.Unity.DOTS.Entities.TaskDriver
 {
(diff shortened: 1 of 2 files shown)


jkeon (approved):
Looks good. 

I made my own commit to this to lock everything down that didn't need to be exposed. Required implementing some explicit interface calls but that's alright. 

Just FYI in case you wanted to review that.

Issue #192, opened 2023-03-31
EntitySpawnSystem - Devise a way to have definitions spawn extra entities by archetype


mbaker3:
There is no efficient way to spawn multiple entities from a single `IEntitySpawnDefinition`. This is common practice for entities that generate other entities as a side effect, own proxy entities, or who's components have been split up for performance reasons.

The solution for this may not add to `IEntitySpawnDefinition` but instead create another type that wraps/groups many definitions to a higher concept actor/instance.

For now, a workaround has been implemented as part of (#193) that allows a definition to be created within the `PopulateOnEntity` method of another definition. This workaround gets the job done but isn't as efficient as instantiating from archtype.

Maybe the proper solution is to provide a mechanism to fetch archetypes for a given definition as needed. That's really the only piece missing here. If this is the direction we take consider whether all definitions should be instantiated through the `IEntitySpawnDefinitionExtension.CreateAndPopulate<T>()` method.

jkeon:
This is now merged

PR #193, opened 2023-03-31
Add IEntitySpawnDefinitionExtension.CreateAndPopulate()

mbaker3:
Add a mechanism to create and populate an `IEntitySpawnDefinitionExtension` on an `EntityCommandBuffer` when the `Archetype` from `EntitySpawnSystem` is not available.

This is a temporary workaround until #192 is resolved.

If an `IEntitySpawnDefinitionExtension` needs to spawn multiple entities as part of its creation there is no simple way to leverage other `IEntitySpawnDefinitionExtension` instances to help with the creation.

`IEntitySpawnDefinitionExtension` can now have `CreateAndPopulate<T>()` called on them to create and populate an entity from the definition and return the entity created so the parent definition may store a reference to the entity.

 - None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Entities/Lifecycle/IEntitySpawnDefinitionExtension.cs
+++ b/Scripts/Runtime/Entities/Lifecycle/IEntitySpawnDefinitionExtension.cs
@@ -1,0 +1,30 @@
+using Unity.Entities;
+
+namespace Anvil.Unity.DOTS.Entities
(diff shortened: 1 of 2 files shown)


jkeon (approved):
Nice, this is a good workaround and the related issue looks good too.

PR #194, opened 2023-04-04
Entity Lifecycle

jkeon:
Adds the ability to track the lifecycle of entities based on component filters. 

There isn't a nice way to do this.

Entities can come into a world in two ways.

1. They are created.
2. They are imported from another world.

The `IEntityLifecycleStatus` treats both of these cases as "Arrivals" and will give you access to a list of Arrivals for a given frame. 

Entities can exit a world in two ways as well.

1. They are destroyed.
2. They are evicted to another world.

The `IEntityLifecycleStatus` treats both of these cases as "Departures" and will give you access to a list of Departures for a given frame.

There is a special case with destruction where an `Entity` has a `ISystemStateComponent` on it which results in the `Entity` not being destroyed until later on in the frame and thus not counting as a Departure.

By telling the `EntityLifecycleStatusSystem` to count CleanupEntities as Departures, those entities that are in the cleanup state will count as departures and be ignored when they are actually destroyed.

Coupled together this provides an efficient ground truth for building lookups or other data to know what entities are in play without having to handle the different cases separately for each instance.

- None

 - None

 - [ ] Yes
 - [x] No


Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
```diff
@@ -0,0 +1,303 @@
...
+{
+    /// <summary>
+    /// Provides information on the lifecycle of entities.
+    /// </summary>
+    /// <remarks>
+    /// Best practice is to extend this system and place the newly created system at a point in time when structural
+    /// changes have been made. Ex. Right after a <see cref="EntityCommandBufferSystem"/>.
+    /// Created or Imported entities will be treated as "Arrivals" to the World.
+    /// Destroyed or Evicted entities will be treated as "Departures" from the World.
+    /// Entities that are in a "CleanUp" state can be treated as "Departures" before they are actually destroyed by
```
mbaker3, line 20: This appears to be an option not a default or hard-coded behaviour.
I'm happy to have this as the default behaviour but we should specify that and set the default on the constructor param.

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
```diff
@@ -0,0 +1,303 @@
...
+            }
+
+            public static WorldEntityState GetOrCreate(World world, AbstractEntityLifecycleStatusSystem lifecycleStatusSystem)
+            {
+                if (!s_WorldEntityStates.TryGetValue(world, out WorldEntityState worldEntityState))
+                {
+                    worldEntityState = new WorldEntityState(world);
+                    s_WorldEntityStates.Add(world, worldEntityState);
+                }
+                worldEntityState.AddLifecycleStatusSystem(lifecycleStatusSystem);
```
mbaker3, line 143: This should be a separate public method or described in the method name (Ex: `GetOrCreateAndAddLifecycleSystem`)

In the spirit of single responsibility methods my preference is to make `AddLifecycleStatusSystem` a separate public method that needs to be called.

Scripts/Runtime/Entities/Lifecycle/EntityLifecycleStatus.cs
```diff
@@ -0,0 +1,251 @@
...
+    {
+        private readonly AccessControlledValue<NativeParallelHashSet<Entity>> m_Lookup;
+        private readonly AccessControlledValue<NativeList<Entity>> m_ArrivedEntities;
+        private readonly AccessControlledValue<NativeList<Entity>> m_DepartedEntities;
+        private readonly AbstractEntityLifecycleStatusSystem m_OwningSystem;
+        private readonly ComponentType[] m_QueryComponentTypes;
+
+        private EntityQuery m_Query;
+
+        public EntityLifecycleStatus(AbstractEntityLifecycleStatusSystem owningSystem, params ComponentType[] queryComponentTypes)
```
mbaker3, line 22: Maybe we should accept `EntityQueryDesc` instead? Or at add support another day.

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
```diff
@@ -0,0 +1,303 @@
...
+            base.OnDestroy();
+        }
+        
+        /// <summary>
+        /// Creates a new <see cref="IEntityLifecycleStatus"/> object to monitor Arrivals/Departures from a certain
+        /// archetype.
+        /// </summary>
+        /// <param name="componentTypes">The components to filter on.</param>
+        /// <returns>The <see cref="IEntityLifecycleStatus"/> object</returns>
+        protected IEntityLifecycleStatus CreateEntityLifecycleStatus(params ComponentType[] componentTypes)
```
mbaker3, line 74: In actual usage does this ever get called outside of OnCreate?

I'm wondering if we should make an abstract method that the subclass has to implement and pass back all the different queries it wants to track. Knowing to call this is easy to overlook.

Alternatively, could we just pass all the different component queries that we want back in the constructor? They don't require World, EntityManager or the like to create.

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
```diff
@@ -0,0 +1,303 @@
...
+        protected override void OnCreate()
+        {
+            base.OnCreate();
+
+            m_WorldEntityState = WorldEntityState.GetOrCreate(World, this);
+
+            foreach (EntityLifecycleStatus entityLifecycleStatus in m_EntityLifecycleStatus)
+            {
+                entityLifecycleStatus.CreateQuery();
+            }
```
mbaker3, line 53: What risk are we avoiding by creating the queries for the subclass vs just having them create and pass back the query themselves?

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
```diff
@@ -0,0 +1,303 @@
...
+            //If nothing was created or destroyed we can early exit
+            if (createdEntities.Length == 0
+                && destroyedEntities.Length == 0)
+            {
+                m_WorldEntityState.ReleaseCreatedAndDestroyedEntities(dependsOn);
+                Dependency = dependsOn;
+                return;
+            }
+
+            Logger.Debug($"{World} | {UnityEngine.Time.frameCount} - Created: {createdEntities.Length} - Destroyed: {destroyedEntities.Length}");
```
mbaker3, line 99: Accidentally left over from debugging?

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
```diff
@@ -0,0 +1,303 @@
...
+        {
+            JobHandle dependsOn = Dependency;
+            dependsOn = m_WorldEntityState.AcquireCreatedAndDestroyedEntities(
+                dependsOn,
+                out NativeArray<Entity>.ReadOnly createdEntities,
+                out NativeArray<Entity>.ReadOnly destroyedEntities,
+                m_ShouldCleanupEntitiesCountAsDepartures);
+
+            //If nothing was created or destroyed we can early exit
+            if (createdEntities.Length == 0
```
mbaker3, line 91: Why do we care if `createdEntities.Length == 0` if we don't do anything with the collection?

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
```diff
@@ -0,0 +1,303 @@
...
+            {
+                m_World = world;
+                m_LifecycleStatusSystems = new HashSet<AbstractEntityLifecycleStatusSystem>();
+                m_AccessController = new AccessController();
+
+                m_WorldEntitiesState = new NativeList<int>(ChunkUtil.MaxElementsPerChunk<Entity>(), Allocator.Persistent);
+                m_WorldCreatedEntities = new NativeList<Entity>(ChunkUtil.MaxElementsPerChunk<Entity>(), Allocator.Persistent);
+                m_WorldDestroyedEntities = new NativeList<Entity>(ChunkUtil.MaxElementsPerChunk<Entity>(), Allocator.Persistent);
+                m_WorldCleanupEntities = new NativeParallelHashSet<Entity>(ChunkUtil.MaxElementsPerChunk<Entity>(), Allocator.Persistent);
+                m_CleanupFrame = new NativeReference<int>(-1, Allocator.Persistent);
```
mbaker3, line 169: well that's neat...

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
```diff
@@ -0,0 +1,303 @@
...
+
+            public JobHandle AcquireCreatedAndDestroyedEntities(
+                JobHandle dependsOn,
+                out NativeArray<Entity>.ReadOnly createdEntities,
+                out NativeArray<Entity>.ReadOnly destroyedEntities,
+                bool shouldIncludeCleanupEntities)
+            {
+                m_AccessController.Acquire(AccessType.ExclusiveWrite);
+                
+                m_World.EntityManager.GetCreatedAndDestroyedEntities(m_WorldEntitiesState, m_WorldCreatedEntities, m_WorldDestroyedEntities);
```
mbaker3, line 207: I don't know enough about this call but does this get the created/destroyed entities at the current moment or the ones that were created/destroyed at the start of the frame?

If it's current then having the Acquire method also fetch updated data is counter-intuitive.

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
```diff
@@ -0,0 +1,303 @@
...
+
+                    IncludeCleanupEntitiesJob includeCleanupEntitiesJob = new IncludeCleanupEntitiesJob(
+                        m_WorldCleanupEntities,
+                        m_WorldDestroyedEntities,
+                        cleanupEntities,
+                        m_CleanupFrame,
+                        UnityEngine.Time.frameCount);
+                    cleanupDependsOn = includeCleanupEntitiesJob.Schedule(cleanupDependsOn);
+                    cleanupEntities.Dispose(cleanupDependsOn);
+                    cleanupDependsOn.Complete();
```
mbaker3, line 221: curious why you have to complete this. Can't you just roll the dependency into `dependsOn` and pass it back to the caller?

mbaker3: oh, I see you need those `m_WorldCreatedEntities` and `m_WorldDestroyedEntities` resolved right away. Why are you passing through a `dependsOn` and treadting this like it's async then? Is anything actually touching `dependsOn` in here? Something fishy going on...

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
```diff
@@ -0,0 +1,303 @@
...
+                m_WorldDestroyedEntities = worldDestroyedEntities;
+                m_PendingCleanupEntities = pendingCleanupEntities;
+                m_CleanupFrame = cleanupFrame;
+                m_CurrentFrame = currentFrame;
+            }
+
+            public void Execute()
+            {
+                //First we need to ensure that none of the destroyed entities were entities we already counted
+                //as destroyed because they were clean up entities the previous time we called this
```
mbaker3, line 266: Is it guaranteed that cleanup entities will be destroyed the next frame? or is it up to when the system components get removed?

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
```diff
@@ -0,0 +1,303 @@
...
+                
+                //We'll add all the pending clean up entities to the destroyed list
+                m_WorldDestroyedEntities.AddRange(m_PendingCleanupEntities);
+                
+                //Now we'll go through all the entities that are going to be cleaned up later in the frame.
+                for (int i = 0; i < m_PendingCleanupEntities.Length; ++i)
+                {
+                    //And add them to the lookup so we don't double count them
+                    m_WorldCleanupEntities.Add(m_PendingCleanupEntities[i]);
+                }
```
mbaker3, line 299:
```suggestion
                m_WorldCleanupEntities.UnionWith(m_PendingCleanupEntities)
```

Scripts/Runtime/Entities/Lifecycle/EntityLifecycleStatus.cs
```diff
@@ -0,0 +1,251 @@
...
+        public AccessControlledValue<NativeList<Entity>>.AccessHandle AcquireArrivals()
+        {
+            return m_ArrivedEntities.AcquireWithHandle(AccessType.SharedRead);
+        }
+
+        /// <inheritdoc cref="IEntityLifecycleStatus.AcquireDepartures"/>
+        public AccessControlledValue<NativeList<Entity>>.AccessHandle AcquireDepartures()
+        {
+            return m_DepartedEntities.AcquireWithHandle(AccessType.SharedRead);
+        }
```
mbaker3, line 88: Why not just expose the ACVs for `m_ArrivedEntities` and `m_DepartedEntities`?

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
jkeon: Ok sure, updated.

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
jkeon: I guess none. I had done it this way so that you could do the configuration and assignment of your `IEntityLifecycleStatus` objects in your constructor rather than waiting for `OnCreate`. It also reduces the work you have to do by creating the query for you.

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
jkeon: This function is called by your derived class to create an object that listens/reports the lifecycle status for a certain set of components.

```csharp
        public RenderingEntityLifecycleStatusSystem() : base(true)
        {
            EnvironmentRendererLifecycleStatus = CreateEntityLifecycleStatus(
                ComponentType.ReadOnly<EnvironmentRendererTag>(),
                ComponentType.ReadOnly<EnvironmentID>());
    
            ActorRendererLifecycleStatus = CreateEntityLifecycleStatus(
                ComponentType.ReadOnly<ActorRendererTag>(),
                ComponentType.ReadOnly<ActorID>());
        }
```

I don't think you'll overlook it because you need it to create the object anyway. 

Abstract method would be called by the constructor so I think it's probably better to just leave as the constructor.

I think you're thinking you'd extend this for each set of components you want to track, but it's more that you'd extend the system for each structural point in a frame you want to measure at and then create the inner lifecycle status objects for each set of components you want to track at that point.

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
jkeon: > we can early exit

😄

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
jkeon: Yep, deleted.

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
jkeon: Ok done

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
jkeon: Indeed! First time I've ever used it.

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
jkeon: The reason for this is that Unity will operate on the underlying NativeLists on the main thread at the time of the call.

Imagine the scenario where you have two of these systems in the world at different points in the frame. One for Simulation, one for Rendering for example.

Sim goes first, the `GetCreatedAndDestroyedEntities` call happens and you get back the data arrays synchronously. 

You then schedule some jobs to operate on that data. 

If the Rendering one gets to go while those jobs are still in flight, those data arrays that they were operating on have now been modified. 

By using this acquire/release pattern, if any jobs were in flight, they'll get completed before we call `GetCreatedAndDestroyedEntities` again.

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
jkeon: Good catch on the fishyness, I've updated to properly combine and complete the incoming job handles.

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
jkeon: It is up to when the system components get removed, however in practice this always ends up being in the frame because the systems who remove those components get their queries triggered.

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
jkeon: Well done!

Scripts/Runtime/Entities/Lifecycle/EntityLifecycleStatus.cs
jkeon: I'm fine with adding it another day, I find I normally just like using the params array for faster typing and not needing to create the desc object

Scripts/Runtime/Entities/Lifecycle/EntityLifecycleStatus.cs
jkeon: I don't want you to be able to write to them and #195 was made after this so I don't have `IReadOnlyAccessControlledValue` yet. 

Added TODO

jkeon:
Edits made but most were just answered your questions. Happy to chat about them if that's prefered

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
mbaker3: Ok, this does limit how advanced the query can be. It's fine for now but maybe want to consider the change in the future?

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
mbaker3: Ah ok, I didn't realize you'd keep reference to the created status objects. I'd just let consumers create this object and then register it with the subclass rather than having a factory method but this is ok for now.

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
mbaker3: But you don't use `createdEntities` at all afterward? Seems like you only care that `destroyedEntities.Length==0`

Scripts/Runtime/Entities/Lifecycle/EntityLifecycleStatus.cs
mbaker3: Yea, maybe an overload then. I could see wanting to submit a more advanced query that used things like exclusions.

mbaker3 (dismissed):
Looking good. Some comments to think of creating a few more TODOs and issues for the future and one still not sure about that 0 length check.

Feel free to merge if you don't agree.

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
jkeon: Yeah i think it's completely fine to add in support for EntityQueryDesc in the future. Added TODO

Scripts/Runtime/Entities/Lifecycle/AbstractEntityLifecycleStatusSystem.cs
jkeon: Updated the comment to explain more what's going on. 

We don't use the `createdEntities` because it's not as efficient as using queries. But we do use the length as a signal to tell us that "new entities (of any type) have been created" therefore we want our LifecycleStatus objects to run their Updates which will use their underlying queries to detect if there are any new entities that match what they are looking for.

If I remove this, then they will only update and find new entities if we also happened to destroy entities too.

Scripts/Runtime/Entities/Lifecycle/EntityLifecycleStatus.cs
jkeon: Yep that's fair, although you can do `ComponentType.Exclude<T>`

Added the TODO

mbaker3 (approved):

PR #200, opened 2023-04-13
EntityPersistentDataReader/Writer - Add helper methods

mbaker3:
Add helper methods to `EntityPersistentDataReader` and `EntityPersistentDataWriter`
 - `GetKeyArray`
 - `GetValueArray`
 - `GetEnumerator`
 - `RemoveAndDispose` (Writer only)

Consumers of EPD readers/writers are not able to iterate over all of the persistent data. This is an unusual usage bust still required in some situations.

Data in an EPD implements `IDisposable` but it's easy for the developer to forget to dispose the elements when removing them.

The pass through methods to the backing collection are now exposed.
 - `GetKeyArray`
 - `GetValueArray`
 - `GetEnumerator`

Add `RemoveAndDispose` to the writer to help make it easier to correctly remove EPD elements. I'm open to making this the default behaviour of `Remove` and instead having a `RemoveWithoutDispose` as a special case method that can be called. Thoughts?

 - None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/JobDataInteraction/EntityPersistentDataReader.cs
+++ b/Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/JobDataInteraction/EntityPersistentDataReader.cs
@@ -1,3 +1,5 @@
+using System.Collections;
+using System.Collections.Generic;
 using Unity.Collections;
 using Unity.Collections.LowLevel.Unsafe;
 using Unity.Entities;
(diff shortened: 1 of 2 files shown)


jkeon (approved):
> Add RemoveAndDispose to the writer to help make it easier to correctly remove EPD elements. I'm open to making this the default behaviour of Remove and instead having a RemoveWithoutDispose as a special case method that can be called. Thoughts?

I think it's probably better to have this as the special case. Regular `Remove` works just like it would on any other `Dictionary` or similar collection, it's just removed and you could have that object referenced elsewhere. 

The `RemoveAndDispose` lets you know you're doing something else that's not the norm but is helpful.

mbaker3:
Sounds good. I just figured that because you're dealing with persistent data...which is something special. There's almost no case where you would want to remove and not dispose.

PR #202, opened 2023-04-13
IJobConfig Requirement Delegation

mbaker3:
Add `AddRequirementsFrom()` to `IJobConfig` to allow requirement configuration to be offloaded to another method.

Generally this will reference a static method on the trigger job so that requirements can be listed more closely to their usage rather than at the configuration call site.

Offloading requirement configuration cannot be done as part of the chained pattern used for all other requirement configuration.

Using `AddRequirementsFrom()` developers can now offload all or a portion of requirement configuration to another method.

 - None

 - None

 - [ ] Yes 
 - [x] No (unless you have your own `IJobConfig` implementation)

### DIFF
--- a/Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/JobConfig/AbstractJobConfig.cs
+++ b/Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/JobConfig/AbstractJobConfig.cs
@@ -264,6 +264,12 @@ namespace Anvil.Unity.DOTS.Entities.TaskDriver
             return this;
         }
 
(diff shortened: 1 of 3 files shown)


jkeon (commented):
I think this is good, but personally I'd like to see the `Func<T, IJobConfig, IJobConfig>` be a typed delegate like we do in `JobConfigScheduleDelegates`.

The reason being that we can more easily understand what the params mean (by giving them names) and documenting the delegate itself.

mbaker3:
Good call, Done

jkeon (approved):

PR #204, opened 2023-04-13
AbstractTaskDriver - Assert subtask instance has no parent.

mbaker3:
Add an assertion to `AddSubTaskDriver` to make sure the subtask instance hasn't already been added to another Task Driver.

Though it will produce incorrect behaviour it is possible to add one subtask driver instance to multiple task drivers (or even the same task driver multiple times)

An assertion will be thrown when a subtask driver instance is added multiple times.

 - None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Entities/TaskDriver/AbstractTaskDriver.cs
+++ b/Scripts/Runtime/Entities/TaskDriver/AbstractTaskDriver.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using System.Reflection;
 using Unity.Entities;
 using Unity.Jobs;
+using Debug = UnityEngine.Debug;
 
 namespace Anvil.Unity.DOTS.Entities.TaskDriver
 {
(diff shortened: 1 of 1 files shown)


jkeon (approved):
Looks good but.....

🚲  - Thoughts on conditional ANVIL_DEBUG_SAFETY method with nice exception instead?

mbaker3:
Can open a new PR to adjust if we decide to change but to me this is a super cheap check that will only come up which you're in the process of adding subtasks and developing this feature. It's not like an edge case behaviour at runtime will suddenly trigger this condition.

So a message isn't required. It's pretty apparent what's gone wrong if this triggers.
In terms of `ANVIL_DEBUG_SAFETY` vs `DEBUG`, really if there was an editor specific assert I would have used that because this is never going to happen in a build if you've already tested in the editor.

Rationale
 - Cleaner.
 - Never going to happen in a build.
 - Obvious what has gone wrong.

I vote we leave it as is.

jkeon:
🚚  💥  🚲 
💀

PR #205, opened 2023-04-13
Add Enumerable ToFixedString Extension Methods

mbaker3:
Provide a reasonably convenient way to convert collections of Entities and `IToFixedString` implementations to a Fixed String representation. 

Introduce `IToFixedString` to provide an interface for any type to be able to generate a burst compatible string output of itself. The burst counterpart to `object.ToString()`.

The resulting string is `[element1String,elemetn2string,...]`

Developers
 - Have to manually generate fixed strings to represent collections of objects in a burst context.
 - Have no way of defining a ToFixedString method in a consistent way.

ToFixedString() may be called on any of the built in or Anvil burst compatible collections (or enumerators) that contain elements which implement `IToFixedString`. The generic parameters required to call this method aren't ideal but I haven't been able to devise a better solution.

Collections of entities are a special case of extension methods since they have a ToFixedString method but do not implement IToFixedString. Leveraging aliasing we're able to wrap the entities in an `IToFixedString` implementation without any additional allocations.
The entity collection calls just happen to be much nicer to call with no generic parameters required.

 - None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Entities/Data/EntityEnumerableToFixedStringExtension.cs
+++ b/Scripts/Runtime/Entities/Data/EntityEnumerableToFixedStringExtension.cs
@@ -1,6 +1,7 @@
 using Anvil.Unity.DOTS.Core;
 using Anvil.Unity.DOTS.Data;
 using System;
+using System.Runtime.CompilerServices;
 using Unity.Burst;
 using Unity.Collections;
 using Unity.Collections.LowLevel.Unsafe;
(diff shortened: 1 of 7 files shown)


Scripts/Runtime/Entities/Data/EntityEnumerableToFixedStringExtension.cs
```diff
@@ -0,0 +1,244 @@
...
+            return UnsafeUtility.As<FixedList4096Bytes<Entity>, FixedList4096Bytes<EntityWrapper>>(ref collection)
+                .ToFixedString<EntityWrapper, FixedString64Bytes, FixedString4096Bytes>();
+        }
+
+        /// <summary>
+        /// Wraps an entity and implements <see cref="IToFixedString{T}"/>.
+        /// </summary>
+        private struct EntityWrapper : IToFixedString<FixedString64Bytes>, IEquatable<EntityWrapper>
+        {
+            private Entity Entity;
```
jkeon, line 237:
```suggestion
            private Entity m_Entity;
```

jkeon (approved):
One minor change.

Your programming professors seeing you write about aliasing and solving this:
https://media.giphy.com/media/xKy2w6LehxxHa/giphy.gif

Me reading this:
https://media.giphy.com/media/eax0rh3OERAYg/giphy.gif

jkeon (approved):

PR #211, opened 2023-04-13
i192 Spawning Archetypes - See: https://github.com/decline-cookies/anvil-unity-dots/pull/198

jkeon:
Allowing easy proxy entities to be created within a definition

There was no nice easy way to get a performant spawn of a proxy entity from within an existing `IEntitySpawnDefinition`'s `PopulateOnEntity` function.

- Added the existence of `EntitySpawnHelper`.
    - This adds a Burst safe lookup for `EntityArchetypes` and prototype entities, allowing for the easy creation of proxy entities.
- Able to consolidate and remove a bunch of code for the different code paths on Archetype vs Prototype spawning.
- Using reflection to generate the lookup for Archetypes and determining if we can use Burst or not.
- You can register and unregister the `Prototype` entity for a given definition. No need to pass it in on every call.

- #192 

 - #195 

 - [x] Yes - Slight API changes for functions requiring a prototype
 - [ ] No


mbaker3 (approved):
Approved Via https://github.com/decline-cookies/anvil-unity-dots/pull/198

PR #213, opened 2023-04-14
Fixing an order of operations error on system destroy

jkeon:
There was an exception being thrown due systems being destroyed in random order and one system depending on another.

See above

We just check to see if we've been disposed already. In that case, what we were trying to elegantly remove has already been disposed.

- None

 - None

 - [ ] Yes
 - [x] No

### DIFF
--- a/Scripts/Runtime/Entities/Lifecycle/EntitySpawnSystem.cs
+++ b/Scripts/Runtime/Entities/Lifecycle/EntitySpawnSystem.cs
@@ -66,9 +66,9 @@ namespace Anvil.Unity.DOTS.Entities
         protected override void OnDestroy()
         {
             m_ActiveEntitySpawners.Clear();
+            m_EntitySpawners.DisposeAllValuesAndClear();
             m_EntityArchetypes.Dispose();
             m_EntityPrototypes.Dispose();
-            m_EntitySpawners.DisposeAllValuesAndClear();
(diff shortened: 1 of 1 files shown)


mbaker3 (approved):

Scripts/Runtime/Entities/Lifecycle/EntitySpawnSystem.cs
```diff
@@ -370,6 +370,13 @@ public void RegisterEntityPrototypeForDefinition<TEntitySpawnDefinition>(Entity
         public void UnregisterEntityPrototypeForDefinition<TEntitySpawnDefinition>(bool shouldDestroy)
             where TEntitySpawnDefinition : unmanaged, IEntitySpawnDefinition
         {
+            //If we're tearing down and this system was destroyed before whoever was trying to unregister, we 
+            //can't do anything. It's ok, because our Dispose would have handled freeing everything up.
+            if (m_EntityPrototypes.IsDisposed)
```
mbaker3, line 375: 🚲 Might want to check `this.IsDisposed` instead? Unless the disposal of the other fields is causing the call to this method.

jkeon: There is no `IsDisposed` on the system and it feels a bit weird to introduce one just for this system.

PR #216, opened 2023-04-17
Access Controlled Values - Expand flexibility

mbaker3:
Create Shared and Exclusive write access controlled value interfaces. This allows developers to expose ACVs with write specific access (Ex: Stream writers)

Also added a flexible access controlled value interface. This allows developers to access control values that need more flexibility than what `AccessControlledValue<T>` offers.

Finally, added `AccessControlledValue<T>.AccessHandle.CreateDerived<U>()` to allow ACV implementations to transform/wrap an access controlled value before exposing it to the consumer but continue to be backed by the access controller managing access to the underlying data.

All of this work begins to lay the groundwork for unifying the TaskDriver/Stream/ACV interfaces as much as possible and giving developers the opportunity to write access controlled types that are automatically able to be pulled into an `IJobConfig`

## Tag Alongs
 - Remove "Only" from all of the names.
 - Adapt the `IJobConfig` methods to accept the new, more specific, ACV interfaces

 - There are no interfaces to expose write only ACVs.
    - Ex: Data stream writers
 - There is no way to provide access control to a value that may want to do some work on each access or release request. The developer's only option is to wrap an ACV instance and re-create/expose their own, incompatible, interface.
 - If a raw value that is being access controlled needs to be transformed or wrapped per-request (Ex: Stream Writers) there is no way to do this without applying access control to all values in an instance.
    - There's no issue with this but it makes it unclear in the implementation what parts of the type actually require the access control.

 - `ISharedWriteAccessControlledValue<T>` and `IExclusiveWriteAccessControlledValue<T>` have been introduced to provided shared and exclusive write only interfaces.
    - As part of this work, the names were getting really unwieldy so "Only" has been removed from all type and method names.
 - `IAccessControlledValue<T>` is has been added to allow any instance to provide flexible access control without needing to subclass `AccessControlledValue<T>`
    - `IAccessControleldValue<T>` doesn't extend all of the limited access interfaces. This limits how much a developer has to implement to meet the interface but I'm on the fence this one. (see inline comment)
 - `AccessControlledValue<T>` implements all of the new interfaces since it provides all of this functionality.
 - Add `AccessControlledValue<T>.AccessHandle.CreateDerived<U>()` to allow a public facing value to be access controlled value to be exposed while pinning the access lifecycle to an internal ACV.

 - None

 - None

 - [x] Yes - Update any existing references to ReadOnly on the ACV types and methods to remove "Only"
 - [ ] No


Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/JobConfig/AbstractJobConfig.cs
```diff
@@ -174,24 +174,24 @@ public IJobConfig RequestCancelFor(AbstractTaskDriver taskDriver)
...
-        public IJobConfig RequireGenericDataForRead<TData>(IReadOnlyAccessControlledValue<TData> collection)
+        public IJobConfig RequireGenericDataForRead<TData>(IReadAccessControlledValue<TData> collection)
             where TData : struct
         {
             AddAccessWrapper(new GenericDataReadOnlyAccessWrapper<TData>(collection, Usage.Default));
             return this;
         }
 
-        public IJobConfig RequireGenericDataForWrite<TData>(AccessControlledValue<TData> collection)
+        public IJobConfig RequireGenericDataForWrite<TData>(ISharedWriteAccessControlledValue<TData> collection)
```
jkeon, line 184: Minor but we should refactor this function name to be:

```suggestion
        public IJobConfig RequireGenericDataForSharedWrite<TData>(ISharedWriteAccessControlledValue<TData> collection)
```

Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/JobConfig/Interface/IJobConfig.cs
```diff
@@ -88,7 +88,7 @@ public IJobConfig RequireGenericDataForRead<TData>(IReadOnlyAccessControlledValu
         /// for use in your job.
         /// </remarks>
         /// <returns>A reference to itself to continue chaining configuration methods</returns>
-        public IJobConfig RequireGenericDataForWrite<TData>(AccessControlledValue<TData> data)
+        public IJobConfig RequireGenericDataForWrite<TData>(ISharedWriteAccessControlledValue<TData> data)
```
jkeon, line 91:
```suggestion
        public IJobConfig RequireGenericDataForSharedWrite<TData>(ISharedWriteAccessControlledValue<TData> data)
```

Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/JobData/AbstractJobData.cs
```diff
@@ -92,7 +92,7 @@ public TData GetGenericDataForReading<TData>()
         public TData GetGenericDataForWriting<TData>()
```
jkeon, line 92:
```suggestion
        public TData GetGenericDataForSharedWriting<TData>()
```

Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/Wrapper/GenericExclusiveWriteDataAccessWrapper.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        private readonly IExclusiveWriteAccessControlledValue<TData> m_AccessControlledData;
+        private TData m_Data;
+
+        public TData Data
+        {
+            get => m_Data;
+        }
+
+        public GenericExclusiveWriteDataAccessWrapper(IExclusiveWriteAccessControlledValue<TData> data, AccessType accessType, AbstractJobConfig.Usage usage) : base(accessType, usage)
```
jkeon, line 17:
```suggestion
        public GenericExclusiveWriteDataAccessWrapper(IExclusiveWriteAccessControlledValue<TData> data, AbstractJobConfig.Usage usage) : base(AccessType.ExclusiveWrite, usage)
```

jkeon: Might want to check the SharedWrite and Read versions of this type of wrapper. Less error prone.

Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/Wrapper/GenericSharedWriteDataAccessWrapper.cs
```diff
@@ -3,25 +3,25 @@
...
+        private readonly ISharedWriteAccessControlledValue<TData> m_AccessControlledData;
         private TData m_Data;
 
         public TData Data
         {
             get => m_Data;
         }
 
-        public GenericDataAccessWrapper(AccessControlledValue<TData> data, AccessType accessType, AbstractJobConfig.Usage usage) : base(accessType, usage)
+        public GenericSharedWriteDataAccessWrapper(ISharedWriteAccessControlledValue<TData> data, AccessType accessType, AbstractJobConfig.Usage usage) : base(accessType, usage)
```
jkeon, line 17:
```suggestion
        public GenericSharedWriteDataAccessWrapper(ISharedWriteAccessControlledValue<TData> data, AbstractJobConfig.Usage usage) : base(AccessType.SharedWrite, usage)
```

Scripts/Runtime/Job/AccessControl/IAccessControlledValue.cs
```diff
@@ -0,0 +1,73 @@
+using Unity.Jobs;
+
+namespace Anvil.Unity.DOTS.Jobs
+{
+    /// <summary>
+    /// An interface that describes an access controlled value that can grant read/shared write and exclusive write
+    /// access.
+    /// </summary>
+    /// <typeparam name="T">The type of value being access controlled.</typeparam>
+    public interface IAccessControlledValue<T>
```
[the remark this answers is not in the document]
jkeon: Yeah I see what you mean. 

What if `IAccessControlledValue` only implemented the `Release` and `ReleaseAsync` methods. 
`IReadAccessControlledValue`, `IExclusiveAccessControlledValue` and `ISharedWriteAccessControlledValue` all inherit from that. 

Then there is final interface which lets you decide what access you want called `IConfigurableAccessControlledValue` that inherits off of `IAccessControlledValue`. 

Naming up for debate, I'm not too invested in that but at least then they all have the common `IAccessControlledValue` as a base.

jkeon (changes requested):
Looks good, some minor comments/suggestions

Scripts/Runtime/Job/AccessControl/IAccessControlledValue.cs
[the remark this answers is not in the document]
mbaker3: That approach doesn't really get us any free compatibility. It just avoids the duplicate definition of Release/ReleaseAsync.

When I was working on this, trying to avoid duplicating the release definitions I did consider what you've proposed (or using Abstract in the name). The problems are:
 - An object with just Release/ReleaseAsync isn't really a thing. You'd never use that interface other than extending it with another interface
 - Abstract interfaces aren't a thing so there's nothing to prevent using it on its own.

At the end of the day duplicate release definitions aren't a huge deal. If you modify one implementation or one interface all the other ones are going to error so you'll get to the bottom of the issue quickly.

The original comment I was more thinking about a developer that only implements `IAccessControlledValue` (in its current form) and while that type is surely compatible with `IReadAccessControlledValue`'s operations it can't be used unless the authoring developer remembers to also implement that interface.

Maybe this is something we just kick down the road and solve with extension methods, default implementations, or a wrapper in the future.

Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/JobConfig/AbstractJobConfig.cs
mbaker3: Fixed

Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/JobConfig/Interface/IJobConfig.cs
mbaker3: Fixed

Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/Wrapper/GenericExclusiveWriteDataAccessWrapper.cs
mbaker3: Fixed

Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/Wrapper/GenericSharedWriteDataAccessWrapper.cs
mbaker3: Fixed

mbaker3:
Comments addressed. Ready for re-review

Scripts/Runtime/Job/AccessControl/IAccessControlledValue.cs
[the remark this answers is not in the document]
jkeon: Ah i see, yeah I think it's fine to kick down the road. 

The other problem to add to that list is you can't put a bunch of `I*AccessControlledValue`s in the same collection where * is Read or SharedWrite or ExclusiveWrite etc. But not critical right now.

jkeon (approved):

PR #221, opened 2023-04-19
Adding extensions to get elements as refs from Arrays

jkeon:
Adding similar methods to `UnsafeArray` and `NativeArray` as were already contained on `UnsafeList` and `DynamicBufferExtension`

- We didn't have these methods.

- Now we do.

 - None

 - None

 - [ ] Yes
 - [x] No

### DIFF
--- a/Scripts/Runtime/Data/Collections/Util/NativeArrayExtension.cs
+++ b/Scripts/Runtime/Data/Collections/Util/NativeArrayExtension.cs
@@ -11,6 +11,29 @@ namespace Anvil.Unity.DOTS.Data
     public static class NativeArrayExtension
     {
         /// <summary>
+        /// Returns a reference to the element at a given index.
+        /// </summary>
+        /// <param name="nativeArray">The <see cref="NativeArray{T}"/></param>
+        /// <param name="index">The index to access. Must be in the range of [0..Length).</param>
+        /// <returns>A reference to the element at the index.</returns>
+        public static unsafe ref T ElementAt<T>(this NativeArray<T> nativeArray, int index) where T : unmanaged
+        {
+            Debug.Assert(index >= 0 && index < nativeArray.Length);
+            return ref UnsafeUtility.ArrayElementAsRef<T>(nativeArray.GetUnsafePtr(), index);
+        }
+        
+        /// <summary>
+        /// Returns a read only reference to the element at a given index.
+        /// </summary>
+        /// <param name="nativeArray">The <see cref="NativeArray{T}"/></param>
+        /// <param name="index">The index to access. Must be in the range of [0..Length).</param>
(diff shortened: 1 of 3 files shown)


mbaker3 (approved):

Scripts/Runtime/Data/Collections/Util/NativeArrayExtension.cs
```diff
@@ -10,6 +10,29 @@ namespace Anvil.Unity.DOTS.Data
...
+        
+        /// <summary>
+        /// Returns a read only reference to the element at a given index.
+        /// </summary>
+        /// <param name="nativeArray">The <see cref="NativeArray{T}"/></param>
+        /// <param name="index">The index to access. Must be in the range of [0..Length).</param>
+        /// <returns>A reference to the element at the index.</returns>
+        public static ref readonly T ElementAtReadOnly<T>(this NativeArray<T> nativeArray, int index) where T : unmanaged
+        {
+            return ref nativeArray.ElementAt(index);
```
mbaker3, line 33: Shouldn't we use `array.GetUnsafeReadOnlyPtr()` here?

Scripts/Runtime/Data/Collections/Util/UnsafeArrayExtension.cs
```diff
@@ -0,0 +1,34 @@
...
+        
+        /// <summary>
+        /// Returns a read only reference to the element at a given index.
+        /// </summary>
+        /// <param name="unsafeArray">The <see cref="UnsafeArray{T}"/></param>
+        /// <param name="index">The index to access. Must be in the range of [0..Length).</param>
+        /// <returns>A reference to the element at the index.</returns>
+        public static ref readonly T ElementAtReadOnly<T>(this UnsafeArray<T> unsafeArray, int index) where T : unmanaged
+        {
+            return ref unsafeArray.ElementAt(index);
```
mbaker3, line 31: Shouldn't we use `array.GetUnsafeReadOnlyPtr()` here?

mbaker3 (changes requested):

Scripts/Runtime/Data/Collections/Util/UnsafeArrayExtension.cs
jkeon: No, because that function doesn't exist. 

With `NativeArray` there are the two functions that both just return the same pointer but the safety check they do inside checks for Read/Write access or ReadOnly access respectively. 

The pointer they return is still just a pointer so you can write as much as you want to the ReadOnly one but that's obviously dangerous. 

With the `UnsafeArray` there are no safety checks so only `GetUnsafePtr` exists.

mbaker3 (approved):

PR #226, opened 2023-04-21
ComponentTypeDependencyExtension - Add

mbaker3:
Provides ability to get the job dependency on an arbitrary component or set of components.

@jkeon This may provide the opportunity to decouple task drivers from systems (if we want that)

**Question:**
Does making these extension methods of `ComponentSystemBase` and `EntityManager` make sense?
Alternatively (or additionally) we can have the extension methods work off of the `ComponentType<T>` instances themselves.
Examples:
 - `ComponentType<MyComponent>.ReadOnly().GetDependency(EntityManager)` (or system)
 -  `myComponentCollection.GetDependency(EntityManager)` (or system)

`EntityManager` and `ComponentSystemBase` are the providers for the pointer to the underlying `ComponentDependencyManager` instance. That's why the current extension methods are setup as they are.

Currently, the only way to get the scheduling dependency of a collection of components is through an `EntityQuery`.

With a reference to `EntityManager` or `ComponentSystemBase` instance developers can calculate the scheduling dependency on an arbitrary collection of components.

 - None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/AssemblyInjections/Unity.Entities/ComponentTypeDependencyExtension.cs
+++ b/Scripts/AssemblyInjections/Unity.Entities/ComponentTypeDependencyExtension.cs
@@ -1,0 +1,171 @@
+using Unity.Collections;
+using Unity.Collections.LowLevel.Unsafe;
+using Unity.Entities;
+using Unity.Jobs;
+using UnityEngine;
+
+/// <summary>
+/// A collection of extension methods to help calculate the scheduling dependencies on <see cref="ComponentType"/>s.
+/// </summary>
+[BurstCompatible]
+public static class ComponentTypeDependencyExtension
+{
+    private static UnsafeList<int> s_WriteTypeList_ScratchPad = new UnsafeList<int>(0, Allocator.Persistent);
+    private static UnsafeList<int> s_ReadTypeList_ScratchPad = new UnsafeList<int>(0, Allocator.Persistent);
+
+    /// <summary>
+    /// Get the dependency of an individual component type.
+    /// Useful for scheduling jobs that depend on components out of band with a system's
+    /// <see cref="ComponentSystemBase.OnUpdate"/>.
+    /// </summary>
+    /// <param name="manager"> The <see cref="World"/>'s <see cref="EntityManager"/>.</param>
+    /// <param name="component">The component type to get the dependency of.</param>
+    /// <returns>The dependency for the component type</returns>
+    public static unsafe JobHandle GetDependency(this EntityManager manager, ComponentType componentType)
+    {
+        return GetDependency(manager.GetUncheckedEntityDataAccess()->DependencyManager, componentType);
+    }
+
+    /// <summary>
+    /// Get the combined dependency of a collection of component types.
+    /// Useful for scheduling jobs that depend on components out of band with a system's
+    /// <see cref="ComponentSystemBase.OnUpdate"/>.
+    /// </summary>
+    /// <param name="manager"> The <see cref="World"/>'s <see cref="EntityManager"/>.</param>
+    /// <param name="component">The component types to calculate the dependency of.</param>
+    /// <returns>The combined dependency for the component types</returns>
+    [NotBurstCompatible]
+    public static unsafe JobHandle GetDependency(this EntityManager manager, params ComponentType[] componentTypes)
+    {
+        return GetDependency(manager.GetUncheckedEntityDataAccess()->DependencyManager, componentTypes);
+    }
+
+    /// <summary>
+    /// Get the combined dependency of a collection of component types.
+    /// Useful for scheduling jobs that depend on components out of band with a system's
+    /// <see cref="ComponentSystemBase.OnUpdate"/>.
+    /// </summary>
+    /// <param name="manager"> The <see cref="World"/>'s <see cref="EntityManager"/>.</param>
+    /// <param name="component">The component types to calculate the dependency of.</param>
+    /// <returns>The combined dependency for the component types</returns>
+    public static unsafe JobHandle GetDependency(this EntityManager manager, NativeArray<ComponentType> componentTypes)
+    {
+        return GetDependency(manager.GetUncheckedEntityDataAccess()->DependencyManager, componentTypes);
+    }
+
+    /// <summary>
+    /// Get the dependency of an individual component type.
+    /// Useful for scheduling jobs out of band with a system's <see cref="ComponentSystemBase.OnUpdate"/>.
+    /// </summary>
+    /// <param name="system">
+    /// The system running the out of band job. If not in a system use the <see cref="EntityManager"/> extension instead.
+    /// </param>
+    /// <param name="component">The component type to get the dependency of.</param>
+    /// <returns>The dependency for the component type</returns>
+    public static unsafe JobHandle GetDependency(this ComponentSystemBase system, ComponentType componentType)
+    {
+        return GetDependency(system.CheckedState()->m_DependencyManager, componentType);
+    }
+
+    /// <summary>
+    /// Get the combined dependency of a collection of component types.
+    /// Useful for scheduling jobs out of band with a system's <see cref="ComponentSystemBase.OnUpdate"/>.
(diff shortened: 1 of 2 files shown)


Scripts/AssemblyInjections/Unity.Entities/ComponentTypeDependencyExtension.cs
```diff
@@ -0,0 +1,171 @@
...
+using UnityEngine;
+
+/// <summary>
+/// A collection of extension methods to help calculate the scheduling dependencies on <see cref="ComponentType"/>s.
+/// </summary>
+[BurstCompatible]
+public static class ComponentTypeDependencyExtension
+{
+    private static UnsafeList<int> s_WriteTypeList_ScratchPad = new UnsafeList<int>(0, Allocator.Persistent);
+    private static UnsafeList<int> s_ReadTypeList_ScratchPad = new UnsafeList<int>(0, Allocator.Persistent);
```
jkeon, line 14: You might get the Memory leak error here because nothing ever disposes the collection before Unity quits.
I've been using the following pattern for this:

```csharp
        private static void Init()
        {
              if (s_WriteTypeList_ScratchPad.IsCreated)
              {
                     s_WriteTypeList.Dispose();
              }
              s_WriteTypeList_ScratchPad = new UnsafeList<int>(0, Allocator.Persistent);
          }
```

Scripts/AssemblyInjections/Unity.Entities/ComponentTypeDependencyExtension.cs
```diff
@@ -0,0 +1,171 @@
...
+using Unity.Jobs;
+using UnityEngine;
+
+/// <summary>
+/// A collection of extension methods to help calculate the scheduling dependencies on <see cref="ComponentType"/>s.
+/// </summary>
+[BurstCompatible]
+public static class ComponentTypeDependencyExtension
+{
+    private static UnsafeList<int> s_WriteTypeList_ScratchPad = new UnsafeList<int>(0, Allocator.Persistent);
```
jkeon, line 13: Should we allow it to allocate an initial amount like 16? Tiny amount of memory but avoids the realloc when this is first used and probably won't go over?

At 0, we'll realloc to 1, then 2, then 4 and then 8 each time we hit 1, 2, 3, and 5 items respectively

jkeon (dismissed):
This is awesome!

> This may provide the opportunity to decouple task drivers from systems (if we want that)

Hmmn, I think that would be good. It would reduce complexity for sure. But it would make the ability to order things relative to systems a bit trickier. 

We'd have to have "groups" of TaskDrivers that execute off a TaskDriverGroupLauncherSystem so you could order your other systems relative to them. 

Which is probably fine and not often used.

My only question is how does the other side of this work? 

I have a TaskDriver job that operates out of band and requires access to `MyComponent : IComponentData`. 

So TaskDriver gets the dependency for me so I know when I can use it which is great. 

How do I let Unity know to update it's dependency for `MyComponent` to wait on my TaskDriver job?

jkeon (commented):
Oops, approved too soon. Let me know what you think of the inline comments.

jkeon:
Ah... sorry. I just saw this in action here: https://github.com/decline-cookies/station-x/pull/430/

I think we should remove the `ComponentSystemBase` extensions and just have the EntityManager one. 

This line:
`dependsOn = JobHandle.CombineDependencies(dependsOn, this.GetDependency(ComponentType.ReadWrite<IsSelected>()));`

Reads to me like we are getting this system's dependency for the component which is a bit confusing since my system may not interact with the component at all. 

`dependsOn = JobHandle.CombineDependencies(dependsOn, EntityManager.GetDependency(ComponentType.ReadWrite<IsSelected>()));`

This reads better that we are trying to get whatever this component's global access dependency is. 

This would also read well:

`dependsOn = JobHandle.CombineDependencies(dependsOn, ComponentType.ReadWrite<IsSelected>().GetDepencency(ref EntityManager));

Scripts/AssemblyInjections/Unity.Entities/ComponentTypeDependencyExtension.cs
mbaker3: Not sure I follow. If the domain gets reloaded the sandbox of memory gets reset right?
Otherwise, the static collections won't get re-instantiated anyway.

Your pattern just seems to reset the collection(s) between play sessions without code changes. I guess that does help with deterministic behaviour though

Scripts/AssemblyInjections/Unity.Entities/ComponentTypeDependencyExtension.cs
mbaker3: I was following Unity's pattern for Systems. They do the same where it's 0 alloc. I think the idea is that if you never use this feature it doesn't take up any memory.

Scripts/AssemblyInjections/Unity.Entities/ComponentTypeDependencyExtension.cs
jkeon: > If the domain gets reloaded the sandbox of memory gets reset right?

Yes, but if you don't dispose your native collections, then Unity detects that and squawks at you because you didn't dispose before the sandbox was reset.

Scripts/AssemblyInjections/Unity.Entities/ComponentTypeDependencyExtension.cs
jkeon: Interesting. I wouldn't have thought that was a good idea but fair enough.

mbaker3: Yea, realistically, most of the time on the first frame you'll get your max alloc settled and be done. It's only if you kick off some obscure job later that requires a bunch of dependencies that you'll see a new alloc.

If it was an issue we could always warm the extension ourselves by making a call with a large number of components.

Scripts/AssemblyInjections/Unity.Entities/ComponentTypeDependencyExtension.cs
mbaker3: ah, ok. will make the edit

mbaker3: FYI this error doesn't trigger for `UnsafeList`. I added the change anyway for more per-run behaviour.

mbaker3: Done

Scripts/AssemblyInjections/Unity.Entities/ComponentTypeDependencyExtension.cs
```diff
@@ -10,20 +11,30 @@
...
 {
-    private static UnsafeList<int> s_WriteTypeList_ScratchPad = new UnsafeList<int>(0, Allocator.Persistent);
-    private static UnsafeList<int> s_ReadTypeList_ScratchPad = new UnsafeList<int>(0, Allocator.Persistent);
+    private static UnsafeList<int> s_WriteTypeList_ScratchPad;
+    private static UnsafeList<int> s_ReadTypeList_ScratchPad;
+
+    [RuntimeInitializeOnLoadMethod(RuntimeInitializeLoadType.SubsystemRegistration)]
+    private static void Init()
+    {
+        s_WriteTypeList_ScratchPad.Dispose();
```
jkeon, line 20: I think you have to check if `IsCreated` is called. On MOST collections this will error, but we probably should still have it just in case.

jkeon (approved):
Just the one IsCreated check, otherwise looks great!

Scripts/AssemblyInjections/Unity.Entities/ComponentTypeDependencyExtension.cs
mbaker3: OK, fixed

jkeon (approved):

Issue #139, opened 2023-01-16
Change all `ENABLE_UNITY_COLLECTIONS_CHECKS` to `ANVIL_DEBUG_SAFETY`


jkeon:
Do a pass through the code base to change these over.

mbaker3:
Related to #90

PR #231, opened 2023-04-26
Adding a Safety Check for Task Driver Systems writing to Driver Streams

jkeon:
There could be an issue where TaskDriver Systems get configured to write to a Task Driver data stream.
In the event that there is more than one TaskDriver, the System job will only actually write to the first Task Driver that exists. This will lead to data loss or false responses in the first Task Driver.

There is nothing to prevent this error, so you could shoot yourself in the foot.

A check runs to make sure that this doesn't happen and throws and exception. 

 - None

 - None

 - [x] Yes - If you were breaking this rule, you will now get an error and need to update your jobs. 
 - [ ] No

### Tag-Alongs
- Changed `ENABLE_UNITY_COLLECTIONS_CHECKS` to `ANVIL_DEBUG_SAFETY` as per #139


mbaker3 (approved):
My feet thank you

PR #235, opened 2023-04-30
Task Driver Clean Up

### LINKED ISSUE #196: TaskDriver - Make `Configure...` methods `protected` and publicly expose the cancel stream
```[tasklist]
### Tasks
- [ ] Set all `TaskDriver.Configure...` methods to `protected`
- [ ] Expose Cancel stream publicly
``` 

## Rationale

Currently the configure methods on task driver instances are all public (`ConfigureDriverJobTriggeredBy()`, etc..).

Before task driver streams had consolidation phases it was necessary to schedule one task drivers jobs on another task driver's update phase. This is no longer the case.

Publicly exposing the configure methods is
 - No longer necessary
 - Allows task drivers to be created that are much more difficult to understand since their jobs aren't scheduled to run at the same phase that their backing system is.
 - Breaks the ability to reliably guard against duplicate scheduling on the same data stream with the same job (see `TaskSet.Debug_EnsureNoDuplicateJobSchedulingDelegates()`)
    - It incorrectly prevents multiple instances of a task driver to listen to the data on a different, single, task driver instance's stream.

jkeon:
Changing `public` methods to `protected` and exposing CancelRequest and CancelComplete streams publicly as outlined in #196 

Less than ideal setup which results in more code, special paths, and a less consistent pattern for developers to learn and use.

- Less code!
- More consistency
- Stronger alignment with what you are scheduling jobs on.
- Public facing `IDriverCancelRequestDataStream` and `ISystemCancelRequestDataStream`
   - It would be good to get rid of these eventually and make the `CancelRequest` be a `IEntityProxyDataInstance` just like `CancelComplete`

- Resolves #196 

 - None

 - [x] Yes - API changes. Pretty easy to migrate though.
 - [ ] No


Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/IAbstractDataStream.cs
```diff
@@ -74,5 +74,8 @@ public interface IAbstractDataStream<TInstance> : IAbstractDataStream
     /// The first is a parallel writable collection to be able to write pending instances to.
     /// The second is a narrow array collection for reading.
     /// </summary>
-    public interface IAbstractDataStream { }
```
mbaker3, line 77: non-style compliant 🤖

mbaker3 (dismissed):
Minor style issue. Careful committing changes that aren't related to the PR :)

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/IAbstractDataStream.cs
jkeon: Oops my bad

mbaker3 (approved):

Issue #233, opened 2023-04-30
Deep Migration


jkeon:
Building off of #232 there is a chance that Entity references aren't remapped if you bury them deep inside collections. 

- There are still some ways you can miss Entities. If you have an EPD then you have a relationship of Key: Entity to Value: Something. 
   - The Key can be remapped no problem
   - The Something might be a Collection like UnsafeParallelHashMap. If the key or something in that collection's value has an Entity reference, there's no way right now for us to know. 
   - I think I can scan this an account for it, but I might have to add the check for each type of Unsafe Collection.

jkeon:
Added a PR #237 to emit a warning. 

To implement this fully we'll want to do the following:

1. Possibly just scan all types in the Assemblies.
    1. Locate the `Entity` references and check for collections that might be hiding `Entity` references.
2. When we find a collection, we register that to the `TypeOffsetInfo` for the type so that we know where in the struct that collection is located and also what type is is. HashMap vs Array etc. 
3. When we go to remap, there's the basic remap of `Entity` fields like normal
4. But we also check if there are any fields that might have `Entity` references hidden inside them and run those as well.
5. And we also check if there are any Collections fields that might have `Entity` references hidden inside them and run those too. 

For the nested struct check, it's just piping through the same remap function.
For the collections check, we'll need to check the "type" of collection via the `TypeOffsetInfo` and map that to a function that exists to be able to remap the contents of the collection. 
 - The remapping for a HashMap will be different vs an Array for example. 
 - If no remapping function exists for a type, we'll error to let the developer register a custom one. (In case they build their own custom collection).
 - For each instance in the collection, we just run through the regular remap function. 

The recursive nature will ensure that we nab every single nested Entity reference even if it's in a collection of collections of collections etc.

jkeon:
From : https://github.com/decline-cookies/anvil-unity-dots/pull/237#discussion_r1181785278
> I don't know where you want to put this for future reference but we discussed how there may be cases where a pointer is being stored that we do not want to dive into an migrate. Maybe that memory address has already been migrated or something.
> 
> We want to figure out how to handle these situations. Maybe require developers to annotate unsafe fields with a [RemapUnsafeEntityDataOnMove] or something. Would need an [ExcludeUnsafeEntityDataOnMove] (names WIP)

There may be cases where we don't want a collection or pointer to be remapped, so we'll make the developers opt in on this to both and squawk when they don't let us know how to handle it.

PR #237, opened 2023-05-01
i233 Deep Migration Warning

jkeon:
Scanning types that are being registered for Migration and emitting a warning if there's a chance we could be hiding Entity references inside them. 

Precursor to more involved work in #233 

No way to tell other than developer thoroughness

We now scan the type being registered for Entity remapping and see if anything could be hidden via an Unsafe collection's pointer. 

We then emit a warning to the console so developers can ensure it is handled. 

- None, although this helps with #233 

 - #232 
 - 
 - [ ] Yes 
 - [x] No


Scripts/Runtime/Entities/Lifecycle/Migration/MigrationUtil.cs
```diff
@@ -113,23 +116,43 @@ public static void RegisterTypeForEntityPatching(Type type)
...
                     entityOffsetStartIndex,
                     s_EntityOffsetList.Length));
-            
+
             //The size of the underlying data could have changed such that we re-allocated the memory, so we'll update
             //our shared statics
             UpdateSharedStatics();
         }
 
+        private static void ScanForCollections(string parentPathString, Type type)
```
mbaker3, line 125: Should this be a safety only operation? I don't think this would come up uniquely in release

Scripts/Runtime/Entities/Lifecycle/Migration/MigrationUtil.cs
```diff
@@ -113,23 +116,43 @@ public static void RegisterTypeForEntityPatching(Type type)
...
+            foreach (FieldInfo field in fields)
+            {
+                Type fieldType = field.FieldType;
+                if (fieldType.IsPrimitive)
+                {
+                    continue;
+                }
+                if (fieldType.IsPointer)
+                {
+                    Debug.LogWarning($"{parentPathString}/{type.GetReadableName()} has a field named {field.Name} which is a pointer. This is probably a collection. As a result, we cannot automatically patch any entity references inside this collection. It would need to be handled manually until automatic work can handle in from https://github.com/decline-cookies/anvil-unity-dots/issues/233.");
```
mbaker3, line 138:
```suggestion
                    // TODO: #233 - Update instructions when implemented.
                    Debug.LogWarning($"{parentPathString}/{type.GetReadableName()} has a field named {field.Name} which is a pointer. This is probably a collection. As a result, we cannot automatically patch any entity references inside this collection.");
```

Scripts/Runtime/Entities/Lifecycle/Migration/MigrationUtil.cs
```diff
@@ -113,23 +116,43 @@ public static void RegisterTypeForEntityPatching(Type type)
...
+            //One of these fields might be a collection or contain a collection, best way to tell is to scan it and see if it has a pointer
+            FieldInfo[] fields = type.GetFields(BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic);
+            foreach (FieldInfo field in fields)
+            {
+                Type fieldType = field.FieldType;
+                if (fieldType.IsPrimitive)
+                {
+                    continue;
+                }
+                if (fieldType.IsPointer)
```
mbaker3, line 136: I don't know where you want to put this for future reference but we discussed how there may be cases where a pointer is being stored that we do not want to dive into an migrate. Maybe that memory address has already been migrated or something.

We want to figure out how to handle these situations. Maybe require developers to annotate unsafe fields with a `[RemapUnsafeEntityDataOnMove]` or something. Would need an `[ExcludeUnsafeEntityDataOnMove]` (names WIP)

mbaker3 (commented):
Looks good. Minor suggestions

Scripts/Runtime/Entities/Lifecycle/Migration/MigrationUtil.cs
jkeon: Done

Scripts/Runtime/Entities/Lifecycle/Migration/MigrationUtil.cs
jkeon: Cool, logged in #233

mbaker3 (approved):

PR #238, opened 2023-05-02
EntityPersistentData - Rework to be API compatible with ACV

mbaker3:
Reworked the API for `EntityPersistentData` to be compatible with the appropriate AccessControlledValue interfaces. This allows more flexible use with other ACV based features.

The `EntityPersistentData` accomplishes many of the same things that the ACV interfaces do but are just slightly different. This makes them incompatible with types written to generically deal with ACV types.

`EntityPersistentData` implements `ISharedWriteAccessControlledValue<EntityPersistentDataWriter<T>>` and `IReadAccessControlledValue<EntityPersistentDataReader<T>>` allowing it to continue delivering specialized readers/writers depending on requested access but also be compatible as a specialized ACV.

In the future we can evaluate whether it makes sense to add Exclusive Write support and support the entire ACV API.

 - None

 - None

 - [x] Yes 
 - [ ] No

Find and replace:
 - `AcquireReader...` -> `AcquireRead...`
 - `AcquireWriter...` -> `AcquireSharedWrite...`
 - `ReleaseWriter...` -> `Release...`
 - `ReleaseReader...` -> `Release...`

### DIFF
--- a/Scripts/Runtime/Entities/PersistentData/Data/EntityPersistentData.cs
+++ b/Scripts/Runtime/Entities/PersistentData/Data/EntityPersistentData.cs
@@ -92,8 +92,8 @@ namespace Anvil.Unity.DOTS.Entities
             //Launch the migration job to get that burst speed
             dependsOn = JobHandle.CombineDependencies(
                 dependsOn,
-                AcquireWriterAsync(out EntityPersistentDataWriter<T> currentData),
-                destinationEntityPersistentData.AcquireWriterAsync(out EntityPersistentDataWriter<T> destinationData));
+                AcquireSharedWriteAsync(out EntityPersistentDataWriter<T> currentData),
+                destinationEntityPersistentData.AcquireSharedWriteAsync(out EntityPersistentDataWriter<T> destinationData));
 
             MigrateJob migrateJob = new MigrateJob(
                 currentData,
@@ -101,8 +101,8 @@ namespace Anvil.Unity.DOTS.Entities
                 ref remapArray);
             dependsOn = migrateJob.Schedule(dependsOn);
 
-            destinationEntityPersistentData.ReleaseWriterAsync(dependsOn);
-            ReleaseWriterAsync(dependsOn);
+            destinationEntityPersistentData.ReleaseAsync(dependsOn);
+            ReleaseAsync(dependsOn);
 
             return dependsOn;
         }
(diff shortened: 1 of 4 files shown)


Scripts/Runtime/Entities/PersistentData/Data/EntityPersistentData.cs
```diff
@@ -97,18 +90,19 @@ public JobHandle MigrateTo(JobHandle dependsOn, IMigratablePersistentData destin
             EntityPersistentData<T> destinationEntityPersistentData = (EntityPersistentData<T>)destinationPersistentData;
 
             //Launch the migration job to get that burst speed
-            dependsOn = JobHandle.CombineDependencies(dependsOn,
-                AcquireWriterAsync(out EntityPersistentDataWriter<T> currentData),
-                destinationEntityPersistentData.AcquireWriterAsync(out EntityPersistentDataWriter<T> destinationData));
+            dependsOn = JobHandle.CombineDependencies(
+                dependsOn,
+                AcquireSharedWriteAsync(out EntityPersistentDataWriter<T> currentData),
+                destinationEntityPersistentData.AcquireSharedWriteAsync(out EntityPersistentDataWriter<T> destinationData));
```
jkeon, line 96: I didn't notice this before because of the API, but these should really be Exclusive. Is it a huge pain to add the Exclusive support now?

Scripts/Runtime/Entities/PersistentData/Data/Interface/IEntityPersistentData.cs
```diff
@@ -10,36 +11,8 @@ namespace Anvil.Unity.DOTS.Entities
...
-
-        /// <summary>
-        /// Allows other jobs or code to use to underlying data for the <see cref="EntityPersistentDataWriter{TInstance}"/>
-        /// and ensures data integrity across those other usages.
-        /// </summary>
-        public void ReleaseWriter();
-    }
+    public interface IEntityPersistentData<T> : IAbstractPersistentData,
+                                                IReadOnlyEntityPersistentData<T>,
+                                                ISharedWriteAccessControlledValue<EntityPersistentDataWriter<T>>
```
jkeon, line 16: Also Exclusive if we add that

jkeon (changes requested):
Looks great, just one additional request

Scripts/Runtime/Entities/PersistentData/Data/EntityPersistentData.cs
mbaker3: Fixed

Scripts/Runtime/Entities/PersistentData/Data/Interface/IEntityPersistentData.cs
mbaker3: Fixed

mbaker3:
Comments addressed. Added exclusive write support and changed migration to use it.

jkeon (approved):

PR #239, opened 2023-05-02
WrappedAccessControlledValue - Add

mbaker3:
Provide a type to wrap and transform another ACV in an ACV compatible format.

If you'd like to expose a variant of an access controlled value it involves implementing a new type that implements at least one of the access controlled value interfaces. Most of this type is generic boilerplate with only the logic transforming the base type to the variant being unique.

`WrappedAccessControlledValue` and `WrappedReadAccessControlledValue` provide generic implementations that allow the developer to pass in the base ACV and a delegate that transforms the type from the base to desired type.

##### Common
``` csharp
private readonly AccessControlledValue<Dictionary<string, int>> m_MyLookup = new AccessControlledValue<Dictionary<string, int>>(new Dictionary<string, int>());

public readonly struct LookupPresenceReader
{
   private Dictionary<string, int> m_Lookup;

   public bool IsPresent(string key)
      => m_Lookup.Contains(key);

   public LookupPresenceReader(Dictionary<string, int> lookup)
   {
      m_Lookup = lookup;
   }
}
```

##### Old
``` csharp
public IReadAccessControlledValue<LookupPresenceReader> IsInLookup { get; } 
   = new PresenceLookup(m_MyLookup);

private class PresenceLookup : AbstractAnvilBase, IReadAccessControlledValue<LookupPresenceReader>
{
   private readonly IReadAccessControlledValue<Dictionary<string, int>> m_Lookup;

    public PresenceLookup(IReadAccessControlledValue<Dictionary<string, int>> lookup)
    {
        m_Lookup = lookup;
    }

    public AccessControlledValue<LookupPresenceReader>.AccessHandle AcquireWithReadHandle()
    {
        var handle = m_Lookup.AcquireWithReadHandle();
        LookupPresenceReader reader = new LookupPresenceReader(handle.Value);

        return AccessControlledValue<LookupPresenceReader>.AccessHandle.CreateDerived(handle, reader);
    }

    public LookupPresenceReader AcquireRead()
    {
        return new LookupPresenceReader(m_Lookup.AcquireRead());
    }

    public JobHandle AcquireReadAsync(out LookupPresenceReader value)
    {
        JobHandle handle = m_Lookup.AcquireReadAsync(out var lookup);
        value = new LookupPresenceReader(lookup);
        return handle;
    }

    public void Release()
    {
        m_Lookup.Release();
    }

    public void ReleaseAsync(JobHandle releaseAccessDependency)
    {
        m_Lookup.ReleaseAsync(releaseAccessDependency);
    }
}
```

##### NEW - With wrapped ACV
``` csharp
public IReadAccessControlledValue<LookupPresenceReader> IsInLookup { get; } 
   = new WrappedReadAccessControlledValue<LookupPresenceReader, Dictionary<string, int>>(m_MyLookup);
```
 - None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Job/AccessControl/WrappedAccessControlledValue.cs
+++ b/Scripts/Runtime/Job/AccessControl/WrappedAccessControlledValue.cs
@@ -1,0 +1,148 @@
+using Anvil.CSharp.Core;
+using Unity.Jobs;
+
+namespace Anvil.Unity.DOTS.Jobs
+{
+    /// <summary>
+    /// Wraps an <see cref="AccessControlledValue{T}"/> and allows the controlled value to be transformed
+    /// before delivering a <see cref="TWrapped"/> instance to the consumer.
+    ///
+    /// This is an effective way of providing a constrained view of a broader internal access controlled value.
+    /// Ex: A reader/writer to an internal stream.
+    /// </summary>
+    /// <typeparam name="TWrapped">The type of the wrapped instance delivered to the consumer.</typeparam>
+    /// <typeparam name="TBase">The base access controlled value that is being wrapped.</typeparam>
+    public class WrappedAccessControlledValue<TWrapped, TBase> : AbstractAnvilBase,
+                                                                 IAccessControlledValue<TWrapped>,
+                                                                 IReadAccessControlledValue<TWrapped>,
+                                                                 ISharedWriteAccessControlledValue<TWrapped>,
+                                                                 IExclusiveWriteAccessControlledValue<TWrapped>
+    {
+        /// <summary>
+        /// Given the base value <see cref="TBase"/> produces the <see cref="TWrapped"/> to deliver to the consumer
+        /// of this instance.
+        /// </summary>
+        public delegate TWrapped WrapBaseValueDelegate(TBase baseValue);
+
+        private readonly AccessControlledValue<TBase> m_BaseACV;
+        private readonly WrapBaseValueDelegate m_WrapBaseValue;
+
+        /// <summary>
+        /// Creates a new instance of a <see cref="WrappedAccessControlledValue{TWrapped,TBase}"/>.
+        /// </summary>
+        /// <param name="baseACV">The base <see cref="AccessControlledValue{T}"/> to drive access.</param>
+        /// <param name="wrapBaseValue">Given the base value produces the wrapper value to deliver to the consumer.</param>
+        public WrappedAccessControlledValue(AccessControlledValue<TBase> baseACV, WrapBaseValueDelegate wrapBaseValue)
+        {
+            m_BaseACV = baseACV;
+            m_WrapBaseValue = wrapBaseValue;
+        }
+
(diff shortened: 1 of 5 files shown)


Scripts/Runtime/Job/AccessControl/WrappedAccessControlledValue.cs
```diff
@@ -0,0 +1,148 @@
...
+
+        private readonly AccessControlledValue<TBase> m_BaseACV;
+        private readonly WrapBaseValueDelegate m_WrapBaseValue;
+
+        /// <summary>
+        /// Creates a new instance of a <see cref="WrappedAccessControlledValue{TWrapped,TBase}"/>.
+        /// </summary>
+        /// <param name="baseACV">The base <see cref="AccessControlledValue{T}"/> to drive access.</param>
+        /// <param name="wrapBaseValue">Given the base value produces the wrapper value to deliver to the consumer.</param>
+        public WrappedAccessControlledValue(AccessControlledValue<TBase> baseACV, WrapBaseValueDelegate wrapBaseValue)
```
jkeon, line 35: I think using the delegate is completely fine in this case but just bringing it up for discussion. 

The other pattern i've seen is the use of a converter class like here: https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.comparer-1?view=net-8.0

So we would have `WrappedAccessControlledValue<TWrapped, TBase, TWrapper<TBase, TWrapper>> where TWrapper : IWrapper<TBase, TWrapper>`

You wouldn't need to create a lambda because we would instantiate the instance of the IWrapper inside here and just pass it through the concrete method of converting a TBase to a TWrapped. 

It's definitely better performance wise, but how much better I have no idea. 

The additional benefit is that if you have `WrappedAccessControlledValue`'s located in different parts of the app that both need to convert the same TBase to TWrapped, you don't need to manually make sure your lambdas are the same. You can have that logic consistent in the same place. 

This could also just be a TODO or Github issue.

jkeon (commented):
Looks good, just one implementation detail to discuss or postpone.

Scripts/Runtime/Job/AccessControl/WrappedAccessControlledValue.cs
mbaker3: That's an interesting solution but I'm not sure it works well for this.
 - We'd want to pass in the converter instance so that the developer can pull in other dependencies to construct the wrapper if required.
 - The boilerplate required for a separate converter class starts to diminish the value of the wrapper to begin with. We're not much less code then creating the boilerplate required without the wrapper.

If we're concerned about performance we may as well just make `WrappedAccessControlledValue` abstract and have developers subclass it to define their wrap creator logic.

Happy to discuss though if you still see a case.

jkeon: Ok those are fair counter-points. The main one winning me over is the need to pull in other dependencies. 

We can always revisit if we need to and it becomes a problem.

PR #242, opened 2023-05-05
AbstractJobData/Config - Replace Get* methods with a generic fulfill

mbaker3:
Replace all `Get...` methods on `AbstractJobData` and `AbstractJobConfig` with a common set of overloaded `FulFill` methods.

From:
`myData = jobData.GetDBFEForRead<MyElement>()`

To:
`jobData.FulFill(out myData)`

Each different type of data being pulled into a job requires a different get method call on `jobData`. This is tedious to write and an unnecessary call to change when changing requirements of the data (Ex: reader now needs to be a writer)

Developers only need to pass in their typed field into `jobData.FulFill(out myField)` and the instance will be resolved for them. This is a simple resolution using method overloads so it's done at compile time.

 - None

 - None

 - [x] Yes 
 - [ ] No

Change all calls to the following methods
 - `GetCancelRequestsWriter`
 - `GetDataStreamWriter`
 - `GetDataStreamReader`
 - `GetThreadPersistentDataAccessorForWriting`
 - `GetEntityPersistentDataReader`
 - `GetEntityPersistentDataWriter`
 - `GetCDFEReader`
 - `GetCDFEWriter`
 - `GetDBFEForRead`
 - `GetDBFEForExclusiveWrite`

To use the new `FulFill` method.

#### Example
``` csharp
// Old
myData = jobData.GetDBFEForRead<MyElement>()

// New
jobData.FulFill(out myData)
```

### DIFF
--- a/Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/JobConfig/AbstractJobConfig.cs
+++ b/Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/JobConfig/AbstractJobConfig.cs
@@ -475,22 +475,22 @@ namespace Anvil.Unity.DOTS.Entities.TaskDriver
             return genericDataAccessWrapper.Data;
         }
 
-        internal ThreadPersistentData<TData> GetThreadPersistentData<TData>()
+        internal void Fulfill<TData>(out ThreadPersistentData<TData> instance)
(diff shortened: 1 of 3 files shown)


jkeon (approved):

mbaker3:
Weird bug with github. This was Merged...

PR #248, opened 2023-05-11
EntitySpawner - Add SpawnImmediate for collections

mbaker3:
Add `SpawnImmediate` for collections of definitions.

There are `SpawnDeferred` overloads for collections of definitions but only individual entities may be spawned immediately.

 - An overload of `SpawnImmediate` that accepts a `NativeArray<TDefinition>`
 - An overload of `SpawnImmediate` that accepts a `NativeArray<TDefinition>` and returns a `NativeArray<Entity>` of spawned entities.

 - None

 - None

 - [ ] Yes 
 - [x] No


jkeon (approved):
I think this looks good. I would prefer that we don't spawn using the EntityManager and instead spawn using the ECB so that it mimics the exact same code path that we go down during a job. 

But i see that currently the API where it returns the NativeArray of Entities isn't possible without it.

jkeon (approved):

PR #249, opened 2023-05-16
Entity Spawner always using Burst

jkeon:
As part of the work to have Entity Spawner's being job structs that return Entity references, this PR ensures that Spawning is always using Burst. 

This is a TRANSITION PR. Everything works, and it gets closer to the final goal but I expect it to change as the Spawners become job structs so there are some weird API aspects and I'm not bothering with docs at this point. 

This state should be short lived.

Because spawning a definition could result in using a `SharedComponent` we have to detect this and use a special non-bursted job. 

With spawning ultimately happening in whatever job the end user wants  (in the future), we need to ensure that we don't have to make the end user job non-bursted as well.

We will make use of function pointers into managed code which requires the use of a lookup on the managed side to get an instance of an ECB that is currently being used.

Everything is wrapped in a new `EntityCommandBufferWithID` which will likely go away as it becomes part of the Spawner job struct.

- None

 - None

 - [x] Yes - Usage of the new `EntityCommandBufferWithID`
 - [ ] No


Scripts/Runtime/Entities/Lifecycle/Spawner/EntityCommandBufferWithID.cs
```diff
@@ -0,0 +1,152 @@
...
+            EntityCommandBufferWithIDManager.Unregister(id);
+        }
+        
+
+        public readonly uint ID;
+
+        [NativeDisableContainerSafetyRestriction] [NativeDisableUnsafePtrRestriction]
+        private EntityCommandBuffer m_EntityCommandBuffer;
+
+        public EntityCommandBufferWithID(uint id, EntityCommandBuffer entityCommandBuffer)
```
mbaker3, line 28: Can we have this created within this struct? Since we've disabled safety we should try to guarantee that the provided ECB isn't used elsewhere.

Maybe just pass in an activation delegate so the consumer can define where the ECB comes from.

mbaker3 (commented):
Looking good. Just a couple of safety oriented comments

Scripts/Runtime/Entities/Lifecycle/Spawner/EntityCommandBufferWithID.cs
jkeon: Done. 

We take in the `EntityCommandBufferSystem` if needed. We'll use it or we'll just create you a new one for temp usage.

mbaker3 (approved):

PR #250, opened 2023-05-24
WorldExtension - Sequential Create

mbaker3:
Create `WorldExtension` and a method to bulk create systems sequentially ensuring each system has completed `OnCreate` before moving to the next system.

#### Tag Alongs
 - `WorldUtil.SetupTopLevelGroupsInCurrentPlayerLoop` Improve docs
    - Incorrectly specified that system couldn't already exist
    - Highlight difference to `ScriptBehaviourUpdateOrder.AppendWorldToCurrentPlayerLoop`

Unity's existing solution `World.GetOrCreateSystemsAndLogException` instantiates all system's before calling `ComponentSystemBase.OnCreate` on each system.

This means that systems calling `world.GetOrCreateSystem` in their own `OnCreate` method may get back a system instance that has not yet had `OnCreate` called.

This is particularly problematic for `TaskDriver` instances that need system references to be fully initialized to configure jobs against the system's exposed data.

Executing `world.GetOrCreateSystemsSequentiallyAndLogExceptions` will create each system in the list sequentially (if required) calling `ComponentSystemBase.OnComplete` on the system before progressing to the next system.

 - None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Entities/Util/WorldExtension.cs
+++ b/Scripts/Runtime/Entities/Util/WorldExtension.cs
@@ -1,0 +1,57 @@
+using Anvil.CSharp.Logging;
+using System;
+using System.Collections.Generic;
+using Unity.Entities;
+using UnityEngine;
+
+namespace Anvil.Unity.DOTS.Entities
+{
+    /// <summary>
+    /// A collection of extension methods for <see cref="World"/>.
+    /// </summary>
+    /// <remarks>
+    /// Note: Most utilities are in <see cref="WorldUtil"/>. These methods are extension for better discoverability
+    ///       of replacement or supplemental behaviour to what <see cref="World"/> already offers.
+    /// </remarks>
+    public static class WorldExtension
+    {
+        /// <summary>
+        /// Creates (if required) the provided system types in the world sequentially and logs any errors during creation.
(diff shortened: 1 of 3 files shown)


Scripts/Runtime/Entities/Util/WorldExtension.cs
```diff
@@ -0,0 +1,57 @@
...
+            foreach (Type systemType in systemList)
+            {
+                try
+                {
+                    if (typeof(ComponentSystemBase).IsAssignableFrom(systemType))
+                        world.GetOrCreateSystem(systemType);
+                    else if (typeof(ISystem).IsAssignableFrom(systemType))
+                        world.GetOrCreateUnmanagedSystem(systemType);
+                    else
+                        throw new InvalidOperationException($"Invalid system type. Type:{systemType.GetReadableName()}");
```
jkeon, line 46: I know this is how Unity does it, but can we add the `{}` brackets?

jkeon (changes requested):
Looks good, one minor request

Scripts/Runtime/Entities/Util/WorldExtension.cs
mbaker3: oops yes!

mbaker3:
requested changes addressed.

jkeon (approved):

PR #253, opened 2023-05-27
System logger world context

mbaker3:
Add world name prefix to `AbstractAnvilSystemBase.Logger`.

#### Tag Alongs
 - `BurstableLogger` - Remove redundant string length value when reporting that a string was truncated. We don't know the source string's length so this always matches the fixed string's capacity.

When there are multiple worlds the developer must include the world name themselves or guess which world a system was a part of when emitting the log message.

All log messages emitted from `AbstractAnvilSystemBase` instances automatically include the world name.
Ex: "(thread: 5) (Default World) [MySystem.OnUpdate] My message"

 - None

 - None

 - [ ] Yes 
 - [x] No


jkeon (approved):
Great change!

PR #254, opened 2023-05-27
EntityLifecycleStatus - Expose Arrival/Departure ACVs directly

mbaker3:
Expose Arrival and Departure collections from `EntityLifecycleStatus` as ACV interfaces.

Reading arrival and departure entities from `EntityLifecycleStatus` within a Task Driver job is cumbersome because the acquisition API doesn't conform to an ACV interface.

Arrival and Departure lookups are exposed as `IReadAccessControlledValue<NativeList<Entity>>` getters.

 - None

 - None

 - [x] Yes 
 - [ ] No

Any use of the arrivals and departures collections will need to be adapted to use the new getters. The conversion should be familiar as they're just exposed ACVs
Ex: `entityStatus.AcquireArrivalsAsync(out var myList)` -> `entityStatus.ArrivedEntities.AcquireReadAsync(out var myList)`

### DIFF
--- a/Scripts/Runtime/Entities/Lifecycle/EntityLifecycleStatus.cs
+++ b/Scripts/Runtime/Entities/Lifecycle/EntityLifecycleStatus.cs
@@ -18,6 +18,16 @@ namespace Anvil.Unity.DOTS.Entities
 
         private EntityQuery m_Query;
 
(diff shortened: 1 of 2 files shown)


jkeon (approved):
Nice

PR #255, opened 2023-05-29
No box equals

mbaker3:
Implement a generic equality check for `unmanaged` types that doesn't cause them to be boxed.
Leveraging this allows the assertion in `AccessController` to execute without any managed allocations.

Trying to perform an equality check between `JobHandle` instances results in boxing. This applies to any value types that don't implement `IEquatable<T>` or the `==` operator.

`AccessController`'s assertion produces garbage on every execution.

`UnsafeUtil.Equals_NoBox<T>()` provides a way to check the equality between two `unamanged` instances by directly comparing their memory.

`JobHandle.Equals_NoBox` provides a way to perform an equality check between two `JobHandle` instances without boxing.

`AccessController`'s assertion no longer produces garbage.

 - None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Job/Util/JobHandleExtension.cs
+++ b/Scripts/Runtime/Job/Util/JobHandleExtension.cs
@@ -1,7 +1,13 @@
+using Anvil.Unity.DOTS.Util;
+using System.Runtime.CompilerServices;
+using Unity.Collections.LowLevel.Unsafe;
 using Unity.Jobs;
 
 namespace Anvil.Unity.DOTS.Jobs
 {
+    /// <summary>
+    /// A collection of extension methods for <see cref="JobHandle"/>.
+    /// </summary>
     public static class JobHandleExtension
     {
         /// <summary>
@@ -30,6 +36,18 @@ namespace Anvil.Unity.DOTS.Jobs
         public static bool IsDependencyOf(this JobHandle job, JobHandle candidateJob)
         {
             return JobHandle.CheckFenceIsDependencyOrDidSyncFence(job, candidateJob);
+        }
+
+        /// <summary>
+        /// Determines whether two <see cref="JobHandle"/> instances are equal without boxing.
+        /// </summary>
+        /// <param name="job1">The first <see cref="JobHandle"/> compare.</param>
+        /// <param name="job2">The second <see cref="JobHandle"/> compare.</param>
+        /// <returns>True if the <see cref="JobHandle"/>s are the same.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static unsafe bool Equals_NoBox(this in JobHandle job1, in JobHandle job2)
+        {
+            return UnsafeUtil.Equals_NoBox(in job1, in job2);
         }
     }
 }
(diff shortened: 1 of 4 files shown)


Scripts/Runtime/Job/Util/JobHandleExtension.cs
```diff
@@ -31,5 +37,17 @@ public static bool IsDependencyOf(this JobHandle job, JobHandle candidateJob)
...
         }
+
+        /// <summary>
+        /// Determines whether two <see cref="JobHandle"/> instances are equal without boxing.
+        /// </summary>
+        /// <param name="job1">The first <see cref="JobHandle"/> compare.</param>
+        /// <param name="job2">The second <see cref="JobHandle"/> compare.</param>
+        /// <returns>True if the <see cref="JobHandle"/>s are the same.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static unsafe bool Equals_NoBox(this in JobHandle job1, in JobHandle job2)
```
jkeon, line 48: Should use `ref` because `JobHandle` is not a `readonly` struct and so an defensive copy will be made.

```suggestion
        public static unsafe bool Equals_NoBox(this ref JobHandle job1, ref JobHandle job2)
```

Scripts/Runtime/Util/UnsafeUtil.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// </remarks>
+        public static bool Equals_NoBox<T>(in T value1, in T value2)
+            where T : unmanaged
+        {
+            return UnsafeUtility.MemCmp(
+                    UnsafeUtilityExtensions.AddressOf(in value1),
+                    UnsafeUtilityExtensions.AddressOf(in value2),
+                    UnsafeUtility.SizeOf<T>())
+                == 0;
+        }
```
jkeon, line 28: I get wanting to use `in` to signify that this method won't change anything but odds are pretty good that the type of `T` won't always be a `readonly` struct. If it's not a `readonly` struct then there will be a defensive copy. 

`ref` is less clear and allows us to modify if we want to, but it does guarantee we're operating on the original address and there isn't a copy going on. Almost all of Unity's API's are using `ref` instead of `in` for the same reason.

```suggestion
        public static bool Equals_NoBox<T>(ref T value1, ref T value2)
            where T : unmanaged
        {
            return UnsafeUtility.MemCmp(
                    UnsafeUtility.AddressOf(ref value1),
                    UnsafeUtility.AddressOf(ref value2),
                    UnsafeUtility.SizeOf<T>())
                == 0;
        }
```

jkeon (changes requested):
Looks great, but I think we should be using `ref` instead of `in`

Scripts/Runtime/Job/Util/JobHandleExtension.cs
mbaker3: Defensive copy is only made if you interact with a method (including getters unless marked `readonly`)

Using `in` allows us to use inline returned values when calling vs `ref` that would require us to assign to a temp variable.

Scripts/Runtime/Util/UnsafeUtil.cs
mbaker3: See other rationale. There won't be a defensive copy.

Scripts/Runtime/Job/Util/JobHandleExtension.cs
jkeon: Interesting.... I was wrong here, my bad.

jkeon (approved):

PR #257, opened 2023-05-31
Adding ECB Internal helper functions

jkeon:
Adding some assembly injections to be able to access the internals of `EntityCommandBuffer` and `EntityCommandBuffer.ParallelWriter`

We can't access some of the internals.

We now can access the internals.

1. `EntityCommandBuffer.DidPlayback`
- Allows us to know if a given `EntityCommandBuffer` has been played back either manually or via the `EntityCommandBufferSystem`

2. `EntityCommandBuffer.ParallelWriter.SetThreadIndex`
- Allows us to set the thread index the writer points to so that it writes the correct `ThreadChain`

- None

 - None

 - [ ] Yes
 - [x] No

### DIFF
--- a/Scripts/AssemblyInjections/Unity.Entities/EntityCommandBufferInternal.cs
+++ b/Scripts/AssemblyInjections/Unity.Entities/EntityCommandBufferInternal.cs
@@ -3,7 +3,7 @@ using Unity.Entities;
 namespace Anvil.Unity.DOTS.Entities
 {
     /// <summary>
-    /// Helper class to access the internals in an <see cref="EntityCommandBuffer"/>
+    /// Helper class to access the internals of an <see cref="EntityCommandBuffer"/>
     /// </summary>
     public static class EntityCommandBufferInternal
     {
@@ -11,10 +11,12 @@ namespace Anvil.Unity.DOTS.Entities
         /// Returns whether the <see cref="EntityCommandBuffer"/> has been played back or not.
         /// </summary>
         /// <param name="ecb">The <see cref="EntityCommandBuffer"/> to check.</param>
-        /// <returns>true if it has been played back, false if not</returns>
+        /// <returns>
+        /// true if it has not been disposed and has been played back, false otherwise.
+        /// </returns>
         public static unsafe bool DidPlayback(this EntityCommandBuffer ecb)
         {
-            return ecb.m_Data == null || ecb.m_Data->m_DidPlayback;
+            return ecb.IsCreated && ecb.m_Data->m_DidPlayback;
         }
     }
 }
--- a/Scripts/AssemblyInjections/Unity.Entities/EntityCommandBufferInternal.cs
+++ b/Scripts/AssemblyInjections/Unity.Entities/EntityCommandBufferInternal.cs
@@ -1,0 +1,20 @@
+using Unity.Entities;
+
+namespace Anvil.Unity.DOTS.Entities
+{
+    /// <summary>
+    /// Helper class to access the internals in an <see cref="EntityCommandBuffer"/>
+    /// </summary>
+    public static class EntityCommandBufferInternal
+    {
+        /// <summary>
+        /// Returns whether the <see cref="EntityCommandBuffer"/> has been played back or not.
+        /// </summary>
+        /// <param name="ecb">The <see cref="EntityCommandBuffer"/> to check.</param>
+        /// <returns>true if it has been played back, false if not</returns>
+        public static unsafe bool DidPlayback(this EntityCommandBuffer ecb)
+        {
+            return ecb.m_Data == null || ecb.m_Data->m_DidPlayback;
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


Scripts/AssemblyInjections/Unity.Entities/EntityCommandBufferInternal.cs
```diff
@@ -0,0 +1,20 @@
...
+    public static class EntityCommandBufferInternal
+    {
+        /// <summary>
+        /// Returns whether the <see cref="EntityCommandBuffer"/> has been played back or not.
+        /// </summary>
+        /// <param name="ecb">The <see cref="EntityCommandBuffer"/> to check.</param>
+        /// <returns>true if it has been played back, false if not</returns>
+        public static unsafe bool DidPlayback(this EntityCommandBuffer ecb)
+        {
+            return ecb.m_Data == null || ecb.m_Data->m_DidPlayback;
```
mbaker3, line 17: Right now if `m_Data == null` we return true. It that the intended behaviour we want?

That means `default` will return true as will disposed ECBs.

```suggestion
            return ecb.m_Data != null && ecb.m_Data->m_DidPlayback;
```

May also want to use `IsCreated` instead of the data null check 🤷

Scripts/AssemblyInjections/Unity.Entities/EntityCommandBufferInternal.cs
```diff
@@ -0,0 +1,20 @@
+using Unity.Entities;
+
+namespace Anvil.Unity.DOTS.Entities
+{
+    /// <summary>
+    /// Helper class to access the internals in an <see cref="EntityCommandBuffer"/>
```
mbaker3, line 6: #Consistency
```suggestion
    /// Helper class to access the internals of an <see cref="EntityCommandBuffer"/>
```

Scripts/AssemblyInjections/Unity.Entities/EntityCommandBufferParallelWriterInternal.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// </summary>
+    public static class EntityCommandBufferParallelWriterInternal
+    {
+        /// <summary>
+        /// Sets the Thread Index on a <see cref="EntityCommandBuffer.ParallelWriter"/> so that commands are written
+        /// to the correct Thread Chain
+        /// </summary>
+        /// <param name="ecbParallelWriter">The <see cref="EntityCommandBuffer.ParallelWriter"/></param>
+        /// <param name="threadIndex">The thread index to set</param>
+        public static void SetThreadIndex(ref this EntityCommandBuffer.ParallelWriter ecbParallelWriter, int threadIndex)
```
mbaker3, line 16: Can we add a `<remark>` that mentions this shouldn't usually be required and describe a situation where it's necessary?

mbaker3 (changes requested):

Scripts/AssemblyInjections/Unity.Entities/EntityCommandBufferInternal.cs
jkeon: Done

Scripts/AssemblyInjections/Unity.Entities/EntityCommandBufferInternal.cs
jkeon: It is but I guess technically this function does too much. 

I've changed it to just be a function to check if the ECB has been played back and callers of this will do their own check for `IsCreated` as well. 

The reason being that `EntityCommandBufferSystem` will dispose of the ECB after playback so we need to check both conditions for the application level code. If the data is still there, then it wasn't run through the System and we can check if it was played back, if the data isn't there then it was run through the system and we can assume it was played back.

Scripts/AssemblyInjections/Unity.Entities/EntityCommandBufferParallelWriterInternal.cs
jkeon: Yes done.

Scripts/AssemblyInjections/Unity.Entities/EntityCommandBufferInternal.cs
mbaker3: Right, yea that's tricky. You really need to understand the context around the question you're asking because in a vacuum you have no idea if the instance is `default` or was disposed

mbaker3 (approved):

Scripts/AssemblyInjections/Unity.Entities/EntityCommandBufferInternal.cs
jkeon: Agreed. I think for ECBs in particular, with Unity's current usage, if you're asking this question it's to see if you can clean up after yourself and a `default` or disposed instance will prompt the same behaviour.

PR #258, opened 2023-05-31
Atomic Burstable ID Provider

jkeon:
Adding in an `IDProvider` but that is atomic and burstable allowing for generating ID's in Bursted Jobs and in parallel.

`IDProvider` is a class and cannot be used in jobs.

`IDProviderAtomicBurstable` is a struct that is Burst Safe and allows the ID to be generated in an atomic fashion so it is also thread safe.

- None

 - #257 

 - [ ] Yes
 - [x] No


Scripts/Runtime/Data/IDProviderAtomicBurstable.cs
```diff
@@ -0,0 +1,76 @@
...
+using UnityEngine;
+
+namespace Anvil.Unity.DOTS.Data
+{
+    /// <summary>
+    /// An <see cref="IDProvider"/> that is compatible with Burst and will get the next ID atomically so it is
+    /// safe to use in threaded context.
+    /// </summary>
+    [BurstCompatible]
+    public readonly unsafe struct IDProviderAtomicBurstable
```
mbaker3, line 14: Based on the other burstable type variants we have this makes the name consistent.

```suggestion
    public readonly unsafe struct BurstableAtomicIDProvider
```

Scripts/Runtime/Data/IDProviderAtomicBurstable.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// <summary>
+        /// Creates a new instance of <see cref="IDProviderAtomicBurstable"/>.
+        /// </summary>
+        /// <remarks>
+        /// NOTE: Because this is a struct, it could be created by using the default constructor which will fail.
+        /// Debug.Asserts will catch this at runtime when used.
+        /// It is expected to create this struct and pass in an explicit supply warning threshold instead, the
+        /// <see cref="DEFAULT_SUPPLY_WARNING_THRESHOLD"/> is provided to make this easy to do so.
+        ///
+        /// ALSO NOTE: Because this must be compatible with Burst, there is no event to signify when the
```
mbaker3, line 34: I wonder if we should create a sibling system that does this for the developer and emits the event. Could just be a TODO

Scripts/Runtime/Data/IDProviderAtomicBurstable.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// <see cref="SupplyWarningThreshold"/> has been met like in <see cref="IDProvider"/>. Instead, you must
+        /// poll this struct periodically and call <see cref="HasIDExceededSupplyWarningThreshold"/>.
+        /// </remarks>
+        /// <param name="supplyWarningThreshold">The threshold to signify the ID supply is near exhaustion</param>
+        public IDProviderAtomicBurstable(uint supplyWarningThreshold)
```
mbaker3, line 39: Set a default for the parameter like `IDProvider` does

Scripts/Runtime/Data/IDProviderAtomicBurstable.cs
mbaker3, line 14: I don't think there's currently a use case but should we have a TODO to make this generic like `AbstractIDProvider<T>`?

mbaker3 (changes requested):

Scripts/Runtime/Data/IDProviderAtomicBurstable.cs
jkeon: Done

jkeon: I've added a far future TODO that relies on C#11 so we can have generic math via the static interfaces. Since this is a struct we can't inherit and there's no way to do anything with the `T` part yet.

Scripts/Runtime/Data/IDProviderAtomicBurstable.cs
jkeon: Added a TODO to #265 
Good idea.

Scripts/Runtime/Data/IDProviderAtomicBurstable.cs
jkeon: That doesn't work with structs. (in this version of C# anyway). 

If you create the new instance and don't explicitly pass in a value, it will call the default constructor with no parameters. 

The `Debug.Assert` in `GetNextID` traps for this. 

Using the system in #265 could make this nicer by making the constructor internal and having the creation go through the System. But for now, the docs handle it.

mbaker3: ah right, ok. Good catch

mbaker3 (approved):

PR #259, opened 2023-05-31
Entity Spawner Revamp

jkeon:
Entity Spawners are now job structs that allow for creating an Entity in a job and getting back the deferred Entity from the Entity Command Buffer that will create it.

- Spawning Entities are fire and forget. You have to develop some way to pick up created Entities later on. 
- This is really difficult when you need to assign specific instances to "slots" or associate relationships.

- You can acquire an `EntitySpawner` struct off of an `EntitySpawnSystem` and use that in your jobs.
- `EntitySpawner`s can create any `IEntitySpawnDefinition`, no need to have it typed beforehand anymore.
- They will return the deferred `Entity` that the underlying `EntityCommandBuffer` creates. You can use the API on the `EntitySpawner` to assign relationships that will be patched later on. 

 - None

 - #258 

 - [x] Yes - Spawning is different now although the API migration is relatively straight-forward. 
 - [ ] No


Scripts/Runtime/Entities/Lifecycle/EntitySpawnSystem.cs
```diff
@@ -30,22 +31,47 @@ namespace Anvil.Unity.DOTS.Entities
...
+            //This way we don't run out of IDs when acquiring every frame but we also don't need to force the call site
+            //to manage and store an ID.
+            
+            //If we have an ID free to use in the queue, just get it
+            if (m_InstanceIDQueue.Count > 0)
+            {
+                return m_InstanceIDQueue.Dequeue();
+            }
+            //If we don't, then increment our pool of IDs and return that instead.
+            m_LargestInstanceIDIssued++;
```
mbaker3, line 73: Better served as an IDProvider? `UShortIDProvider` if you're feeling fancy.

Scripts/Runtime/Entities/Lifecycle/EntitySpawnSystem.cs
```diff
@@ -30,22 +31,47 @@ namespace Anvil.Unity.DOTS.Entities
...
+            //We'd use a Pool<T> here but it doesn't work with primitives so we get around that with the Queue/largest
+            //instance id issued combo.
+            //This way we don't run out of IDs when acquiring every frame but we also don't need to force the call site
+            //to manage and store an ID.
+            
+            //If we have an ID free to use in the queue, just get it
+            if (m_InstanceIDQueue.Count > 0)
+            {
+                return m_InstanceIDQueue.Dequeue();
+            }
```
mbaker3, line 71: Should we have a `PooledIDProvider` or use `Pool<T>` for this? 

I looked at the `Pool<T>` implementation and it would be trivial to support value types.

Scripts/Runtime/Entities/Lifecycle/EntitySpawnSystem.cs
```diff
@@ -171,44 +151,216 @@ public void UnregisterEntityPrototypeForDefinition<TEntitySpawnDefinition>(bool
...
+        {
+            return AcquireSpawner(false, m_EntityPrototypes.Acquire(AccessType.SharedRead));
+        }
 
-            foreach (IEntitySpawner entitySpawner in m_ActiveEntitySpawners)
+        private void ReleaseImmediate(in EntitySpawner entitySpawner)
+        {
+            m_EntityPrototypes.Release();
+            entitySpawner.Playback(EntityManager);
+            entitySpawner.DisposeECB();
```
mbaker3, line 203: I'm not sure we should be playing back immediately on an immediate release. Shouldn't we still be waiting for the ECB playback moment that this `EntitySpawnSystem` is configured to execute?

If you need a job handle for `m_CommandBufferSystem.AddJobHandleForProducer` you can just pass default since your "work" with the ECB is done.

mbaker3: Oh I see, this isn't actually an ACV compatible type. Is there a way to require these in jobs without setting up a custom `Schedule` function?

Would really like to maintain compatibility with the "Require" pattern.

Scripts/Runtime/Entities/Lifecycle/JobInteraction/EntitySpawner.cs
```diff
@@ -0,0 +1,342 @@
...
+        
+        internal EntitySpawner(
+            SpawnerID id,
+            EntityCommandBuffer ecb,
+            NativeParallelHashMap<long, EntityArchetype> archetypeLookup,
+            NativeParallelHashMap<long, Entity> prototypeLookup)
+        {
+            ID = id;
+            m_ECB = ecb;
+            m_ECBWriter = m_ECB.AsParallelWriter();
```
mbaker3, line 55: Is there any overhead associated with always asking for a parallel writer?
In most TaskDriver scenarios the writer won't be written to in parallel.

mbaker3 (changes requested):
Looking good but spawners are no longer ACV compatible now and look much more difficult to bring into task drivers. Can we fix this?

Unfortunately I also don't think this will be of use to the `CreateAndIntegrateTaskDriver` since we're not able to patch the reference in the EPD off the ECB.

Scripts/Runtime/Entities/Lifecycle/EntitySpawnSystem.cs
jkeon: Yes I originally wanted to use a Pool but couldn't and it was a tangent to go down. 

Added a TODO and linked to: https://github.com/decline-cookies/anvil-csharp-core/issues/147

Scripts/Runtime/Entities/Lifecycle/EntitySpawnSystem.cs
jkeon: These functions are private. 

The public facing ones are 

`public JobHandle AcquireAsync(out EntitySpawner value)`
`public void ReleaseAsync(JobHandle releaseAccessDependency)`

I'm ok to implement one of the `IAccessController` interfaces but they add in a bunch of extra stuff like `Dispose` and the synchronous `Acquire` functions that don't quite make sense. 

I could explicitly implement those functions and just toss a `NotSupported` exception but that does feel a bit dirty.

Scripts/Runtime/Entities/Lifecycle/JobInteraction/EntitySpawner.cs
jkeon: There is probably some small amount of overhead just in terms of making the different lanes that a parallel writer can write to but that's a necessary evil.

With `EntitySpawner`s being included in jobs to directly spawn, we have to allow for parallel jobs to write otherwise we'd be constraining all jobs to execute on one thread only.

> In most TaskDriver scenarios the writer won't be written to in parallel.

I'm not sure that's true? All TaskDriver jobs are designed to go wide with most using `BatchStrategy.MaximizeChunk` so with low volume it MIGHT only execute on one thread, but with more than one chunk worth of data it will definitely go on other threads.

Scripts/Runtime/Entities/Lifecycle/EntitySpawnSystem.cs
mbaker3: Looks like all you have to do is change one line to support value types in `Pool<T>` that defaults the acquired value to `null`

Scripts/Runtime/Entities/Lifecycle/EntitySpawnSystem.cs
mbaker3: `IAccessController` doesn't have `Dispose`

For the sync interface there's no real benefit over SpawnImmediate but no a problem to support either. Maybe you want to schedule an entity to exist at the system's sync point and not until then.

Scripts/Runtime/Entities/Lifecycle/JobInteraction/EntitySpawner.cs
mbaker3: > With EntitySpawners being included in jobs to directly spawn, we have to allow for parallel jobs to write otherwise we'd be constraining all jobs to execute on one thread only.

We couldn't have an opt in to a parallel writer? Most jobs that I've written so far the spawn are running on a single thread. They tend to be low volume.

> I'm not sure that's true? All TaskDriver jobs are designed to go wide with most using BatchStrategy.MaximizeChunk so with low volume it MIGHT only execute on one thread, but with more than one chunk worth of data it will definitely go on other threads.

Are they? Almost all the Task Drivers I've written use `Schedule` for the jobs not `ScheduleParallel` to avoid clobbering EPD entries. The BatchStrategy is irrelevant in that situation I thought.

Scripts/Runtime/Entities/Lifecycle/EntitySpawnSystem.cs
jkeon: Ok, I added support with the JobConfigs for `Require` and `Fulfill` so you can get an `EntitySpawner`.

I don't think the ACV approach really makes sense for an EntitySpawner since it is the underlying `Prototypes` hash map that is actually ACV'd. 

You can still use the System like usual with `SpawnDeferred` or `SpawnImmediate` for main thread spawning. Or you can `AcquireAsync` for custom job spawning.

TaskDrivers can easily use it via `RequireEntitySpawner`

jkeon:
@mbaker3 This is ready for re-review

Scripts/Runtime/Entities/Lifecycle/EntitySpawnSystem.cs
mbaker3: hmm, ok. The way the EntitySpawnWriter's adapted before was the same idea where the writer wasn't actually what required the access control it was the data that backed it.

All these special interfaces just create headaches whenever we need to use them in generic contexts (like JobConfig). I guess we can wait until I hit that some day and convert it then.

Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/JobData/AbstractJobData.cs
```diff
@@ -51,6 +51,20 @@ public void Fulfill<TInstance>(out DataStreamActiveReader<TInstance> instance)
...
+        
+        public void Fulfill(out EntitySpawner entitySpawner)
+        {
+            entitySpawner = GetEntitySpawner();
+        }
+
+        public EntitySpawner GetEntitySpawner()
+        {
+            return m_JobConfig.GetEntitySpawner();
+        }
```
mbaker3, line 76: What's this for? I don't think any of the other types have a separate get method.

jkeon: All the methods exist for this if you don't use the `Fulfill` path. 

`GetEntityNativeArrayFromQuery`
`GetGenericDataForReading`

etc.

Scripts/Runtime/Entities/Lifecycle/EntitySpawnSystem.cs
jkeon: Discussed and while not great, it's ok enough for now.

Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/JobData/AbstractJobData.cs
mbaker3: Disucssed, Gets only exist when the fulfill for the request is ambiguous. GetEntitySpawner removed.

jkeon: Discussed and we want people to only use the `Fulfill` path if they can. Removed this function.

mbaker3 (approved):

PR #260, opened 2023-06-01
EntityProxyDataStream - Fix empty array on main thread access

mbaker3:
@jkeon Feel free to tag this for further improvement or future fixing in an issue. This was a naive attempt at fixing the issue.

Using the deferred array in a synchronous context always results in a 0 length array.

`EntityProxyDataStream` now correctly detects when the read dependency is complete and constructs a normal `NativeArray<T>` and passes that into the `DataStreamActiveReader`.

 - None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Data/Collections/DeferredNativeArray.cs
+++ b/Scripts/Runtime/Data/Collections/DeferredNativeArray.cs
@@ -459,7 +459,7 @@ namespace Anvil.Unity.DOTS.Data
 #if ENABLE_UNITY_COLLECTIONS_CHECKS
             AtomicSafetyHandle.CheckExistsAndThrow(m_Safety);
 #endif
-            NativeArray<T> array = NativeArrayUnsafeUtility.ConvertExistingDataToNativeArray<T>(m_BufferInfo->Buffer, m_BufferInfo->Length, Allocator.Invalid);
+            NativeArray<T> array = NativeArrayUnsafeUtility.ConvertExistingDataToNativeArray<T>(m_BufferInfo->Buffer, m_BufferInfo->Length, m_BufferInfo->DeferredAllocator);
 
 #if ENABLE_UNITY_COLLECTIONS_CHECKS
             NativeArrayUnsafeUtility.SetAtomicSafetyHandle(ref array, m_Safety);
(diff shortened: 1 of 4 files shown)


Scripts/Runtime/Data/Collections/DeferredNativeArray.cs
```diff
@@ -446,6 +446,21 @@ public unsafe NativeArray<T> AsDeferredJobArray()
...
+            return array;
+        }
+
+        public unsafe NativeArray<T> AsArray()
+        {
+            //This whole function taken from NativeList.AsDeferredJobArray
+#if ENABLE_UNITY_COLLECTIONS_CHECKS
+            AtomicSafetyHandle.CheckExistsAndThrow(m_Safety);
+#endif
+            NativeArray<T> array = NativeArrayUnsafeUtility.ConvertExistingDataToNativeArray<T>(m_BufferInfo->Buffer, m_BufferInfo->Length, Allocator.Invalid);
```
jkeon, line 462: Should this be `Allocator.Invalid` or whatever the buffer's allocator is?

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/EntityProxyDataStream.cs
```diff
@@ -168,7 +169,10 @@ public DataStreamPendingWriter<TInstance> CreateDataStreamPendingWriter()
 
         public DataStreamActiveReader<TInstance> CreateDataStreamActiveReader()
         {
-            return new DataStreamActiveReader<TInstance>(m_ActiveArrayData.DeferredJobArray);
+            bool isDeferredRequired = !m_ActiveArrayData.GetDependency(AccessType.SharedRead).IsCompleted;
```
jkeon, line 172: Maybe just a quick comment here as to why SharedRead and why that means we need to use the deferred?

jkeon (changes requested):
Looks great, just a couple small ones.

Scripts/Runtime/Data/Collections/DeferredNativeArray.cs
mbaker3: I suppose so. Fixed!

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/EntityProxyDataStream.cs
mbaker3: Good call, Added.

mbaker3:
Comments addressed. Ready for re-review

jkeon (approved):

PR #261, opened 2023-06-01
Allow EntityCommandBuffers to be required on Task Driver Jobs

mbaker3:
Task driver jobs can now "Require" an ECB for their job and fulfill off the `jobData`.

#### Tag Along
 - Fixed a bunch of docs ion `IJobConfig` and `AbstractJobData`

Developers must manually override the `Schedule` delegate to acquire and pass in an `EntityCommandBuffer` when desired in the job.

Developers may now use the same "Require"/"Fulfill" API that they do for all other job dependencies.

 - None

 - None

 - [ ] Yes 
 - [x] No


jkeon (approved):
Looks great! 

We might want to support an ECB that isn't governed by a `EntityCommandBufferSystem` but we can add that support if and when we need it.

mbaker3:
> We might want to support an ECB that isn't governed by a EntityCommandBufferSystem but we can add that support if and when we need it.

Yea, that'll have to be some sort of other interface in that case. Or maybe we can get away with duck typing 🤷

PR #262, opened 2023-06-01
AbstractAnvilSystemBase - Expose GetEntityQuery(EntityQueryDesc)

mbaker3:
Publicly expose `GetEntityQuery(EntityQueryDesc)` on `AbstractAnvilSystemBase` so that types that uses the "OwningSystem" pattern can get queries that set define `EntityQueryOptions`

There is no way to specify `EntityQueryOptions` with the more common `GetEntityQuery(`ComponentType<T>`) overload.

Developers have access to `GetEntityQuery(EntityQueryDesc)` so that more advanced queries can be defined for creation.

 - None

 - None

 - [ ] Yes 
 - [x] No


jkeon (approved):
Well we knew this was coming. Good change!

PR #264, opened 2023-06-02
Massively optimizing TaskDriver performance

jkeon:
Optimized TaskDrivers so they won't schedule jobs if no one has written to them.

TaskDrivers always run their jobs even if there is no data to operate on. This leads to a lot of scheduling overhead that adds up.

`TaskDriverManagementSystem` forces a sync after consolidation. This is to prevent issues with circular writing patterns such as:

- Write Pending
- Consolidate to Active (which gets a write handle to Pending again so we can clear it even though we're not actually writing)
- Which will cause the Consolidation to happen again next frame for 0 elements. 

By forcing the consolidation, we can also check the length to ensure we actually have data when scheduling.

`AbstractData` exposes a boolean to let callers know if its data could have been invalidated. If so, it's worth scheduling the job. Otherwise, no one has touched it. 

`AbstractJobConfig` will only schedule jobs if the scheduling data has data to work on.
`AbstractDataSource` will only consolidate data if there has data to work on.

- None

 - #259 

 - [ ] Yes
 - [x] No


Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/Cancellation/CancelProgressFlowNode.cs
```diff
@@ -58,6 +58,14 @@ public override string ToString()
 
         private JobHandle ScheduleCheckCancelProgressJob(JobHandle dependsOn)
         {
+            if (!m_ProgressLookupData.IsDataInvalidated && m_ParentProgressLookupData is
+            {
+                IsDataInvalidated: false
+            })
```
mbaker3, line 64: just to double check, this is just syntax sugar. It doesn't resolve to some sort of reflection.

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/DataSource/Data/AbstractData.cs
```diff
@@ -33,7 +34,17 @@ public static DataTargetID GenerateWorldUniqueID(IDataOwner dataOwner, Type abst
...
+        /// Whether the underlying data has potentially been updated by something getting write access to it.
+        /// </summary>
+        public virtual bool IsDataInvalidated
+        {
+            //If the current Read dependency has changed from what we last stored, then someone has written here
+            get => m_AccessController
+                .GetDependencyFor(AccessType.SharedRead)
+                .Equals_NoBox(m_LastSharedReadAccessJobHandle);
+        }
+
```
mbaker3, line 47: This only works if nobody has called `AcquireAsync(AccessType.SharedRead)` since write access was gained.

I think we'd be better off having the caller pass in the read handle at the last time they requested read because data invalidation is relative to the perspective of the data consumer not the provider.

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/Cancellation/CancelProgressFlowNode.cs
jkeon: Yeah it's just Rider merging the following into a pattern.

`if (!m_ProgressLookupData.IsDataInvalidated && (m_ParentProgressLookupData != null && !m_ParentProgressLookupData.IsDataInvalidated))`

I've never seen it do it like that before but there isn't any reflection or allocations going on.

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/DataSource/Data/AbstractData.cs
jkeon: I'm not sure I fully understand. I've written out a flow below. Am I missing something? 

I think we are handling this where `AbstractData` is the consumer and handles storing the read handle for the last time it requested read. 


TDA tries to read from AbstractData
- LastSharedReadAccess set to the Data's Read Dependency (1)

TDB tries to read from same AbstractData
- LastSharedReadAccess set to the Data's Read Dependency (still 1)

TDC tries to write to the same AbstractData
- Data's dependencies move up the read is now (2)

TDD tries to read from the same AbstractData
- Data is invalid because the AbstractData's LastSharedReadAccess that was stored is not the same anymore.
- LastSharedReadAccess set to the Data's Read Dependency (now 2)

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/Cancellation/CancelProgressFlowNode.cs
mbaker3: OK, just wanted to double check because I have seen that pattern matching style produce less efficient code

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/DataSource/Data/AbstractData.cs
mbaker3: After your last point what if TDA then tries to read from AbstractData?

It doesn't know that the data has changed since its last read and will think that `IsDataInvalidated`

From what I'm understanding, this results in one member's read operation hiding the data change from all other potential readers. It's only the first member that reads after a write that gets to know that the data is invalidated.

jkeon: Ugh... you're right!

Good point. I've moved the storage and checking of the read handles to where the caller for the jobs are which makes sense. We really want to know if we've been written to since the last time we ran this specific job. If we have, then our job needs to run again. If we haven't, then nothing has changed and so we don't need to run our job.

jkeon:
@mbaker3 Ready for re-review

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/Cancellation/CancelProgressFlowNode.cs
```diff
@@ -81,16 +86,13 @@ private JobHandle ScheduleCheckCancelProgressJob(JobHandle dependsOn)
                 m_CancelCompleteData.PendingWriter,
                 m_CancelCompleteDataTargetID,
                 m_ProgressLookupData.DataOwner.WorldUniqueID,
-                m_Parent != null ? m_ParentProgressLookupData.Lookup : default);
+                m_ParentProgressLookupData?.Lookup ?? default);
```
mbaker3, line 89: I think there's a performance implication to this because it creates an intermediate `Nullable<UnsafeParallelHashMap<EntityProxyInstanceID, bool>>` before hitting the `??`

mbaker3 (approved):
Looks good. 
One concern about an extra unnecessary stack alloc but feel free to skip if I'm wrong.

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/Cancellation/CancelProgressFlowNode.cs
jkeon: Interesting, I went down a rabbit hole and it appears the `??` is better. It's also what Rider often suggests. 

https://stackoverflow.com/questions/13385503/which-works-faster-null-coalesce-ternary-or-if-statement

mbaker3: Huh, ok the compiler is smarter than I thought!

Verified:
https://sharplab.io/#v2:D4AQTAjAsAUCDMACciDCiDetE8d3CiAzgC4BOArgMYmIBqAhgDYUCmAyudSfjljLkF4BuAL69hgwigBKrAGaomDIkQn8hBJIxYcuNRAFkAnp0o0A3BM05rNwnMXLVACgCUdzRps+TZ7ogAvIgAJgoMFEwkViI+OOKxiAk+EhIADmQAlgBuDCSsiI5KKkSIALYA+ibFqjGCEtIALEbumJ6SNrlk5cY6QeVVxjVEAPwAdH76tCMjoeGR0e0AKqykLmW9zG51YqmJTYgrazps/gb5pB6J3poSCaJAA=

PR #268, opened 2023-06-17
TaskDriver - Fix edge case behaviours

mbaker3:
After the optimization pass in #264 a number of edge case issues were discovered and fixed.

## Synchronous Writing
Acquiring a synchronous writer would not invalidate the data stream in a way that consumers could detect. Rather than relying on the shared read `JobHandle` to detect data invalidation we now increment a data version `uint` that changes on each write. Also, the original IsDataInvalidated check was backward but that's fixed now and I buried that little tidbit mid paragraph to save @jkeon some embarrassment.

## Cancel Unwinding
Data streams configured with `CancelBehaviour.Unwind` would not have their unwinding job triggered. This was because the regular active data stream was being checked for invalidation (and length). Changing the data invalidation check to interrogate the `pendingCancelActiveDataStream` and track its data version fixed the issue. As a result of the bug caused by confusing naming all instances of `pendingCancel...` and `pendingCancelActive...` have been changed to `activeCancel...` to avoid any confusion with the "pending" term used to represent the thread count wide data structures used by the non blocking writers.

## `CancelProgressCheck` Scheduling Without Work
`CancelProgressFlowNode` was tracking the last processed version of the data for both itself and its parent. Since all children are processed before the parent the parent version would always get invalidated for the child's next frame.
1. Child invalidates parent and self check cancel progress
2. Parent invalidates self during check cancel progress
--- next frame ---
3. Child thinks parent's data has changed since it has the data version from before the parent checked it's cancel progress.

To solve, each node now tracks its own last data version. The child will ask the parent if the parent's data is invalidated since it only matters if the data has changed since the last check cancel progress pass.

## `CancelProgressCheck` Scheduling Multiple Times For Each `TaskDriverSystem` Type 
For the purposes of cancellation, each `TaskDriver` instance has its corresponding `TaskDriverSystem` as a child. However there is only one `TaskDriverSystem` instance per `TaskDriver` type per World. This means that each `TaskDriverSystem` ends up having multiple `CancelProgressFlowNode` instances pointing to it and, as a result, separate data invalidation tracking. This means that the system streams were:
 - Having their CancelProgressChecks run multiple times per frame (once per TD instance). This was a waste of time.
 - Incrementing the data version multiple times each frame when more than one TD instance existed. This caused all systems to cancel check every frame even when there were no "real" potential writes from the outside.

To solve, each `TaskDriver` or `TaskDriverSystem` instance that a `CancelProgressFlowNode` represents now stores its last processed data version per TD/TDS instance NOT per `CancelProgressFlowNode` instance. So multiple `CancelProgressFlowNode`s that point to the same TD/TDS instance will all share a data version and only the first `CancelProgressFlowNode` processed will perform the CancelProgressCheck

## Repeated, Self, Data Invalidation
There were a number of cases where data invalidation was being detect, current data version stored, and then a write handle acquired when processing the data. The final acquisition would cause the data version to increment guaranteeing that the data would be considered "invalid" on the next frame.

"Last Data Version" is now read AFTER all acquire request are made so that the process doesn't detect invalidation caused by its own processing. This occurred in:
 - `AbstractDataSource`
 - `CancelProgressFlowNode`
 - `CancelScheduleInfo`
 - `DataStreamScheduleInfo`
 - `UpdateScheduleInfo`

## Fix Cancellation Soft Lock
Fixed soft lock if a cancel request encounters a stream configured to unwind that doesn't have any data in it.

A job configured on an unwind stream that has no data will not schedule. Since the job doesn't schedule it doesn't get a write handle to the cancel progress lookup and then doesn't increment the data version.

Since the data version doesn't increment the CancelProgressFlowNode doesn't re-process and the cancel tree is stuck in a perpetual state of waiting for the unwind to complete. Sending subsequent cancel requests will bump the processing and allow the cancel tree to resolve but this will happen for each node in the tree where no data is in the stream!

## Prevent Stale Data from persisting in data sources
Fix bug where active data would remain in the stream/lookup when the data version no longer changed.
This is an issue for any data source that is read when the data version doesn't change (Ex: CancelRequestDataSource).
Now all data sources will get one extra consolidation after the last data version change.

Specifically, this fixed an issue where cancellation would never complete for a given key when a stream configured to unwind was present in the cancellation tree, had no elements, and no writers after cancellation.

 - None

 - None

 - [ ] Yes 
 - [x] No


Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/Cancellation/CancelProgressFlowNode.cs
```diff
@@ -19,13 +22,11 @@ internal class CancelProgressFlowNode : AbstractAnvilBase
         private readonly ITaskSetOwner m_TaskSetOwner;
         private readonly CancelProgressFlowNode m_Parent;
         private readonly ActiveLookupData<EntityProxyInstanceID> m_ProgressLookupData;
-        private readonly ActiveLookupData<EntityProxyInstanceID> m_ParentProgressLookupData;
         private readonly PendingData<EntityProxyInstanceWrapper<CancelComplete>> m_CancelCompleteData;
         private readonly DataTargetID m_CancelCompleteDataTargetID;
 
+        private OwnerSharedDataVersion m_LastProcessedDataVersion;
```
jkeon, line 28: Depends on: `OwnerSharedDataVersion` comment below.

```suggestion
        private readonly OwnerSharedDataVersion m_LastProcessedDataVersion;
```

Could make readonly?

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/Cancellation/CancelProgressFlowNode.cs
```diff
@@ -212,5 +212,89 @@ private bool CheckIfWillComplete(bool isStillProcessing, ref EntityProxyInstance
...
+        ///
+        /// Call <see cref="Dispose()"/> on the instance when it is no longer required to release shared version
+        /// references.
+        /// </summary>
+        /// <remarks>
+        /// Example:
+        /// Multiple <see cref="CancelProgressFlowNode"/>s point to each <see cref="TaskDriverSystem{TTaskDriverType}"/>
+        /// instance.
+        /// </remarks>
+        private struct OwnerSharedDataVersion : IDisposable
```
jkeon, line 229: I don't have a problem with this, but curious why this is a struct and not a class? 

It only gets used in managed land? Might be more simple and not need the private inner class, could use the instance as `readonly` in the `CancelProgressFlowNode` etc.

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/DataSource/Data/AbstractData.cs
```diff
@@ -29,24 +29,39 @@ public static DataTargetID GenerateWorldUniqueID(IDataOwner dataOwner, Type abst
...
+        /// </summary>
+        /// <remarks>
+        /// Store this value at the time of read access and on next potential read use <see cref="IsDataInvalidated"/>
+        /// to check whether the data has potentially changed. This is the safest option as it accounts for other potential
+        /// conditions in the data.
+        ///
+        /// If comparing version numbers directly make sure to always use an equality check. The version number can
+        /// theoretically wrap and overflow so it's possible for the current value to be less than the previous value.
+        /// </remarks>
+        public uint Version { get; private set; }
```
jkeon, line 46: Not necessary for this PR but I wonder if we should wrap the `uint` into a custom struct that we only implement the equality check for and not the greater/less than checks. 

That would make it safer to use. Future TODO/PR of course though.

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/DataSource/Data/AbstractData.cs
```diff
@@ -102,5 +117,17 @@ public JobHandle GetDependencyFor(AccessType accessType)
         {
             return m_AccessController.GetDependencyFor(accessType);
         }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private void IncrementVersionIfWrite(AccessType accessType)
+        {
+            if (accessType is AccessType.ExclusiveWrite or AccessType.SharedWrite)
+            {
+                unchecked
```
jkeon, line 126: I think it's fine to leave this just to signify that we have thought about it and we want to have the wrap around behaviour but I don't think we need it. By default it should be unchecked.

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/statements/checked-and-unchecked

jkeon (changes requested):
> Also, the original IsDataInvalidated check was backward but that's fixed now and I buried that little tidbit mid paragraph to save @jkeon some embarrassment.

A thousand gratitudes!

Seeing you go through this and wrap your head around it brings a tear to my eye, this was complicated and I'm happy I didn't have to wade back in here. 

Looks really good, great solutions!

Only a few comments that are pretty minor but worth looking at it.

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/DataSource/Data/AbstractData.cs
mbaker3: You're right that it's not necessary with the default compiler config but if someone turns on overflow checking then we don't want this to trigger when it overflows because it's expected and handled behaviour.

By contrast, `IDProvider` doesn't have an explicit `unchecked` increment because a consumer may want an exception if they don't handle the threshold warnings.

That's my rationale anyway.

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/Cancellation/CancelProgressFlowNode.cs
mbaker3: struct because it's lighter weight and not necessary. Less burden on the GC.

I'll admit, this is a bit convoluted but I'm worried about a set of potential errors if the consumer of the shared/ref counted value calls dispose or release multiple times. Having the wrapper/handle sort of type allows us to protect against Dispose being called multiple times and breaking the ref count.

Rather than a ref count, the alternative would be to maintain a `HashSet` of instances that have acquired the shared value. Then on release/dispose we only decrement the count if the instance exists in the collection and remove it. This approach seemed heavier weight than the wrapper/handle  struct though.

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/Cancellation/CancelProgressFlowNode.cs
```diff
@@ -37,17 +38,13 @@ public CancelProgressFlowNode(ITaskSetOwner taskSetOwner, CancelProgressFlowNode
...
-
-            if (m_Parent != null)
-            {
-                m_ParentProgressLookupData = m_Parent.m_TaskSetOwner.TaskSet.CancelProgressDataStream.ActiveLookupData;
-            }
         }
 
         protected override void DisposeSelf()
         {
+            m_LastProcessedDataVersion.Dispose();
```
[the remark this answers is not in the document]
mbaker3: Doesn't work because the type is a `struct`. This should always be initialized though because we init it in the constructor.

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/Cancellation/CancelProgressFlowNode.cs
mbaker3: We COULD do this but there are two drawbacks that make this a bad idea
 - Would have to add `readonly` to the `ProgressLookupDataVersion` getter/setter. This is valid because the property doesn't actually mutate the struct but seems odd given that there's a setter present.
 - Calling `Dispose()` would cause the struct to be copied on each call. This would circumvent the whole purpose of the struct existing because the managed reference wouldn't be null out in the version stored on the field.

mbaker3:
@jkeon Responded to comments. Open to discuss on a call if you want.

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/Cancellation/CancelProgressFlowNode.cs
```diff
@@ -212,5 +212,89 @@ private bool CheckIfWillComplete(bool isStillProcessing, ref EntityProxyInstance
...
+
+
+            public OwnerSharedDataVersion(ITaskSetOwner owner)
+            {
+                m_Ref = VersionRefWrapper.Acquire(owner);
+            }
+
+            public void Dispose()
+            {
+                m_Ref.Release();
```
jkeon, line 247:
```suggestion
                m_Ref?.Release();
```

If we call this twice, we'll get an error.

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/Cancellation/CancelProgressFlowNode.cs
jkeon: > struct because it's lighter weight and not necessary. Less burden on the GC.

I think it's the same weight no? A class would just be a pointer, this struct just wraps a reference to a class which would just be the pointer. 

I would agree less burden on the GC if we were created and destroying throughout the lifecycle, but these will get created once at startup and disposed once at teardown. 

> I'll admit, this is a bit convoluted but I'm worried about a set of potential errors if the consumer of the shared/ref counted value calls dispose or release multiple times. Having the wrapper/handle sort of type allows us to protect against Dispose being called multiple times and breaking the ref count.
>
> Rather than a ref count, the alternative would be to maintain a HashSet of instances that have acquired the shared value. Then on release/dispose we only decrement the count if the instance exists in the collection and remove it. This approach seemed heavier weight than the wrapper/handle struct though.

Ok I see the value in having the second time you call dispose do a no-op. With that though we will have an error in the `Release` call, I've made a separate comment about it below.

The HashSet approach would work but yeah, you've already done this so it's fine. I think i was more concerned about us leaning in toward structs because we're so ingrained with DOTS now and this isn't really the place where it makes a difference and it pushes us out of our patterns for OO land with the `readonly` etc. 

I think you could accomplish the same thing with this being a class and having the static inner dictionary for the record-keeping. `PersistentDataSystem` does this today. 

The issue with multiple dispose calls could be mitigated by just checking `IsDisposed` before decrementing. 

All that said, this is here and it works, so I'm cool with it staying like this.

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/Cancellation/CancelProgressFlowNode.cs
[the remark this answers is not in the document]
jkeon: Yeah it was depending on https://github.com/decline-cookies/anvil-unity-dots/pull/268/files#r1234045416 but I couldn't link to a pending comment.

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/Cancellation/CancelProgressFlowNode.cs
jkeon: Yeah it was depending on https://github.com/decline-cookies/anvil-unity-dots/pull/268/files#r1234045416 but I couldn't link to a pending comment.

jkeon (commented):
Updated with responses

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/DataSource/Data/AbstractData.cs
jkeon: Yep, that's fair.

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/Cancellation/CancelProgressFlowNode.cs
mbaker3: I thought about this some more and there isn't enough safety provided by the struct wrapper to justify using it over class. I think the GC pressure and number of indirect lookups is still a mild concern but I shouldn't be pre-optimising.

`OwnerSharedDataVersion` is now a class.

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/Cancellation/CancelProgressFlowNode.cs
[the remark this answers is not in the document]
mbaker3: now a class and readonly.

Scripts/Runtime/Entities/TaskDriver/TaskSet/TaskData/DataStream/Cancellation/CancelProgressFlowNode.cs
mbaker3: now a class and readonly.

mbaker3:
**Ready for re-review**

Addressed feedback from the review.

Also found a couple more edge cases and fixed them in the commits following the feedback.

## 9bdf306 - Fix Cancellation Soft Lock
Fixed soft lock if a cancel request encounters a stream configured to unwind that doesn't have any data in it.

A job configured on an unwind stream that has no data will not schedule. Since the job doesn't schedule it doesn't get a write handle to the cancel progress lookup and then doesn't increment the data version.

Since the data version doesn't increment the CancelProgressFlowNode doesn't re-process and the cancel tree is stuck in a perpetual state of waiting for the unwind to complete. Sending subsequent cancel requests will bump the processing and allow the cancel tree to resolve but this will happen for each node in the tree where no data is in the stream!

## adf85c0 - Prevent Stale Data from persisting in data sources
Fix bug where active data would remain in the stream/lookup when the data version no longer changed.
This is an issue for any data source that is read when the data version doesn't change (Ex: CancelRequestDataSource).
Now all data sources will get one extra consolidation after the last data version change.

Specifically, this fixed an issue where cancellation would never complete for a given key when a stream configured to unwind was present in the cancellation tree, had no elements, and no writers after cancellation.

jkeon (approved):
Looks great, thanks for going on that wild journey with me, deep down the caverns of the debugger!

mbaker3:
All good, thanks for helping me figure these out. I have a slightly better understanding of how this all works now!

PR #269, opened 2023-06-23
Exposing the ability to schedule Vanilla Unity jobs off of DataStreams

jkeon:
TaskDrivers utilize DataStreams to do their work but sometimes you need to run a job in a Vanilla Unity system based on those datastreams. 

No nice way to handle this.

Introducing a wrapper for a DataStream to manage scheduling a job off of a DataStream while checking if the data has possibly been invalidated or not. This makes it easy to include in your system and not have to worry about managing the last known version etc.

- None

 - None

 - [ ] Yes
 - [x] No


mbaker3 (approved):
nice one!

PR #270, opened 2023-06-26
Changing to use Key instead of Entity

jkeon:
We often use `IEntityProxyInstance.Entity` as a "Key" 
This change just makes the naming better reflect that.

Used to be called `Entity`

Now is called `Key`

- None

 - None

 - [x] Yes - Simple renaming
 - [ ] No


mbaker3 (approved):
Looks good.

Doesn't have to be part of this PR but does `IEntityProxyInstance` still make sense as a name?
The name is 2-3 refactors separated from its original inception and, I don't think, it makes much sense anymore. `IEntityKeyedInstance`?
Should we just call these tasks? `IEntityKeyedTask`

jkeon:
@mbaker3 Renamed `IEntityProxyInstance` to `IEntityKeyedTask`

mbaker3 (approved):

PR #272, opened 2023-07-19
HashCodeUtil - Fix Docs, Add overloads for more params

mbaker3:
- Improve docs in `HashCodeUtil`
 - Add convenience overloads to accept 3 and 4 hash codes to combine.
 - Add hint to inline the utility methods
 - Add burst compile attribute to ensure class remains burstable.

There are no overloads to combine more than two hash codes.

`GetHashCode()` for three and four parameters has been added.

 - None

 - None

 - [ ] Yes 
 - [x] No

### DIFF
--- a/Scripts/Runtime/Util/HashCodeUtil.cs
+++ b/Scripts/Runtime/Util/HashCodeUtil.cs
@@ -4,7 +4,7 @@ using Unity.Burst;
 namespace Anvil.Unity.DOTS.Util
 {
     /// <summary>
-    /// Helper class for generating hash codes that is Burst compatible.
+    /// Helper class for generating a hash code that is Burst compatible.
     /// </summary>
     [BurstCompile]
     public static class HashCodeUtil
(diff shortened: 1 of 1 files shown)


Scripts/Runtime/Util/HashCodeUtil.cs
```diff
@@ -1,19 +1,24 @@
+using System.Runtime.CompilerServices;
+using Unity.Burst;
+
 namespace Anvil.Unity.DOTS.Util
 {
     /// <summary>
-    /// Helper class for generating hashcodes that are Burst compatible.
+    /// Helper class for generating hash codes that is Burst compatible.
```
jkeon, line 7:
```suggestion
    /// Helper class for generating a hash code that is Burst compatible.
```

jkeon (changes requested):
Looks good, one minor doc issue

jkeon (approved):

PR #275, opened 2023-09-11
IJobConfig - Add require exclusive write for EPD and CDFE + Cleanup

mbaker3:
- Add require methods for
    - Exclusive `EntityPersistentData` write
    - Exclusive CDFE write
    - Shared DBFE write
 - Cleanup `IJobConfig` and implementations 

Within an TaskDriver there is no way to:
 - Get exclusive write access to `EntityPersistentData`
 - Get exclusive write access to a CDFE
 - Get shared write access to a DBFE

As the use of Task Drivers has expanded well beyond the simple use cases originally imaged the the method naming for write access requirements has become ambiguous. Depending on the type, "Write" in a method name (Ex: `RequireEntityPersistentDataForWrite` could mean that the access handle is shared between all other concurrent write requests globally, all concurrent write requests associated with a Task Driver type's `TaskDriverSystem`, or something else. It's easy to forget or assume what type of access a you're getting as a developer. Shared vs Exclusive needs to be spell explicitly and new terminology needs to be devised to describe the unique write access provided to Unity ECS backed types (CDFE, DBFE, etc..)

IJobConfig: Rename require methods to better describe provided access
 - Introduce the key term "SystemShared" for Unity ECS backed type requests (Ex: CDFE, DBFE, etc...)
   - "SystemShared" denotes that the shared write fence is on the TaskDriver system vs a normal SharedWrite that shares write access across all access scopes. This is due to the way Unity handles component dependencies.
 - Rename RequireCDFEForWrite to RequireCDFEForSystemSharedWrite.
 - Rename RequireEntityPersistentDataForWrite to RequireEntityPersistentDataForSharedWrite

IJobConfig: Additional require types
 - Add RequireEntityPersistentDataForExclusiveWrite
 - Add RequireCDFEForExclusiveWrite
 - Add RequireDBFEForSystemSharedWrite

IJobConfig + Implementations: Cleanup
 - Add comment headers to all sections
 - Reorder methods to consistently declare require methods in the order read, shared write, exclusive write
 - Refactor implementations to match the method order defined in IJobConfig

**Notes:**
The following methods were left as is for the reasons given below. I'm not against changing their names if you think it would be beneficial and have some naming ideas.
 - `RequireThreadPersistentDataForWrite` has been left as is since there is no concept of shared vs exclusive access.
 - `RequireDataStreamForWrite` has been left as is since exclusive writing isn't a user space operation that should ever be performed.

 - None

 - None

 - [x] Yes - See the method renames above.
 - [ ] No

### DIFF
--- a/Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/JobConfig/Interface/IJobConfig.cs
+++ b/Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/JobConfig/Interface/IJobConfig.cs
@@ -245,7 +245,7 @@ namespace Anvil.Unity.DOTS.Entities.TaskDriver
 
         /// <summary>
         /// Specifies a <see cref="BufferFromEntity{T}"/> to be written to in a system scoped shared-write context.
-        /// This means that write access is shared between all Task Driver instances of the same type in the same &lt;see cref="World"/&gt;.
+        /// This means that write access is shared between all Task Driver instances of the same type in the same <see cref="World"/>.
         /// </summary>
         /// <typeparam name="T">The type of <see cref="IBufferElementData"/> in the DBFE.</typeparam>
         /// <returns>A reference to itself to continue chaining configuration methods.</returns>
--- a/Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/JobConfig/Interface/IJobConfig.cs
+++ b/Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/JobConfig/Interface/IJobConfig.cs
@@ -44,6 +44,11 @@ namespace Anvil.Unity.DOTS.Entities.TaskDriver
         /// <returns>A reference to itself to continue chaining configuration methods.</returns>
         public IJobConfig RunOnce();
 
+
+        //*************************************************************************************************************
+        // CONFIGURATION - REQUIRED DATA - Data Stream
+        //*************************************************************************************************************
+
         /// <summary>
         /// Specifies a <see cref="IAbstractDataStream{TInstance}"/> to be written to in a shared-write context.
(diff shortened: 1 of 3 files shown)


Scripts/Runtime/Entities/TaskDriver/TaskSet/Job/JobConfig/Interface/IJobConfig.cs
```diff
@@ -193,6 +243,15 @@ public IJobConfig RequireCDFEForWrite<T>()
         public IJobConfig RequireDBFEForRead<T>()
             where T : struct, IBufferElementData;
 
+        /// <summary>
+        /// Specifies a <see cref="BufferFromEntity{T}"/> to be written to in a system scoped shared-write context.
+        /// This means that write access is shared between all Task Driver instances of the same type in the same &lt;see cref="World"/&gt;.
```
jkeon, line 248: Formatting break for the angle brackets?

jkeon (dismissed):
Looks good, one minor formatting break. 

I think it's fine to leave the ThreadPersistentData as just "Write". Technically, it's a shared write if we want to be specific. 

Same thing for the data stream, I'm fine with it being just "Write" but again if we wanted to be specific, it's a Shared Write.

mbaker3:
Ready for re-review.

> I think it's fine to leave the ThreadPersistentData as just "Write". Technically, it's a shared write if we want to be specific.
> 
> Same thing for the data stream, I'm fine with it being just "Write" but again if we wanted to be specific, it's a Shared Write.

Agreed, Shared write felt a bit weird since it's not something you need to be concerned about the way you are with other types. ...at least not if used correctly.

jkeon (approved):

PR #276, opened 2023-09-12
Remove `IComponentData` from `PrototypeVariant`

mbaker3:
Remove `IComponentData` from `PrototypeVariant`. If projects want to store the variant on their entities they can create their own component object with `PrototypeVariant` as a field.

Also performed some misc formatting cleanup + spelling in docs.

@jkeon Should `PrototypeVariant` get moved into our app project? I don't see it referenced by anything in `anvil-unity-dots`

`PrototypeVariant` implements `IComponentData`

`PrototypeVariant` no longer implements `IComponentData`

 - None

 - None

 - [x] Yes - Create your own `IComponentData` implementation to wrap a `PrototypeVariant` value.
 - [ ] No

### DIFF
--- a/Scripts/Runtime/Entities/Lifecycle/PrototypeVariant.cs
+++ b/Scripts/Runtime/Entities/Lifecycle/PrototypeVariant.cs
@@ -3,20 +3,19 @@ using Anvil.Unity.DOTS.Util;
 using System;
 using Unity.Burst;
 using Unity.Collections;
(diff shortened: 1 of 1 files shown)


jkeon (approved):
Looks good. 

This is being used by the EntitySpawner and related classes. Not sure why it wasn't showing up for you but it's definitely being used.

PR #281, opened 2023-09-22
Hotfix - Double PopulateOnEntity

jkeon:
Removing `IEntitySpawnDefinitionExtension` as it doesn't provide any value any more.

When using the `IEntitySpawnDefinitionExtension.CreateAndPopulate` methods, it would first call `SpawnDeferredEntity` and then `PopulateOnEntity`.

However, `SpawnDeferredEntity` already calls `PopulateOnEntity` so we end up calling `PopulateOnEntity` twice.

Normally for set operations, this is fine but not very performant. 

But the danger lies in adding to a dynamic buffer where you will now get two adds instead of one.

Removing the second `PopulateOnEntity` call is easy enough in `CreateAndPopulate` but then `CreateAndPopulate` is just a proxy for `SpawnDeferredEntity`.

Instead we can just delete this class and callers can call `entitySpawner.SpawnDeferredEntity(definition)` directly. 

- None

 - None

 - [x] Yes - If you were using `CreateAndPopulate` just use `SpawnDeferredEntity` directly.
 - [ ] No


mbaker3 (approved):
Good Call!

Issue #172, opened 2023-03-02
`AbstractDataSystem` - Prevent from existing in `PlayerLoop`


mbaker3:
By their nature `AbstractDataSystem`s do not update but they currently appear in the player loop. This is misleading and can result in developers positioning their systems relative to `AbstractDataSystem`s in the player loop when they really shouldn't.

Ideally we prevent these systems from ever being added to a group or player loop. If that's not feasible then a data system should remove itself from the player loop at first opportunity.

We should also assert against other system's positioning themselves against this system in some way.

PR #285, opened 2023-10-02
AbstractDataSystem - Add to a default group

mbaker3:
Add all `AbstractDataSystem` instances to their own group until #172 is addressed.

This keeps all of the data systems together in the player loop until we have a chance to remove them.

`AbstractDataSystem` instances get littered throughout `SimulationSystemGroup` cluttering the player loop with update order that really doesn't matter.

`AbstractDataSystem` is now added to the group `AbstractDataSystem.Group` in the `InitializationSystemGroup`.

 - None

 - None

 - [ ] Yes 
 - [x] No


jkeon (approved):
Good idea!

PR #292, opened 2023-11-20
`UnsafeArrayUtility` - Add `AsUnsafeArray` overloads

mbaker3:
Add two more overloads for `UnsafeArrayUtility.AsUnsafeArray` that represent `DynamicBuffer<T>` and `UnsafeList<T>` as `UnsafeArray<T>` instances

Users don't have a convenient way to represent a `DynamicBuffer<T>` or `UnsafeList<T>` as an `UnsafeArray<T>`.

Users can now call `AsUnsafeArray` on both `DynamicBuffer<T>` and `UnsafeList<T>` instances.

 - None

 - None

 - [ ] Yes 
 - [x] No


jkeon (approved):
Looks good!

PR #293, opened 2023-11-20
UnsafeArray - Make AsReadOnly a pure method

mbaker3:
Make `UnsafeArray.AsReadOnly` a pure method. This prevents a defensive copy when the instance is `readonly`

`AsReadonly()` is not marked `readonly` even though it will never mutate the instance. This causes a defensive copy when the instance is readonly referenced (`readonly` field, `in`, `ref readonly`, etc...)

`UnsafeArray.AsReadOnly` is marked as `readonly` indicating to the runtime that it's safe to call this method without creating a defensive copy.

 - None

 - None

 - [ ] Yes 
 - [x] No - "free" performance


jkeon (approved):
Love it.

